PR #23, opened 2015-08-12
Write internal modules so we can reference types from global namespace.

alexeagle:
Also adapts each to an external module for users to consume.


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -130,15 +130,32 @@ public String produceDts(Compiler compiler) {
...
-      emitNoSpace("' {");
-      indent();
-      emitBreak();
       TypedVar symbol = topScope.getOwnSlot(provide);
       checkArgument(symbol != null, "goog.provide not defined: %s", provide);
+      String namespace = provide;
+      boolean isDefault = false;
+      if (symbol.getType() != null) {
+        // These types are not expected to have nested properties
+        // so we declare it as a default export
```
rkirov: `nested properties` might not be the right wording.

rkirov:
LGTM.

alexeagle:
merged inw 8c88aed

PR #28, opened 2015-08-20
Generic types

alexeagle:
Previously this happened to work, because we only tested prototype methods on the default exported class

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -3,11 +3,8 @@ package com.google.javascript.cl2dts;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
-import com.google.common.base.Function;
-import com.google.common.base.Joiner;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
-import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
@@ -56,7 +53,6 @@ public class DeclarationGenerator {
 
   private static final Logger logger = Logger.getLogger(DeclarationGenerator.class.getName());
   private static final String INTERNAL_NAMESPACE = "ಠ_ಠ.cl2dts_internal";
-  public static final Joiner ON_COMMA = Joiner.on(", ");
 
   private StringWriter out = new StringWriter();
   private final boolean parseExterns;
@@ -134,6 +130,7 @@ public class DeclarationGenerator {
 
     out = new StringWriter();
     TypedScope topScope = compiler.getTopScope();
+    Iterable<TypedVar> allSymbols = topScope.getAllSymbols();
     for (String provide : provides) {
       TypedVar symbol = topScope.getOwnSlot(provide);
       checkArgument(symbol != null, "goog.provide not defined: %s", provide);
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -148,37 +157,54 @@ public String produceDts(Compiler compiler) {
...
-          if (otherName.startsWith(prefix) && other.getType() != null
-              && !other.getType().isFunctionPrototypeType()) {
-            walkScope(other, namespace, false);
-          }
-        }
-      }
-      unindent();
-      emit("}");
-      emitBreak();
+      currentNamespace = namespace;
```
mprobst: This IMHO introduces a rather surprising piece of global state into the translation. What about extracting a `relativeTo(namespace)` method that returns the appropriate `Function`?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -282,30 +308,33 @@ private void walkScope(TypedVar symbol, final String namespace, boolean isDefaul
...
 
-      Function<ObjectType, String> relativeName = new Function<ObjectType, String>() {
-        @Override public String apply(ObjectType input) {
-          return !input.getDisplayName().startsWith(namespace) ? input.getDisplayName()
-              : input.getDisplayName().substring(namespace.length() + 1);
-        }
-      };
+      if (type.hasAnyTemplateTypes()) {
+        emit("<");
+        emit(FluentIterable.from(ftype.getTemplateTypeMap().getTemplateKeys()).transform(new Function<JSType, String>() {
```
mprobst: YMMV, but I'm not convinced this is substantially simpler than just using an explicit iterator:
```java
Iterator<JSTyle> it = ftype.getTemplateTypeMap().getTemplateKeys();
while (it.hasNext()) {
  emit(it.next().getDisplayName());
  if (it.hasNext()) emit(",");
}
```

That's ballpark the same SLOC, but works with a lot fewer concepts, doesn't yield the control flow to a fluent iterable so easier to step through, and is probably faster to boot (not that it matters).

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -148,37 +157,54 @@ public String produceDts(Compiler compiler) {
...
-      emitBreak();
+      currentNamespace = namespace;
+      declareNamespace(provide, symbol, isDefault, allSymbols);
       declareModule(provide, isDefault);
     }
     checkState(indent == 0, "indent must be zero after printing, but is %s", indent);
     return out.toString();
   }
 
+  private void declareNamespace(String provide, TypedVar symbol,
```
mprobst: Does this need rebasing? The diff looks surprising.

src/test/java/com/google/javascript/cl2dts/types.js
```diff
@@ -11,3 +11,21 @@ types.c = 's';
 
 /** @type {Object} */
 types.d = {};
+
+/** @type {Array<?>} */
```
mprobst: Could we special case `Array<X>` as `X[]`?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: I agree, and fought with this for a while.
The problem is that we need to get a `relativeTo` all the way down in the visitor below, in `caseTemplatizedType`, and we need some significant refactoring to make a different instance of the visitor for each namespace, and pass this through the helper method calls in between.
I started down that path and backed out because of the scale of the change. But we could do that first as a prefactoring and then come back to this...

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: yeah, there shouldn't be three commits on this PR, something looks wrong...

src/test/java/com/google/javascript/cl2dts/types.js
alexeagle: Yes, is it worth it? I suppose users will frequently be exposed to typescript idioms by reading the typings we expose through their editor and error messages, so maybe we want to teach them

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: Okay, fresh with a bit of extra sleep the refactoring was no problem.
Essentially I've introduced an inner class to replace the walkScope method, and the lifecycle of this class is one-per-top-level-symbol. So the namespace and the top level symbol can be state of this class, and it doesn't require changing *every* visit* method to accept a namespace parameter and pass it down.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: Yeah, I enjoyed writing the functional style but I'm not married to it. You're very naughty to suggest an if body without braces just to justify the smaller line count, however :smile: 
changed to iterative

src/test/java/com/google/javascript/cl2dts/types.js
alexeagle: done

alexeagle:
@rkirov ready to merge now. the delta looks kind of crazy even though the refactoring is pretty minor. can walk through in person if you like

rkirov:
I don't quite get the logic for the branching yet, but since this is a second pass, lets merge and I will post some comments when I have more time to think about it.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -147,38 +144,54 @@ public String produceDts(Compiler compiler) {
...
+        !symbol.getType().isObject()) {
+      new TreeWalker(symbol, namespace).walk(isDefault);
+    } else {
+      // JSCompiler treats "foo.x" as one variable name, so collect all provides that start with
+      // $provide + ".".
+      String prefix = provide + ".";
+      for (TypedVar other : allSymbols) {
+        String otherName = other.getName();
+        if (otherName.startsWith(prefix) && other.getType() != null
+            && !other.getType().isFunctionPrototypeType()
```
rkirov: what is the difference between isFunctionPrototypeType and our implementation of isPrototypeMethod?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -147,38 +144,54 @@ public String produceDts(Compiler compiler) {
...
+      boolean isDefault, Iterable<TypedVar> allSymbols) {
+    emitNoSpace("declare namespace ");
+    emitNoSpace(INTERNAL_NAMESPACE);
+    if (!namespace.isEmpty()) {
+      emitNoSpace("." + namespace);
+    }
+    emitNoSpace(" {");
+    indent();
+    emitBreak();
+    if (symbol.getType().isEnumType() || symbol.getType().isFunctionType() ||
```
rkirov: I can't convince myself why this can't just be if (isDefault)

isDefault = `!symbol.getType().isObject() || symbol.getType().isInterface() || symbol.getType().isFunctionType();`
checkHere = `symbol.getType().isEnumType() || symbol.getType().isFunctionType() ||!symbol.getType().isObject()`

so the symmetric diff is `isEnumType` and `isInterface`. Why don't threat enums as default exports? And why would we go into the `else` branch here if the type is interface?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: I don't know what those mean - isFunctionPrototypeType refers to the "owner" and the isPrototypeMethod refers to the "typeOfThis". But I do know that isFunctionPrototypeType isn't sufficient - a test covers the necessity for also checking if it has a this reference.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: reviewed in person, Rado is right, they should use the same conditional. sent follow-up PR.

PR #32, opened 2015-08-25
feature(types): support function type literals

alexeagle:
Also rename the parameters in `fn_params.js` to clarify that the names they are given in the .d.ts output have no relation to the names in the .js file


rkirov:
LGTM

alexeagle:
merged in 8cfb873

PR #33, opened 2015-08-25
fix(bug): enums should also be default export

alexeagle:
thanks @rkirov for reasoning about the difference between our two boolean conditionals. They should be the same condition, and this exposed a bug.


rkirov:
LGTM, pair coded on this one.

PR #35, opened 2015-08-25
Record type

alexeagle:


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -440,7 +459,11 @@ public Void caseStringType() {
 
         @Override
         public Void caseObjectType(ObjectType type) {
-          emit("Object");
+          if (type.isRecordType()) {
```
rkirov: There is no caseRecordType?

alexeagle: nope, I guess this is something closure added on top of rhino's very old type system?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -166,11 +168,20 @@ private void declareNamespace(String namespace, String provide, TypedVar symbol,
       new TreeWalker(symbol, namespace).walk(true);
     } else {
       // JSCompiler treats "foo.x" as one variable name, so collect all provides that start with
-      // $provide + ".".
-      String prefix = provide + ".";
+      // $provide + "." but are not sub-properties.
+      Set<String> desiredSymbols = new HashSet<>();
+
+      TypedVar o = Iterables.find(allSymbols, nameIs(provide));
```
rkirov: I am confused, isn't `symbol` equal to `o`, because `TypedVar symbol = topScope.getOwnSlot(provide);`?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -166,11 +168,20 @@ private void declareNamespace(String namespace, String provide, TypedVar symbol,
...
+      if (o.getType().isObject()) {
```
rkirov: Also, I think we are guaranteed to have isObject true here, because we are in is not default branch.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -166,11 +168,20 @@ private void declareNamespace(String namespace, String provide, TypedVar symbol,
...
+        ObjectType objType = (ObjectType) o.getType();
+        for (String property : objType.getPropertyNames()) {
+          desiredSymbols.add(o.getName() + "." + property);
+        }
+      }
+
       for (TypedVar other : allSymbols) {
```
rkirov: why not iterate just `desiredSymbols`

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: Yes, good observation. I had something else here and refactored to this... earlier I had the top level symbol instead of the one for this provide

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: yes, I had to squint to convince myself of that also

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: because the order is different.
I'm not sure it's that important for the order to match the declarations in the js file, but it seems desirable compared with any other ordering.

rkirov: That's a good point. It is better to match the order.

rkirov:
LGTM.

alexeagle:
merged in 4ed1f00

PR #38, opened 2015-08-26
fix(bug): don't emit 'implements' when a superclass is the implementor

alexeagle:
Found this bug while emitting .d.ts for `javascript/closure/ui/ac/autocomplete.js`

### DIFF
--- a/src/test/java/com/google/javascript/cl2dts/multi_class.d.ts
+++ b/src/test/java/com/google/javascript/cl2dts/multi_class.d.ts
@@ -10,6 +10,8 @@ declare namespace ಠ_ಠ.cl2dts_internal.multi_class {
   }
   interface I2 extends I {
   }
+  class C extends B {
+  }
 }
 declare module 'goog:multi_class' {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/cl2dts/multi_class.d.ts
```diff
@@ -10,6 +10,8 @@ declare namespace ಠ_ಠ.cl2dts_internal.multi_class {
   }
   interface I2 extends I {
   }
+  class C extends B {
+  }
```
rkirov: can you add `class D implements I` for completeness of all possible cases.
- class implements
- class extends
- class implements and extends
- interface extends

rkirov:
LGTM, just add one more test case (I am sure it works :))

src/test/java/com/google/javascript/cl2dts/multi_class.d.ts
alexeagle: yeah, good idea.

alexeagle:
merged as cac4c4b

PR #39, opened 2015-08-26
Add test for one file which depends on another via goog.provide/require.

### LINKED ISSUE #30: Handle multiple inputs with dependencies
When multiple files are presented, and one of them goog.require's a symbol that is goog.provide'd by another, we should
- write one .d.ts file for each .js input
- use our global internal namespace to reference symbols from the other file.

alexeagle:
Write a .d.ts file for each .js input. Use the global namespace to
reference symbols between these files.

Fixes #30

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -2,10 +2,15 @@ package com.google.javascript.cl2dts;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.Iterables.getOnlyElement;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.UnmodifiableIterator;
+import com.google.common.io.Files;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
 import com.google.javascript.jscomp.CommandLineRunner;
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -463,10 +484,12 @@ public Void caseStringType() {
 
         @Override
         public Void caseObjectType(ObjectType type) {
-          if (type.isRecordType()) {
+          if (type.isNativeObjectType()) {
```
rkirov: dumb question, what is the diff between NativeObjectType and ObjectType?

src/test/java/com/google/javascript/cl2dts/require/RequireTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.runners.JUnit4;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.PrintStream;
+import java.nio.file.FileSystems;
+import java.nio.file.Path;
+
+// TODO(alexeagle): should we have an explicit test class like this in each directory?
+// Or should we auto-discover them as we do for test cases in the root directory?
```
rkirov: auto-discover usually doesn't scale well, when we need some flexibility. Rather go for explicit over implicit.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: NativeObjectType seems to be the builtin 'Object' type, as opposed to a
subtype of Object. but I don't really know

rkirov: I guess, something that is not a Function, interface etc. I wonder if that can simplify our isDefault check (obviously, for the next time we revisit that code).

rkirov:
LGTM, but why does the test fail? Looks like it is not seeing one of the dependencies.

alexeagle:
abandoning, this is old and wrong

PR #46, opened 2015-08-26
Support generic methods and functions

alexeagle:


rkirov:
Add a test that has two type params separated with by a comma (unless I missed one already there). Otherwise, LGTM.

alexeagle:
Added that test.

alexeagle:
merged in ea84de5

PR #47, opened 2015-08-26
fix(constructors): handle goog.inherits and emit constructors

### LINKED ISSUE #22: Producing constructor with return type
this input
```
goog.provide('a');

/**
 * @constructor
 */
a.One = function() {};
a.One.prototype.foo = function(x) {};

/**
 * @constructor
 * @extends {a.One}
 */
a.Two = function() {};

goog.inherits(a.Two, a.One);
```

produces a class Two which has an explicit constructor. We have a return type on the result which is invalid syntax.

```
declare module 'goog:a' {
  export class One {
    foo (a : any ) : void ;
  }
  export function foo (a : any ) : void ;
  export class Two {
    constructor ( ...a : any [] ) : any ;
    static superClass_ : Object ;
  }
}
```

alexeagle:
Fixes #22

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -556,7 +556,12 @@ public class DeclarationGenerator {
 
         @Override
         public Void caseFunctionType(FunctionType type) {
-          visitFunctionType(type, "=>");
+          visitFunctionSignature(type);
+          JSType returnType = type.getReturnType();
+          if (returnType != null) {
+            emit("=>");
+            visitType(returnType);
+          }
           return null;
         }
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -641,10 +655,15 @@ private void visitProperties(ObjectType objType, boolean isStatic) {
...
+      visitFunctionType(ftype);
+      JSType type = ftype.getReturnType();
+      if (type != null) {
+        emit(":");
+        visitType(type);
+      }
     }
 
-    private void visitFunctionType(FunctionType ftype, String separator) {
+    private void visitFunctionType(FunctionType ftype) {
```
rkirov: I find it semi-suprising that visiting function type doesn't take care of constructors, add a comment to remind ourselves.

rkirov: or rename to visitFunctionArgs.

rkirov:
LGTM, just one naming nit.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: renamed to visitFunctionSignature

alexeagle:
merged in e0a9f88

PR #50, opened 2015-08-27
fix(naming): Write parameter lists using the original names, where po…

### LINKED ISSUE #48: Try to preserve parameter names
It's going to be really confusing for developers to see different parameter names (a,b,c...) in the type definitions than were declared. It loses the documentation benefit of seeing the names of the parameters in an API you're trying to call. We should preserve them where possible. (only function literals in typescript require manufacturing names for the parameter types)

alexeagle:
…ssible.

Fixes #48


rkirov:
LGTM

alexeagle:
merged in 5ddbb555dece183e2b811d7de04c2d2ccfaf43b8

PR #51, opened 2015-08-27
implement type-var pair for Closure interfaces

### LINKED ISSUE #45: Emitting static members on interface
```
/**
 * An interface that describes a single registered listener.
 * @interface
 */
goog.events.ListenableKey = function() {};

/**
 * Reserves a key to be used for ListenableKey#key field.
 * @return {number} A number to be used to fill ListenableKey#key
 *     field.
 */
goog.events.ListenableKey.reserveKey = function() {
  return ++goog.events.ListenableKey.counter_;
};
```
produces
```
  interface ListenableKey {
    static reserveKey ( ) : number ;
  }
```

but TS doesn't allow static method declarations on an interface

rkirov:
Unlike TS interfaces closure interfaces are allowed to have 'static'
methods and properties. In TS we split the closure interface into a TS
interface along with a record type declaration for a var.

Closes #45

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -5,9 +5,12 @@ import static com.google.common.base.Preconditions.checkState;
 
 import com.google.common.base.Function;
 import com.google.common.base.Predicate;
+import static com.google.common.base.Predicates.not;
+import static com.google.common.base.Predicates.equalTo;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import static com.google.common.collect.Iterables.any;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
 import com.google.javascript.jscomp.CommandLineRunner;
@@ -315,14 +318,7 @@ public class DeclarationGenerator {
       JSType type = symbol.getType();
       if (type.isFunctionType()) {
         FunctionType ftype = (FunctionType) type;
-        if (isDefault && ftype.isInterface()) {
-          // Have to produce a named interface and export that.
-          // https://github.com/Microsoft/TypeScript/issues/3194
-          emit("interface");
-          emit(getUnqualifiedName(symbol));
-          visitObjectType(ftype, ftype.getPrototype());
-          return;
-        }
+
         if (type.isOrdinaryFunction()) {
           emit("function");
           emit(getUnqualifiedName(symbol));
@@ -370,7 +366,19 @@ public class DeclarationGenerator {
           }
         }
 
-        visitObjectType(ftype, ftype.getPrototype());
+        if (ftype.isInterface()) {
+          visitObjectType(ftype, ftype.getPrototype(), false);
+
+          if (any(ftype.getOwnPropertyNames(), not(equalTo("prototype")))) {
+            emit("var");
+            emit(getUnqualifiedName(symbol));
+            emit(":");
+            visitRecordLikeType(ftype);
+            emitBreak();
+          }
+        } else {
+          visitObjectType(ftype, ftype.getPrototype());
+        }
       } else {
         if (type.isEnumType()) {
           visitEnumType((EnumType) type);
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -314,7 +314,13 @@ private void walk(TypedVar symbol, boolean isDefault) {
           // https://github.com/Microsoft/TypeScript/issues/3194
           emit("interface");
           emit(getUnqualifiedName(symbol));
-          visitObjectType(ftype, ftype.getPrototype());
+          visitObjectType(ftype, ftype.getPrototype(), false);
+
+          emit("var");
```
alexeagle: can/should we avoid emitting the var if the interface has no static members?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -573,6 +579,32 @@ public Void caseProxyObjectType(ProxyObjectType type) {
...
+      emit("{");
+      Iterator<String> it = type.getOwnPropertyNames().iterator();
+      Boolean isFirst = true;
+      while (it.hasNext()) {
+        String propName = it.next();
+
+        // The prototype key is implicitly defined in the Object signature.
+        if (propName.equals("prototype")) continue;
+
+        if (isFirst) {
```
alexeagle: can't you just check if it.hasNext()?

rkirov: leaves a comma if the last one is going to be skipped because it is 'prototype'. Iterators can't peek the next value as far as I can tell, so I change the logic to add the comma before each output, instead of after.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: Good catch. I started wondering how test pass in that case and discovered we were forking on isDefault unnecessarily.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -55,6 +57,12 @@
 
   private static final Logger logger = Logger.getLogger(DeclarationGenerator.class.getName());
   private static final String INTERNAL_NAMESPACE = "ಠ_ಠ.cl2dts_internal";
+  public static final Predicate<String> NOT_PROTOTYPE = new Predicate<String>() {
```
alexeagle: unneeded, see below

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -364,7 +365,19 @@ private void walk(TypedVar symbol, boolean isDefault) {
           }
         }
 
-        visitObjectType(ftype, ftype.getPrototype());
+        if (ftype.isInterface()) {
+          visitObjectType(ftype, ftype.getPrototype(), false);
+
+          if (Iterables.any(ftype.getOwnPropertyNames(), NOT_PROTOTYPE)) {
```
alexeagle: you can use Predicates.not(Predicates.equalTo("prototype"))
and static import to get
not(equalTo("prototype"))
opt: static import for "any" as well

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: I've used http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/PeekingIterator.html for this before - I think it's already available in our classpath? See if that makes it less awkward

src/test/java/com/google/javascript/cl2dts/interface_usage.ts
```diff
@@ -2,4 +2,6 @@
 import I from 'goog:interface_exp';
 class C implements I {
     method(): number { return 0; }
-}
\ No newline at end of file
+}
+
+var a: number = I.staticMethod();
```
alexeagle: can you also add a test case with an interface that has no static members? I think that is missing

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: or possibly 
`it = Iterables.filter(it, not(equalTo("prototype"))`
and then iterate `it`

src/test/java/com/google/javascript/cl2dts/interface_usage.ts
rkirov: We already have those in multi_class tests.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: done.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: done.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: done.

rkirov:
PTAL.

PR #53, opened 2015-08-28
fix(bug): don't export properties that are separately provide'd

### LINKED ISSUE #52: Duplicate declarations
```
goog.provide("goog.fx.dom");
goog.provide("goog.fx.dom.ColorTransform");

/** @type {number} */
goog.fx.dom.a = 1;

/** @constructor */
goog.fx.dom.ColorTransform = function() {}
```

we visit the ColorTransform symbol twice: once for the first provide (which is non-default so we traverse all the symbols) and again for the second provide, where it is the default.

alexeagle:
Fixes #52

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -152,15 +152,15 @@ public class DeclarationGenerator {
         int lastDot = symbol.getName().lastIndexOf('.');
         namespace = lastDot >= 0 ? symbol.getName().substring(0, lastDot) : "";
       }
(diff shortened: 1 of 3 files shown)


rkirov:
No usage file, or forgot to add? I think it would be useful to verify that importing `goog:goog.one` and `goog:goog.one.Two` is safe.

rkirov:
LGTM, just add usage file

alexeagle:
I tested this with google3 and there is still a problem with duplicates produced. trying to reproduce in this test case...

alexeagle:
I'll open a separate bug for the other duplication, it seems to be a different cause.

alexeagle:
merged in 4ddddad

PR #55, opened 2015-08-28
fix(types): emit object types

### LINKED ISSUE #54: Type 'Object' is not generic
We are writing eg `  var VOID_TAGS_ : Object < string , boolean > ;`
but this is not valid TS.

Could be `var VOID_TAGS_ : { [s: string]: string; };` instead

alexeagle:
Fixes #54

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -2,15 +2,17 @@ package com.google.javascript.cl2dts;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Predicates.equalTo;
+import static com.google.common.base.Predicates.not;
+import static com.google.common.collect.Iterables.any;
+import static com.google.javascript.rhino.jstype.JSTypeNative.ARRAY_TYPE;
+import static com.google.javascript.rhino.jstype.JSTypeNative.OBJECT_TYPE;
 
 import com.google.common.base.Function;
 import com.google.common.base.Predicate;
-import static com.google.common.base.Predicates.not;
-import static com.google.common.base.Predicates.equalTo;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
-import static com.google.common.collect.Iterables.any;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
 import com.google.javascript.jscomp.CommandLineRunner;
@@ -29,6 +31,7 @@ import com.google.javascript.rhino.jstype.EnumElementType;
 import com.google.javascript.rhino.jstype.EnumType;
 import com.google.javascript.rhino.jstype.FunctionType;
 import com.google.javascript.rhino.jstype.JSType;
+import com.google.javascript.rhino.jstype.JSTypeNative;
 import com.google.javascript.rhino.jstype.JSTypeRegistry;
 import com.google.javascript.rhino.jstype.NamedType;
 import com.google.javascript.rhino.jstype.NoType;
@@ -507,16 +510,38 @@ public class DeclarationGenerator {
         public Void caseTemplatizedType(TemplatizedType type) {
           ObjectType referencedType = type.getReferencedType();
           String templateTypeName = getRelativeName(referencedType);
-          if ("Array".equals(referencedType.getReferenceName())
+          if (typeRegistry.getNativeType(ARRAY_TYPE).equals(referencedType)
               && type.getTemplateTypes().size() == 1) {
             visitType(type.getTemplateTypes().get(0));
             emit("[]");
             return null;
           }
+          Iterator<JSType> it = type.getTemplateTypes().iterator();
+          if (typeRegistry.getNativeType(OBJECT_TYPE).equals(referencedType)) {
+            emit("{ [");
+            // TS allows only number or string as index type of an object
+            // https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.9.4
+            JSType keyType = it.next();
+            if (keyType.isNumberValueType()) {
+              emit("n: number");
+            } else {
+              if (!keyType.isStringValueType()) {
+                emit("/* warning: coerced from " + keyType + " */");
+              }
+              emit("s: string");
+            }
+            emit("]:");
+            visitType(it.next());
+            emit("}");
+            return null;
+          }
           emit(templateTypeName);
           emit("<");
-          for (JSType tmplType : type.getTemplateTypes()) {
-            visitType(tmplType);
+          while (it.hasNext()) {
+            visitType(it.next());
+            if (it.hasNext()) {
+              emit(",");
+            }
           }
           emit(">");
           return null;
--- a/src/test/java/com/google/javascript/cl2dts/types.js
+++ b/src/test/java/com/google/javascript/cl2dts/types.js
@@ -33,20 +33,29 @@ types.Foo.prototype.set = function(t) { };
  * @template V
  * @template W
  */
-types.Foo.prototype.loop = function(t, v) { return v(t); };
+types.Foo.prototype.genericMethod = function(t, v) { return v(t); };
 
 /** @type {types.Foo<string>} */
 types.f = new Foo();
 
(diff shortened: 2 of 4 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -2,15 +2,15 @@
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Predicates.equalTo;
```
rkirov: already added those in my commit, rebase.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -513,10 +513,28 @@ public Void caseTemplatizedType(TemplatizedType type) {
             emit("[]");
             return null;
           }
+          Iterator<JSType> it = type.getTemplateTypes().iterator();
+          if ("Object".equals(referencedType.getReferenceName())) {
+            emit("{ [");
+            // TS allows only number or string as index type of an object
+            // https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.9.4
+            if (it.next().isNumberValueType()) {
+              emit("n: number");
```
rkirov: And the runtime only allows strings as keys, so all that is damn lies :)

check
```
a = {};a[0] = 1; a['0']
```

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -513,10 +513,28 @@ public Void caseTemplatizedType(TemplatizedType type) {
...
+            } else {
+              emit("s: string");
+            }
```
rkirov: should we warn if the input was anything outside `string|number`, so users are not surprised?

src/test/java/com/google/javascript/cl2dts/types.js
```diff
@@ -50,3 +50,23 @@ types.h = {a: 'a', b: 34};
...
+/**
+ * Whether the object/map/hash contains the given value. This is O(n).
+ *
+ * @param {Object<K,V>} obj The object in which to look for val.
+ * @param {V} val The value for which to check.
+ * @return {boolean} true If the map contains the value.
+ * @template K,V
+ */
+types.containsValue = function(obj, val) {
+    for (var key in obj) {
```
rkirov: why include all that junk, we just need `@param {Object<K,V>}`.

rkirov:
LGTM.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: Maybe so, but I think this is correct according to spec?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: Yes, but I don't think we have any channel through Bazel to report those warnings somewhere the user can see. Maybe we could emit a comment into the .d.ts saying what happened here?

alexeagle: added a warning in an inline comment, tell me what you think

src/test/java/com/google/javascript/cl2dts/types.js
alexeagle: done

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: good workaround for now.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: yes, it was just an observation, matching the TS spec is the right thing for us.

rkirov:
still LGTM.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -513,10 +513,32 @@ public Void caseTemplatizedType(TemplatizedType type) {
             emit("[]");
             return null;
           }
+          Iterator<JSType> it = type.getTemplateTypes().iterator();
+          if ("Object".equals(referencedType.getReferenceName())) {
```
mprobst: I think somewhere in the type registry there is a specific constant (marker? something?) that identifies the one, special `Object` type. Maybe you could use that?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -513,10 +513,32 @@ public Void caseTemplatizedType(TemplatizedType type) {
...
+            emit("{ [");
```
mprobst: If the type is not templatized, could we just emit `Object`? Always emitting `{ [key: string]: number }` is a bit misleading - there are usages that are intended to be used as pseudo-maps (`Object<string, SomeThing>`) and usages where `Object` is just the super type of everything (well, most things). I think the latter should be translated as just `Object`, so that TypeScript doesn't offer arbitrary indexing into it.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -513,10 +513,32 @@ public Void caseTemplatizedType(TemplatizedType type) {
...
+            // TS allows only number or string as index type of an object
+            // https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#3.9.4
+            JSType keyType = it.next();
+            if (keyType.isNumberValueType()) {
+              emit("n: number");
+            } else {
+              if (!keyType.isStringValueType()) {
+                emit("/* warning: coerced from " + keyType + "*/");
```
mprobst: Could we print out a warning to the console?

mprobst: Nit: add a trailing space in `"*/*"`?

src/test/java/com/google/javascript/cl2dts/types.js
```diff
@@ -50,3 +50,12 @@ types.h = {a: 'a', b: 34};
...
 types.identity = function(a) { return a; };
+
+/** @type {Object<number, string>} */
+types.j = {a: 'a'};
+
+/**
+ * @param {Object<K,V>} obj The object in which to look for val.
+ * @template K,V
+ */
+types.containsValue = function(obj) { };
```
mprobst: Nit: in general, I think it'd be nicer to have tests that express some concept, instead of just being C&P from the Closure library. It'll make it easier to understand this tests in the future. E.g. here: `types.objectWithGenericKeyType` or so?

src/test/java/com/google/javascript/cl2dts/types.js
```diff
@@ -50,3 +50,12 @@ types.h = {a: 'a', b: 34};
  * @template T
  */
 types.identity = function(a) { return a; };
+
+/** @type {Object<number, string>} */
+types.j = {a: 'a'};
+
+/**
+ * @param {Object<K,V>} obj The object in which to look for val.
```
mprobst: This comment is a bit of a distraction, isn't it?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: indeed, `typeRegistry.getNativeType(JSTypeNative.OBJECT_TYPE)`, fixed here and for Array above

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: we have a test for that already in the `types.js`:
```
/** @type {Object} */
types.d = {};
```

Note this code is in `caseTemplatizedType`

src/test/java/com/google/javascript/cl2dts/types.js
alexeagle: yeah, I've been in a hurry here trying to get a demo working. agree these names would be good documentation

src/test/java/com/google/javascript/cl2dts/types.js
alexeagle: done

PR #57, opened 2015-08-28
fix(bug): allow nullable parameterized types

### LINKED ISSUE #56: generic type param missing from signature
```
* @param {function(this:SCOPE)|{handleEvent:function()}|null} listener Function
 *     or object that has a handleEvent method.
 * @param {number=} opt_delay Milliseconds to wait; default is 0.
 * @param {SCOPE=} opt_handler Object in whose scope to call the listener.
 * @return {number} A handle to the timer ID.
 * @template SCOPE
 */
goog.Timer.callOnce = function(listener, opt_delay, opt_handler) {
```
```
static callOnce < SCOPE > (listener : ( ( ) => any | { handleEvent : ( ) => any } ) , opt_delay ? : number , opt_handler ? : ( ) ) : number ;
```

the type of opt_handler is `( )` which is not valid

alexeagle:
fixes #56

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -654,7 +654,7 @@ public class DeclarationGenerator {
         public boolean apply(JSType input) {
           // Skip - TypeScript does not have explicit null or optional types.
           // Optional types must be handled at the declaration name (`foo?` syntax).
-          return !input.isNullable() && !input.isVoidType();
+          return !input.isNullType() && !input.isVoidType();
         }
       });
       if (alts.size() == 1) {
--- a/src/test/java/com/google/javascript/cl2dts/types.d.ts
+++ b/src/test/java/com/google/javascript/cl2dts/types.d.ts
@@ -5,7 +5,7 @@ declare namespace ಠ_ಠ.cl2dts_internal.types {
   var d : Object ;
   var e : any [] ;
   class Foo < T > {
-    genericMethod < V , W > (t : T , v : V ) : any ;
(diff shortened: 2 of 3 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -629,7 +629,9 @@ private void visitUnionType(UnionType ut) {
         public boolean apply(JSType input) {
           // Skip - TypeScript does not have explicit null or optional types.
           // Optional types must be handled at the declaration name (`foo?` syntax).
-          return !input.isNullable() && !input.isVoidType();
+          return !input.isNullable() && !input.isVoidType()
```
mprobst: Could you add parenthesis so this doesn't rely on operator precedence?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -629,7 +629,9 @@ private void visitUnionType(UnionType ut) {
...
+              // Template types are always nullable, but we should not skip them
+              || input.isTemplateType();
```
mprobst: The intention of the code was that unions commonly contain `Foo|null`, and we'd like to filter out the `null` case. Maybe the Real Bug is that this should not be `!input.isNullable()` but rather `!input.isSpecialNullType()` (I'm making that up)? I.e. the difference between the type permitting/including `null`, and the type actually being the special `null` type?

alexeagle: Yup, just isNullType() looks pretty good to me.

PR #58, opened 2015-08-28
fix(bug): don't print enum declarations as properties

### LINKED ISSUE #43: Provided enum static emit
browserevent.js
```
goog.provide('goog.events.BrowserEvent');
goog.provide('goog.events.BrowserEvent.MouseButton');
/**
 * @constructor
 * @extends {goog.events.Event}
 */
goog.events.BrowserEvent = function(opt_e, opt_currentTarget) {
}
/**
 * Normalized button constants for the mouse.
 * @enum {number}
 */
goog.events.BrowserEvent.MouseButton = {
  LEFT: 0,
  MIDDLE: 1,
  RIGHT: 2
};
```

we emit
```
  class BrowserEvent extends Event {
    isButton (a : BrowserEvent.MouseButton ) : boolean ;
    static MouseButton : BrowserEvent.MouseButton ;
  }
```

should be something like (if valid)

```
declare namespace goog.events {
class BrowserEvent {
  isButton (a: goog.events.BrowserEvent.MouseButton) {
}

declare namespace goog.events.BrowserEvent {
enum MouseButton {
 }
}
```

alexeagle:
Fixes #43

### DIFF
--- a/src/test/java/com/google/javascript/cl2dts/static_provided.d.ts
+++ b/src/test/java/com/google/javascript/cl2dts/static_provided.d.ts
@@ -1,0 +1,26 @@
+declare namespace ಠ_ಠ.cl2dts_internal.a.b.StaticHolder {
+  type AnEnum = number ;
+  var AnEnum : {
+    Y : AnEnum ,
+    X : AnEnum ,
+  };
+}
+declare module 'goog:a.b.StaticHolder.AnEnum' {
+  import alias = ಠ_ಠ.cl2dts_internal.a.b.StaticHolder.AnEnum;
+  export default alias;
+}
+declare namespace ಠ_ಠ.cl2dts_internal.a.b {
+  class StaticHolder {
+  }
+}
+declare module 'goog:a.b.StaticHolder' {
+  import alias = ಠ_ಠ.cl2dts_internal.a.b.StaticHolder;
+  export default alias;
+}
+declare namespace ಠ_ಠ.cl2dts_internal.a.b.StaticHolder {
+  function aFunction ( ) : boolean ;
+}
+declare module 'goog:a.b.StaticHolder.aFunction' {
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/cl2dts/static_provided.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Normalized button constants for the mouse.
+ * @enum {number}
+ */
+goog.events.BrowserEvent.MouseButton = {
+  LEFT: 0,
+  MIDDLE: 1,
+  RIGHT: 2
+};
```
rkirov: Since the test is called static_provided, what about adding a static function that is also provided (I think it should just work).

alexeagle: Nope actually adding that exposes another bug. We get a duplicate def'n of
the static method, inside the class as well as a namespace matching the
class name. Needs more work.

alexeagle: Reworked a bit, now doing the Right Thing - PTAL

rkirov: nice. still LGTM.

src/test/java/com/google/javascript/cl2dts/static_provided.d.ts
```diff
@@ -0,0 +1,27 @@
+declare namespace ಠ_ಠ.cl2dts_internal.goog.events.BrowserEvent {
```
mprobst: Same comment as in other CL: in the interest of future understandability, please consider making the names/types explain what the tested use case it, not just C&P from Closure library.

alexeagle: done

PR #62, opened 2015-08-31
fix(bug): add parens when needed for vararg types

### LINKED ISSUE #61: writing spread args without array type
```
/**
 * Creates the composition of the functions passed in.
 * For example, (goog.functions.compose(f, g))(a) is equivalent to f(g(a)).
 * @param {function(...?):T} fn The final function.
 * @param {...Function} var_args A list of functions.
 * @return {function(...?):T} The composition of all inputs.
 * @template T
 */
goog.functions.compose = function(fn, var_args) {}
```

`function compose < T > (fn : ( ...a : any [] ) => T ,  ...var_args : ( ...a : any [] ) => any [] ) : ( ...a : any [] ) => T ;`

error:
`/ac.jslib.d.ts:194:NaN: A rest parameter must be of an array type.`

alexeagle:
Fixes #61


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -451,10 +451,17 @@ private void visitEnumType(EnumType type) {
       emitBreak();
     }
 
-    private void visitTypeDeclaration(JSType type) {
+    private void visitTypeDeclaration(JSType type, boolean isVarArgs) {
       if (type != null) {
         emit(":");
-        visitType(type);
+        if (isVarArgs && !type.isAllType() && !type.isUnknownType()) {
```
rkirov: don't we output 'any' for all type and unknownType, what is wrong with `args :(any)[]`?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -734,10 +741,7 @@ private void visitFunctionParameters(FunctionType ftype) {
         if (param.isOptionalArg()) {
           emit("?");
         }
-        visitTypeDeclaration(param.getJSType());
-        if (param.isVarArgs()) {
-          emit("[]");
```
rkirov: why not keep the emit here and add the parens?

```
emit("(");
visit(...);
emit(")[]");
```

mprobst:
I think I'll leave this to @rkirov :-)

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: It just looks wacky to surround everything with extra parens. I'm hoping to emit fewer parens if possible without making our code too complex

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: because visitTypeDeclaration emits the colon, so if I need a paren between the colon and the start of the type expression, then I have to inline visitTypeDeclaration here instead of calling it

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: True, the downside is that there is implicit assumption that type.isAllType and type.isUnknownType's output is safe to emit without parens, which is certainly true now, but might change in the future.

I don't feel strongly here, so keep as it is.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: fair point, missed that.

PR #63, opened 2015-08-31
fix(bug): add missing generic type parameters

### LINKED ISSUE #60: TS enforces generic type params are present; closure doesnt
Lots of instance of this `javascript/closure/ui/ac/ac.jslib.d.ts:1409:NaN: Generic type 'EventId<T>' requires 1 type argument(s).`
and more:
`javascript/closure/ui/ac/ac.jslib.d.ts:606:NaN: Generic type 'EventHandler<SCOPE>' requires 1 type argument(s).`
`javascript/closure/ui/ac/ac.jslib.d.ts:1593:NaN: Generic type 'Promise<TYPE, RESOLVER_CONTEXT>' requires 2 type argument(s).`
`javascript/closure/ui/ac/ac.jslib.d.ts:1736:NaN: Generic type 'Array<T>' requires 1 type argument(s).`
etc.

We might be able to determine that the number of type arguments isn't correct by looking at the type the arguments are passed to, and replace the missing ones with 'any'.
Given the number of occurrences (34 in this particular compile) I don't think we want to try to fix all the usage sites.

alexeagle:
Fixes #60

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -13,6 +13,10 @@ import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
+import static com.google.common.collect.Iterables.any;
+import static com.google.javascript.rhino.jstype.JSTypeNative.OBJECT_TYPE;
+
+import com.google.common.collect.UnmodifiableIterator;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
 import com.google.javascript.jscomp.CommandLineRunner;
@@ -350,7 +354,7 @@ public class DeclarationGenerator {
           emit("extends");
           Iterator<ObjectType> it = ftype.getExtendedInterfaces().iterator();
           while (it.hasNext()) {
-            emit(getRelativeName(it.next()));
+            visitType(it.next());
             if (it.hasNext()) {
               emit(",");
             }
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -483,6 +487,11 @@ public Void caseStringType() {
 
         @Override
         public Void caseObjectType(ObjectType type) {
+          // Closure doesn't require that all the type params be declared, but TS does
```
rkirov: How about a type with 2 type params, but only one provided? Is that valid closure?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -515,8 +524,10 @@ public Void caseTemplatizedType(TemplatizedType type) {
           }
           emit(templateTypeName);
           emit("<");
-          for (JSType tmplType : type.getTemplateTypes()) {
-            visitType(tmplType);
+          UnmodifiableIterator<JSType> it = type.getTemplateTypes().iterator();
+          while (it.hasNext()) {
+            visitType(it.next());
+            if (it.hasNext()) emit(",");
```
rkirov: why not use visitTemplateTypes? I feel we have the same logic in two places, because you already fixed adding ',' for multiple template params.

alexeagle: these are two different things - one is a declaration of a parameterized type (T) and the other is providing a type parameter (T => string).
I tried merging these together in one function, but it is pretty awkward since for the first case you need `getTemplateTypesMap().getTemplateKeys()` which is an `Iterable<TemplateType>` and the other you need `getTemplateTypes()` which is an `Iterable<JSType>` which just yields a string, not something we can visit.
I think it's just awkward closure API showing up.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: added a test case for that, and split these tests from `types.js` into new `generics.js`

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: Actually, this is very reasonable explanation - type var declaration vs passing concrete type arguments. Thanks!

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: Nice, add at least one line into generics_usage.ts

alexeagle: done

alexeagle:
merged as 30b3968

PR #65, opened 2015-08-31
fix(bug): allow @dict with no other type info

### LINKED ISSUE #64: Handle @dict
```
 * @const
 * @dict
 * @private
 */
goog.events.onStringMap_ = {};
```
produces
```
var onStringMap_ : null ;
```

alexeagle:
Fixes #64

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -9,6 +9,7 @@ import static com.google.javascript.rhino.jstype.JSTypeNative.ARRAY_TYPE;
 import static com.google.javascript.rhino.jstype.JSTypeNative.OBJECT_TYPE;
 
 import com.google.common.base.Function;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
@@ -196,7 +197,11 @@ public class DeclarationGenerator {
         if (desiredSymbols.contains(otherName) && other.getType() != null
             && !other.getType().isFunctionPrototypeType()
             && !isPrototypeMethod(other)) {
-          treeWalker.walk(other, false);
+          try {
+            treeWalker.walk(other, false);
+          } catch (RuntimeException e) {
+            throw new RuntimeException("Failed to emit for " + other, e);
+          }
         }
       }
     }
@@ -267,6 +272,7 @@ public class DeclarationGenerator {
   }
 
   private void emit(String str) {
+    Preconditions.checkNotNull(str);
     if (!maybeEmitIndent()) {
       out.write(" ");
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -490,6 +496,8 @@ public Void caseStringType() {
         public Void caseObjectType(ObjectType type) {
           if (type.isRecordType()) {
             visitRecordType((RecordType) type);
+          } else if (type.isDict()) {
+            emit("{[s: string]: any}");
```
rkirov: s -> `key or prop`

src/test/java/com/google/javascript/cl2dts/dict.js
```diff
@@ -0,0 +1,13 @@
...
+ * @const
+ * @dict
+ */
+dict.untyped = {};
+
+/**
+ * @dict
+ * @type {{a: Function}}
+ */
+dict.typed = {'a': function() {}};
```
rkirov: How about one with @constructor too 
```
/**
 * @constructor
 * @dict
 */
function Foo() {}
var obj1 = new Foo();
```
taken from https://developers.google.com/closure/compiler/docs/js-for-compiler?hl=en

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -490,6 +496,8 @@ public Void caseStringType() {
         public Void caseObjectType(ObjectType type) {
           if (type.isRecordType()) {
             visitRecordType((RecordType) type);
+          } else if (type.isDict()) {
```
mprobst: Is it possible to have a `@dict` that still has constructor properties?

```js
/** @dict @constructor */
X = function() {};
X.prototype.foo = function() {};
```

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: changed to `key`

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: yes, I added a test case for that, and the closure compiler is happy to produce such a thing.
given the lack of a spec, that means it Must Be Valid Closure

src/test/java/com/google/javascript/cl2dts/dict.js
alexeagle: done, and added a usage for it.
note we don't have to put the bracket accessor in the .d.ts, it can just be a bare class

src/test/java/com/google/javascript/cl2dts/dict_usage.ts
```diff
@@ -0,0 +1,4 @@
+///<reference path="./dict"/>
+import {DictClass} from 'goog:dict';
+var d = new DictClass();
+var s: string = d['thing'];
```
rkirov: add usage for ClassWithDottedProperties

rkirov:
Add usage for `add usage for ClassWithDottedProperties`, otherwise LGTM.

src/test/java/com/google/javascript/cl2dts/dict_usage.ts
alexeagle: done

alexeagle:
merged as 4036ae1

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
mprobst: But in that case we fail to produce a member for `foo` here, right? I'll file an issue.

PR #67, opened 2015-08-31
fix(bug): handle untyped object

### LINKED ISSUE #66: @const with no other type info
```
* @const
 * @private
 */
goog.userAgent.isVersionOrHigherCache_ = {};
```
emits a literal null (or fails with NPE with error-checking)

alexeagle:
fixes #66

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -513,8 +513,10 @@ public class DeclarationGenerator {
             visitRecordType((RecordType) type);
           } else if (type.isDict()) {
             emit("{[key: string]: any}");
-          } else {
+          } else if (type.getReferenceName() != null) {
             emit(getRelativeName(type));
+          } else {
+            emit("Object");
           }
           return null;
         }
--- a/src/test/java/com/google/javascript/cl2dts/types.js
+++ b/src/test/java/com/google/javascript/cl2dts/types.js
@@ -39,3 +39,15 @@ types.j = {a: 'a'};
  * @template K,V
  */
 types.objectWithGenericKeyType = function(obj) { };
+
+//!! marked @const to appear in `compiler.getTopScope().getAllSymbols()`
+/**
+ * @const
+ */
+types.inferredobj = {};
+
+//!! marked @const to appear in `compiler.getTopScope().getAllSymbols()`
+/**
+ * @const
+ */
+types.inferrednum = 1;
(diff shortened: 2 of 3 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -490,8 +490,10 @@ public Void caseStringType() {
         public Void caseObjectType(ObjectType type) {
           if (type.isRecordType()) {
             visitRecordType((RecordType) type);
-          } else {
+          } else if (type.getReferenceName() != null) {
             emit(type.getReferenceName());
+          } else {
+            emit("Object");
```
rkirov: I think we need 'any' as the fallback. `Object` is very restrictive.

src/test/java/com/google/javascript/cl2dts/types.js
```diff
@@ -59,3 +59,13 @@ types.j = {a: 'a'};
  * @template K,V
  */
 types.objectWithGenericKeyType = function(obj) { };
+
+/**
+ * @const
```
rkirov: is const needed for type inference?

alexeagle: Without const, the symbol doesn't appear in the topScope (I assume because it gets removed because it's unused). @export would work just as well

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: We are inside of caseObjectType here, so I don't think it's necessary to widen to `any`. Maybe there's a test case I'm missing though

src/test/java/com/google/javascript/cl2dts/types.js
mprobst: Maybe add a comment explaining that, or at least documenting that this was necessary?

alexeagle: Well, then I don't understand this very well. I tried `@export` and that _doesn't_ appear in `compiler.getTopScope().getAllSymbols()` but I don't know why.
For now, I've added the comment explaining why `@const` is here.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: true.

src/test/java/com/google/javascript/cl2dts/types.js
rkirov: ah, the mysteries of closure. Good enough for now.

PR #70, opened 2015-09-01
fix(bug): look at parents of provides when trimming desired symbols f…

### LINKED ISSUE #68: Duplicate variables in high-level namespace
Get lots of Duplicate identifier still.

```
/javascript/closure/ui/ac/ac.jslib.d.ts:3202:NaN: Duplicate identifier 'fx'.
/javascript/closure/ui/ac/ac.jslib.d.ts:3220:NaN: Duplicate identifier 'fx'.
```

alexeagle:
…or a namespace.

fixes #68

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -192,8 +192,15 @@ public class DeclarationGenerator {
       for (String property : objType.getPropertyNames()) {
         desiredSymbols.add(symbol.getName() + "." + property);
       }
-      // These will be emitted in their own default-exported namespace
-      desiredSymbols.removeAll(provides);
+      // Any provides have their own namespace and should not be emitted in this namespace.
+      for (String provide : provides) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -188,8 +188,14 @@ private void declareNamespace(String namespace, TypedVar symbol, boolean isDefau
       for (String property : objType.getPropertyNames()) {
         desiredSymbols.add(symbol.getName() + "." + property);
       }
-      // These will be emitted in their own default-exported namespace
-      desiredSymbols.removeAll(provides);
+      // Any provides have their own namespace and should not be emitted in this namespace.
```
rkirov: `Any provides (and their implicit parent namespaces)` to explain why we not only handle a provide like `a.b.c.` but also `a.b` and `a`.

alexeagle: done

alexeagle:
merged as 7483c91

PR #73, opened 2015-09-01
fix(bug): handle nested classes and enums.

### LINKED ISSUE #69: handle enums that are explicitly provided and within a provided namespace.
The following pattern while highly suspicious is valid Closure:

```javascript
goog.provide('goog.A');
goog.provide('goog.A.B');

/** @constructor */
goog.A = function() {}

/** @enum */
goog.A.B = {};
```
The current code will create a `var` for B in `goog.A`, and also `declare namespace goog.A.B` which is an error.

see: https://github.com/google/closure-library/blob/master/closure/goog/fx/transition.js

### LINKED ISSUE #71: Self-referencing classes
in provide_single_class.js
```
/**
 * @param {foo.bar.Baz} b
 * @return {boolean}
 */
foo.bar.Baz.prototype.equals = function(b) {
  return false;
};
```

emits
```
declare namespace ಠ_ಠ.cl2dts_internal.foo.bar {
  class Baz {
    field : string ;
    equals (b : foo.bar.Baz ) : boolean ;
    method (a : string ) : number ;
    static staticMethod (a : string ) : number ;
  }
}
```

the Baz name is not correctly qualified, should probably be bare `Baz` or maybe fully-qualified with the look of disapproval

alexeagle:
fixes #71
fixes #69

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -31,7 +31,6 @@ import com.google.javascript.rhino.jstype.EnumElementType;
 import com.google.javascript.rhino.jstype.EnumType;
 import com.google.javascript.rhino.jstype.FunctionType;
 import com.google.javascript.rhino.jstype.JSType;
-import com.google.javascript.rhino.jstype.JSTypeNative;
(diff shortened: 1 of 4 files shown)


rkirov:
We paired on this one, LGTM.

PR #75, opened 2015-09-02
feature(types): don't expose private properties

### LINKED ISSUE #72: don't expose methods marked @private
these aren't intended to be advertised in the API

alexeagle:
Fixes #72

### DIFF
--- a/src/test/java/com/google/javascript/cl2dts/provide_single_class.js
+++ b/src/test/java/com/google/javascript/cl2dts/provide_single_class.js
@@ -32,6 +32,11 @@ foo.bar.Baz.prototype.equals = function(b) {
   return false;
 };
 
+/**
+ * @private
+ */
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/cl2dts/provide_single_class.js
```diff
@@ -32,6 +32,11 @@ foo.bar.Baz.prototype.equals = function(b) {
   return false;
 };
 
+/**
+ * @private
+ */
+foo.bar.Baz.prototype.thisIsPrivate_ = function() {};
```
rkirov: add @private and @constructor to match the closure lib case. Easy to do since no usage needed :)

alexeagle: added

PR #84, opened 2015-09-03
fix(dict): emit the property access explicitly in @dict classes,

### LINKED ISSUE #82: `@dict` constructors with additional properties
Currently the closure code below produces (AFAICT) only `{[k: string]: any}`. We should include all other defined constructor properties in the generated type, and only include the property accessor on top of that.

E.g.:

```js
/** @dict @constructor */
X = function() {};
X.prototype.foo = function() {};
```

Should be:
```ts
class X {
  [k: string]: any;
  foo();
}
```

alexeagle:
even though the usage is allowed without it.

Fixes #82.


mprobst:
LGTM

alexeagle:
merged as 38f2049

PR #92, opened 2015-09-08
fix(bug): surround union alternatives with parens when they contain s…

### LINKED ISSUE #88: @typedef with union type
```
/**
 * A message value that can be handled by a Logger.
 *
 * Functions are treated like callbacks, but are only called when the event's
 * log level is enabled. This is useful for logging messages that are expensive
 * to construct.
 *
 * @typedef {string|function(): string}
 */
goog.debug.Loggable;

/**
 * Logs a message. If the logger is currently enabled for the
 * given message level then the given message is forwarded to all the
 * registered output Handler objects.
 * @param {goog.log.Logger} logger
 * @param {goog.log.Level} level One of the level identifiers.
 * @param {goog.debug.Loggable} msg The message to log.
 * @param {Error|Object=} opt_exception An exception associated with the
 *     message.
 */
goog.log.log = function(logger, level, msg, opt_exception) {
  if (goog.log.ENABLED && logger) {
    logger.log(level, msg, opt_exception);
  }
};
```

emits a bad syntax `( )`
```
function log (logger : goog.debug.Logger , level : goog.debug.Logger.Level , msg : ( string | ( ) => string ) , opt_exception ? : Object ) : void ;
```

alexeagle:
…paces

fixes #88

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -549,14 +549,37 @@ public class DeclarationGenerator {
     private void visitTypeDeclaration(JSType type, boolean isVarArgs) {
       if (type != null) {
         emit(":");
-        if (isVarArgs && !type.isAllType() && !type.isUnknownType()) {
-          emit("(");
-          visitType(type);
-          emit(")");
+        // From https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#a-grammar
+        // ArrayType:
+        //   PrimaryType [no LineTerminator here] [ ]
+        if (isVarArgs) {
+          visitTypeAsPrimary(type);
         } else {
           visitType(type);
         }
         if (isVarArgs) emit("[]");
+      }
+    }
+
+    /**
+     * Adds parentheses to turn a Type grammar production into a PrimaryType.
+     * See https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#a-grammar
+     *
+     * Avoid adding extra parens where the type is already known to be Primary.
+     *
+     * PrimaryType:
+     *   ParenthesizedType
+     * ParenthesizedType:
+     *   ( Type )
+     */
+    private void visitTypeAsPrimary(JSType type) {
+      // These types will produce a non-primary grammar production
+      if (type.isConstructor() || type.isFunctionType() || type.isUnionType()) {
+        emit("(");
+        visitType(type);
+        emit(")");
+      } else {
+        visitType(type);
       }
     }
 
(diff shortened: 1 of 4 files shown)


mprobst:
So I presume this is because operator precedence in TypeScript's union types between `|` and something else doesn't match. What exactly doesn't match? Looking for whitespace seems like a hack that will bite us sooner or later.

alexeagle:
The two examples I know of are
string | () => any
and
() => any []

we could search the string for the arrow operator rather than space, that
solves all the cases I know of.
Looking at the spec for Arrow Function
https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#4.11
they mention precedence briefly but I don't see a grammar definition for
all the operators and their precedence.

mprobst:
The grammar that also specifies operator precedence through the rule ordering is here:
https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#a-grammar

You should be able to follow down from the primary type expression to see what is legal and what needs parentheses. The `Type` rule itself seems to be of particular interest, note how it's branching into `UnionOrIntersectionType` and `FunctionType` at that level, and note what rules are allowed within a `UnionType`.

alexeagle:
okay, I studied the grammar for a while, and I don't think I can precisely map that spec into an emitter. But, I do understand now why the ArrayType and UnionType need a PrimaryType rather than a non-PrimaryType which includes FunctionType and ConstructorType. So I think this is correct.

I added some comments explaining where these grammar productions come from, and changed to search the nested emit for the arrow (FunctionType token) or the ampersand (IntersectionType token).

If you want to comb through the grammar more precisely, then I need some help with that.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -560,6 +561,32 @@ private void visitTypeDeclaration(JSType type, boolean isVarArgs) {
       }
     }
 
+    /**
+     * Add parentheses to turn a Type grammar production into a PrimaryType.
```
mprobst: 3rd party `Adds parentheses...`

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -560,6 +561,32 @@ private void visitTypeDeclaration(JSType type, boolean isVarArgs) {
...
+     *
+     * PrimaryType:
+     *   ParenthesizedType
+     * ParenthesizedType:
+     *   ( Type )
+     */
+    private void visitTypeAsPrimary(JSType type) {
+      // These tokens indicate that the type is non-primary, because it is one of:
+      // UnionType
+      // IntersectionType
```
mprobst: Not `IntersectionType`, right?

mprobst:
LGTMs with one docs/language nit.

PR #93, opened 2015-09-09
use reflection to load default externs.

alexeagle:
We need to avoid CommandLineRunner because it's a naughty dependency in google3.
Reflection is a last resort, other options are:
- make a change in closure compiler to move the getDefaultExterns to AbstractCommandLineRunner, which Tyler said is okay, then wait for a release
- make the case that we want to depend on open_source_jscomp despite their concerns
- copy/paste the code to read externs.zip entries and turn into SourceFile
- remove our ability to understand any default libraries and require all externs be explicitly passed in

Those are all reasonable options, but this change is the easiest.

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -17,7 +17,6 @@ import com.google.common.collect.Iterables;
 import com.google.common.io.Files;
 import com.google.javascript.jscomp.BasicErrorManager;
 import com.google.javascript.jscomp.CheckLevel;
-import com.google.javascript.jscomp.CommandLineRunner;
 import com.google.javascript.jscomp.Compiler;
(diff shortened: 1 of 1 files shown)


alexeagle:
More context: this is needed to be in third_party/java_src

mprobst:
Well, this maintains the same dependency, except worse, because now it's only visible at runtime, right? So this does not seem like an improvement of the situation.

What about making that change to move the method, and then depending on `//third_party/java_src/jscomp:head` until the (weekly?) release?

alexeagle:
This does fix the problem because we don't enter this conditional when
running internally. Previously we weren't classloading it because we used a
jar dependency.
I'll see about moving it.

alexeagle:
Also, since we depend on this in our open-source repo we'll need to wait
more like two weeks to get a release.

alexeagle:
While we wait for a release, an easy fix is to make a local modification in the checked in version that returns an empty list here.

alexeagle:
not pursuing this

PR #96, opened 2015-09-10
fix(interface): statics on interfaces emitted in a namespace.

### LINKED ISSUE #78: duplicate output for interface with nested enum
see goog.fx.Transition (interface) and goog.fx.Transition.EventType (enum).

Output is 
```javascript
declare namespace ಠ_ಠ.cl2dts_internal.goog.fx {                                                      
  interface Transition {                                                                             
    play : any ;                                                                                     
    stop : any ;                                                                                     
  }                                                                                                  
  var Transition : { EventType : Transition.EventType }                                              
}  

declare namespace ಠ_ಠ.cl2dts_internal.goog.fx.Transition {...}
```

I think it is safe to assume that nested classes, enums, and interfaces will have their own goog.provide and plainly skip them from the properties of the parent classes, enums, interfaces.

In this case Transition.EventType is provided explicitly.

rkirov:
We emit the static methods and variable in a namespace, instead of a var
union type, as done previously. That way we do not interfere with any
enums being output.

Closes #78

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -241,14 +241,7 @@ public class DeclarationGenerator {
 
   private void declareNamespace(String namespace, TypedVar symbol, boolean isDefault,
       Compiler compiler, Set<String> provides) {
-    emitNoSpace("declare namespace ");
-    emitNoSpace(INTERNAL_NAMESPACE);
-    if (!namespace.isEmpty()) {
-      emitNoSpace("." + namespace);
-    }
-    emitNoSpace(" {");
-    indent();
-    emitBreak();
+    emitNamespaceBegin(namespace);
     TreeWalker treeWalker = new TreeWalker(namespace, compiler.getTypeRegistry(), provides);
     if (isDefault) {
       treeWalker.walk(symbol, true);
@@ -287,6 +280,29 @@ public class DeclarationGenerator {
         }
       }
     }
+    emitNamespaceEnd();
+    // An extra pass is required for default exported interfaces, because in Closure they might have
+    // static methods. TS does not support static methods on interfaces, thus we create a new namespace for them.
+    if (isDefault && isInterfaceWithStatic(symbol.getType())) {
+      emitNamespaceBegin(symbol.getName());
+      treeWalker.walkDefaultInterface((FunctionType) symbol.getType());
+      emitNamespaceEnd();
+    }
+  }
+
+
+  private void emitNamespaceBegin(String namespace) {
+    emitNoSpace("declare namespace ");
+    emitNoSpace(INTERNAL_NAMESPACE);
+    if (!namespace.isEmpty()) {
+      emitNoSpace("." + namespace);
+    }
+    emitNoSpace(" {");
+    indent();
+    emitBreak();
+  }
+
+  private void emitNamespaceEnd() {
     unindent();
     emit("}");
     emitBreak();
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -287,11 +280,39 @@ private void declareNamespace(String namespace, TypedVar symbol, boolean isDefau
         }
       }
     }
+    emitNamespaceEnd();
+    // An extra pass is required for default exported interfaces, because in Closure they might have
+    // static methods. TS does not support static methods on interfaces, thus we create a new namespace for them.
+    if (isDefault && isInterfaceWithStatic(symbol.getType())) {
+      emitNamespaceBegin(symbol.getName ());
```
alexeagle: nit: extra space

src/test/java/com/google/javascript/cl2dts/interface.js
```diff
@@ -11,3 +12,9 @@ interface_exp.staticMethod = function() {return interface_exp.staticProp;};
 
 /** @type {number} */
 interface_exp.staticProp = 0;
+
+/** @enum {number} */
+interface_exp.SomeEnum = {
+  A: 1,
+  B: 2
+};
```
alexeagle: trailing newline

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -287,11 +280,39 @@ private void declareNamespace(String namespace, TypedVar symbol, boolean isDefau
...
+      treeWalker.walkDefaultInterface((FunctionType) symbol.getType());
+      emitNamespaceEnd();
+    }
+  }
+
+  private boolean isInterfaceWithStatic(JSType type) {
```
alexeagle: opt: relocate the method near the other predicates?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -287,11 +280,39 @@ private void declareNamespace(String namespace, TypedVar symbol, boolean isDefau
...
+    return any(ftype.getOwnPropertyNames(), not(equalTo("prototype")));
+  }
+
+  private void emitNamespaceEnd() {
     unindent();
     emit("}");
     emitBreak();
   }
 
+  private void emitNamespaceBegin(String namespace) {
```
alexeagle: begin before end

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -499,6 +507,21 @@ private void walk(TypedVar symbol, boolean isDefault) {
...
+    private void walkDefaultInterface(FunctionType ftype) {
+      for (String propName : ftype.getOwnPropertyNames()) {
+        if (propName.equals("prototype")) continue;
+        JSType pType = ftype.getPropertyType(propName);
+        // Here we assume the enum is exported and handled separately.
+        if (pType.isEnumType()) continue;
+        emit("var");
+        emit(propName);
+        emit(":");
+        visitType(pType);
```
alexeagle: this was visitRecordLikeType before - changed intentionally?

alexeagle:
LGTM

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: Yes, the output goes from `var recordLike = {foo: A, bar: B}` to 
```
namespace NamespaceLike {
  var foo: A;
  var bar: B;
}
```

I am also removing visitRecordLike since it is not needed any longer. `walkDefaultInterface` can become `visitNamespaceLike` in case we find more usage for it.

PR #98, opened 2015-09-11
fix(enums): support cross-referenced enums

### LINKED ISSUE #77: do not output qualified names on the LHS of type declaration
Currently we output type aliases like this:
```
type goog.fx.TransitionBase.State = number ;
```

Will need to check enum branch and type alias branch.

rkirov:
Closure allows for the following 'enum-aliasing' syntax:

/** @enum */
foo.A = {...};

/** @enum */
foo.B = foo.A;

This commit generates correct output for that case.

Closes #77

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -425,12 +425,19 @@ public class DeclarationGenerator {
     }
 
     private String getUnqualifiedName(TypedVar symbol) {
-      String qualifiedName = symbol.getName();
-      int dotIdx = qualifiedName.lastIndexOf('.');
+      return lastPart(symbol.getName());
+    }
+
+    private String getUnqualifiedName(JSType type) {
+      return lastPart(type.getDisplayName());
+    }
+
+    private String lastPart(String input) {
+      int dotIdx = input.lastIndexOf('.');
       if (dotIdx == -1) {
-        return qualifiedName;
+        return input;
       }
-      return qualifiedName.substring(dotIdx + 1, qualifiedName.length());
+      return input.substring(dotIdx + 1, input.length());
     }
 
     private void walk(TypedVar symbol, boolean isDefault) {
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/cl2dts/nested_enum_usage.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference path="./nested_enums"/>
+import C from 'goog:nested.bar.Baz';
+import NE from 'goog:nested.bar.Baz.NestedEnum';
+import SuperC from 'goog:nested.bar.SuperBaz';
+import NE2 from 'goog:nested.bar.SuperBaz.NestedEnum';
+
+var a: NE = NE.A;
+var a2: NE2 = NE2.A;
+// Because the twp enums are just type aliases of the same underlying type, they are assignable to
```
alexeagle: nit: spelling 'two'

src/test/java/com/google/javascript/cl2dts/nested_enum_usage.ts
```diff
@@ -0,0 +1,11 @@
...
+// each other. This matches Closure's semantics.
+a = a2;
```
alexeagle: newline

src/test/java/com/google/javascript/cl2dts/nested_enums.js
```diff
@@ -0,0 +1,24 @@
...
+ * @constructor
+ * @extends {nested.bar.Baz}
+ */
+nested.bar.SuperBaz = function() {};
+
+/** @enum */
+nested.bar.SuperBaz.NestedEnum = nested.bar.Baz.NestedEnum;
+
+/** @type {nested.bar.SuperBaz.NestedEnum} */
+var a = nested.bar.Baz.NestedEnum.A;
```
alexeagle: newline

src/test/java/com/google/javascript/cl2dts/nested_enums.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+/**
+ * @constructor
+ * @extends {nested.bar.Baz}
+ */
+nested.bar.SuperBaz = function() {};
+
+/** @enum */
+nested.bar.SuperBaz.NestedEnum = nested.bar.Baz.NestedEnum;
```
alexeagle: I suppose there is also the case of aliasing enum from a source that isn't a supertype, but you naturally handle that as well?

alexeagle:
LGTM

src/test/java/com/google/javascript/cl2dts/nested_enums.js
rkirov: True, I wasn't sure what is the issue at first, so I faithfully reproduced the closure setup I saw. But indeed the real issue is aliased enums, so I will remove the superclassing and rename the test to `aliased_enums`.

PR #99, opened 2015-09-16
do not emit static apply and call on functions.

### LINKED ISSUE #76: subclasses with different constructor params break static call method gen
For example if B extends A and takes string instead of number, we generate:

```javascript
declare namespace f {
	class A {
		constructor(n: number);
		static call(a: A, n: number);
	}

	class B extends A {
		constructor(s: string);
		static call(b: B, n: string);
	}
}
```

TS barfs on the static call for B which is required to subtype A's functional type. See 'oog.fx.dom.PredefinedEffect' and 'Animation' for closure lib repro.

Maybe we should not output `call` and `apply` for constructor functions until someone really needs them. I can't come up with a good need for using those in sane TS code.

rkirov:
They are still present for TS users, but provided by lib.d.ts.

Closes #76

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -890,6 +890,8 @@ public class DeclarationGenerator {
           continue;
         }
         if (isStatic) {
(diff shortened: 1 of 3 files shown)


alexeagle:
LGTM nice

PR #102, opened 2015-09-17
Depgraph support

alexeagle:
A depgraph file is a Bazel internal which represents the structure of the js_library rules and their dependencies. This is an explicit output of pinto_library with deps_mgmt = "closure", and is an implicit output of other js_* rules.
The structure of the depgraph file is JSON with three top-level collections:
- roots which are the files presented as srcs of this js_* rule
- nonroots which are the files that come from deps
- resource_roots (??)

So if we read the roots section, and only emit for goog.provides in those inputs, we will prune our output to contain only declarations of types for the srcs and not stuff in the deps.

This way we have a distinct, non-overlapping `.d.ts` file for each js_* rule.


alexeagle:
+ @mprobst as well since it's a significant and interesting change

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -89,10 +93,19 @@
         handler=StringArrayOptionHandler.class)
     List<String> externs = new ArrayList<>();
 
+    @Option(name = "--depgraphs",
+        usage = "file(s) which contains the JSON representation of the dependency graph",
+        metaVar = "file.depgraph...",
+        handler=StringArrayOptionHandler.class)
```
mprobst: nit indent

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -150,6 +163,35 @@ public static void main(String[] args) {
     System.exit(0);
   }
 
+  List<String> parseDepgraphRoots() {
```
mprobst: This looks like a good candidate for code to move to a separate file.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -150,6 +163,35 @@ public static void main(String[] args) {
...
+    List<String> result = new ArrayList<>();
+    for (String file : opts.depgraphFiles) {
+      try (FileReader reader = new FileReader(new File(file))) {
+        List<List> list = new Gson().fromJson(reader, new TypeToken<List<List>>(){}.getType());
```
mprobst: GSON is pretty neat at mapping into a Java class that represents the structure and fields. Ultimately we'll probably need that, would it make sense to just do that now? It'd also make this code more readable IMHO.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -150,6 +163,35 @@ public static void main(String[] args) {
...
+        for (List outer : list) {
```
mprobst: isn't the second list `inner`, if anything?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -150,6 +163,35 @@ public static void main(String[] args) {
...
+              String filename = (String) rootDescriptor.iterator().next();
+              // *-bootstrap.js are automatically added to every rule by Bazel
+              if (!filename.endsWith("-bootstrap.js")) {
+                result.add(filename);
+              }
+            }
+          }
+        }
+      } catch (FileNotFoundException e) {
+        throw new IllegalArgumentException("depgraph file not found: " + file);
```
mprobst: please always include the cause, it can be super helpful in debugging later on.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -211,12 +253,28 @@ protected void printSummary() {}
 
   public String produceDts(Compiler compiler) {
     Set<String> provides = new HashSet<>();
-    // TODO: only inspect inputs which were explicitly provided by the user
+    Set<String> transitiveProvides = new HashSet<>();
+    List<String> depgraphRoots = parseDepgraphRoots();
```
mprobst: I think it'd be nice to only optionally take the depgraph files, so that people might still run the tool outside of bazel. E.g. if you have some Google API that's declared as a Closure style externs file.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -211,12 +253,28 @@ protected void printSummary() {}
...
+      emit("// WARNING: depgraph contains no roots; all provides from all inputs will be walked");
+      emitBreak();
+    }
     for (CompilerInput compilerInput : compiler.getInputsById().values()) {
-      provides.addAll(compilerInput.getProvides());
+      transitiveProvides.addAll(compilerInput.getProvides());
+      if (depgraphRoots.isEmpty() ||
+          depgraphRoots.contains(compilerInput.getSourceFile().getOriginalPath())) {
+        provides.addAll(compilerInput.getProvides());
+        if (opts.debugOutput) {
```
mprobst: This seems like a useful thing to unconditionally output - or do you just want to ignore it for testing? We could strip it there in the test runner before comparing.

src/test/java/com/google/javascript/cl2dts/DepgraphTest.java
```diff
@@ -0,0 +1,35 @@
...
+  @Test
+  public void testParseFile() throws Exception {
+    List<File> testInputFiles = DeclarationGeneratorTests.getTestInputFiles(new FilenameFilter() {
+      @Override
+      public boolean accept(File dir, String name) {
+        return name.endsWith(".depgraph");
+      }
+    });
+    List<String> args = new ArrayList<>();
+    args.add("foo.js");
```
mprobst: hu? Maybe add a comment on what this does?

src/test/java/com/google/javascript/cl2dts/DepgraphTest.java
```diff
@@ -0,0 +1,35 @@
...
+    args.add("--depgraphs");
+    for (File testInputFile : testInputFiles) {
+      args.add(testInputFile.getAbsolutePath());
+    }
+    Options opts = new Options(args.toArray(new String[testInputFiles.size() + 2]));
+    List<String> depgraphRoots = new DeclarationGenerator(opts).parseDepgraphRoots();
```
mprobst: this looks like we're testing on the wrong level - we should restructure the code such that you can test parsing a depgraph file without setting up machinery by going through the main method with opts and everything.

src/test/java/com/google/javascript/cl2dts/RequireFromOtherFileTest.java
```diff
@@ -0,0 +1,56 @@
...
+        "--debug"}))
+        .generateDeclarations();
+    assertThat(out.toString(UTF_8.name())).doesNotContain("Stuff");
+  }
+
+  private File write(String filename, String... contentLines) throws IOException {
+    int dot = filename.lastIndexOf(".");
+    String base = filename.substring(0, dot);
+    String ext = filename.substring(dot + 1);
+    File newFile = File.createTempFile(getClass().getSimpleName() + "_" + base, ext);
```
mprobst: Doesn't JUnit 4 have some sort of a tempfolder test rule?

src/test/java/com/google/javascript/cl2dts/RequireFromOtherFileTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.PrintStream;
+
+public class RequireFromOtherFileTest {
+
+  // Repro for https://github.com/angular/closure-to-dts/issues/101
+  @Test
+  public void testRequireFromOtherFile() throws Exception {
+    File requireFile = write("require.js",
```
mprobst: Maybe call it "testTransitiveProvideNotEmitted"?

src/test/java/com/google/javascript/cl2dts/DepgraphTest.java
```diff
@@ -0,0 +1,35 @@
...
+    assertThat(depgraphRoots)
+        .containsExactly("my/thing/static/js/annotations/annotations-canvas-controller.js");
```
mprobst: For the end2end kind of test this could be, shouldn't we also compare the output `.d.ts` to an expected result?

This seems to be an end2end test from the setup and complexity for debugging, but a unit test from the assertion depth.

src/test/java/com/google/javascript/cl2dts/closure.depgraph
```diff
@@ -0,0 +1,148 @@
...
+    [
+      [
+        "javascript/closure/base.js",
+        [
+          [
+            "custom_annotations",
+            []
+          ],
+          [
+            "is_externs",
```
mprobst: I don't understand any of these flags and fields.

Could you cut it down to just the stuff that's relevant for our test? Same for the file names, I don't think any reader of this test case will understand what that annotations canvas thing is.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: 4 spaces for statement continuation, looks right to me

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: This file could be split up in a number of places, but it's quite convenient right now that there is a single copy command required to pull this into google3 for testing, so I'd prefer to continue growing it for now while we are under very active development

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: yup, good point

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: it is indeed optional, right below you can see that we emit a warning if it is empty and continue on, treating every input as a root

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: Yeah, I was on the fence about google3 users seeing this, but it might help them self-diagnose problems or at least bootstrap our investigation.
Added the stripping on the actual in the assertion

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
mprobst: yep but no indentation around the equals sign.

Don't you have an automated formatter?

src/test/java/com/google/javascript/cl2dts/RequireFromOtherFileTest.java
alexeagle: no longer using temp files in latest snapshot

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
mprobst: I think ts2dart showed us that it's better to split up earlier rather than later. This here would also make testing easier, and the depgraph parsing is really completely unrelated to the AST walking except for the narrow "what's provided" interface.

Shouldn't `cp -r` or even `rsync` also just result in a single command?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
mprobst: But why make it a warning? If the user didn't specify a depgraph, that's just what we do, right?

OTOH, if the user did specify a depgraph and we weren't able to parse anything out of it, we should given an error and fail.

src/test/java/com/google/javascript/cl2dts/closure.depgraph
alexeagle: done. we

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: oh, yeah I fixed that, I wouldn't have called spaces inside the text "indent" so I misunderstood you

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: changed to verify that provided depgraphs have non-empty roots and remove the warning

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: No, it's strictly harder, sometimes I just ctrl-A and then paste into my cider window.
But I'll break it up now. I'll do the Options class and maybe some others that I've been thinking about.

src/test/java/com/google/javascript/cl2dts/DepgraphTest.java
alexeagle: removed

src/test/java/com/google/javascript/cl2dts/DepgraphTest.java
alexeagle: changed to pure unit test

src/test/java/com/google/javascript/cl2dts/DepgraphTest.java
alexeagle: we have other tests that use this and cover any integration bugs

src/test/java/com/google/javascript/cl2dts/RequireFromOtherFileTest.java
alexeagle: renamed

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
alexeagle: I agree we may want that, but I'm not sure we'll need anything beyond roots from the file (for example Laszlo says the provides listed in there may be different from the source closure sees, and we want the latter.)
I'll leave a todo for myself for now, I'd rather spend time fixing some bugs first.

alexeagle: Oh, and for now I've created the Depgraph class so users see this as a Java structure, meaning that refactoring will be local to the file

src/main/java/com/google/javascript/cl2dts/Depgraph.java
```diff
@@ -0,0 +1,60 @@
...
+    if (fileContents.isEmpty()) {
+      return new Depgraph(Collections.<String>emptyList());
+    }
+
+    List<String> roots = new ArrayList<>();
+    for (String depgraph : fileContents) {
+      try {
+        List<List> list = new Gson().fromJson(depgraph, new TypeToken<List<List>>(){}.getType());
+        for (List outer : list) {
+          Iterator i = outer.iterator();
```
mprobst: Could you add generic types for all of these?

src/main/java/com/google/javascript/cl2dts/Options.java
```diff
@@ -0,0 +1,95 @@
...
+      throw new CmdLineException(parser, "No files were given");
+    }
+  }
+
+  Options(boolean skipParseExterns) {
+    this.skipParseExterns = skipParseExterns;
+  }
+
+  public List<String> readDepgraphs() {
+    if (depgraphs != null) {
```
mprobst: AFAICT this only happens in tests, the actual code doesn't need the caching. Maybe refactor to avoid that?

Also, could we build this such that the options class is immutable once we parsed everything? Seems odd to have this change state and hold onto file contents.

src/test/java/com/google/javascript/cl2dts/nullable.js
```diff
@@ -3,15 +3,15 @@ goog.provide('nullable');
 /** @type {?boolean} */
 nullable.w = true;
 
-//!! Explicitly non-nullable.
+// Explicitly non-nullable.
```
mprobst: Do we need these changes?

alexeagle: I think it's misleading to use the special syntax in a place we don't actually strip anything

src/test/java/com/google/javascript/cl2dts/MultiFileTest.java
```diff
@@ -0,0 +1,30 @@
...
+  public void testPruneSymbolsProvidedInOtherFile() throws Exception {
+    String expected = DeclarationGeneratorTests.getTestFileText(input("prune_other/require.d.ts"));
+    assertThatProgram(
+        singletonList(input("prune_other/require.js")),
+        singletonList(input("prune_other/provide.js")))
+        .generatesDeclarations(false, expected);
+  }
+
+  private File input(String relativePath) {
+    Path testDir = FileSystems.getDefault().getPath("src", "test", "java");
```
mprobst: We could share the code to get the base directory, couldn't we?

src/main/java/com/google/javascript/cl2dts/Options.java
alexeagle: this isn't a state change - the depgraphs field is effectively masking the depgraphFiles field, so that tests don't need to write tmpfiles

src/main/java/com/google/javascript/cl2dts/Depgraph.java
alexeagle: the subtypes are mixed - some strings and some nested lists. added more generics to the extent possible

src/main/java/com/google/javascript/cl2dts/Options.java
mprobst: It changes state of this class and holds onto the file contents, that's what I find odd in an Options class that otherwise is just a holder for the command line options.

Couldn't you create a separate entry point that's called by the test code that needs it and the regular code, that happens *after* reading the files? I think that's the common way to do such things?

alexeagle: Sure, we just need another telescoping overload of generateDeclarations - done

alexeagle: But it still didn't change the state of this class :)
You're assuming that depgraphs is modified after reading depgraphFiles which it isn't

mprobst:
LGTM.

I'm boarding now. It'd be really nice if we could get the `depgraphs` thing a bit cleaner, but please go ahead at your discretion if that get's too complex or has too much overhead at this point.

src/test/java/com/google/javascript/cl2dts/MultiFileTest.java
alexeagle: On Fri, Sep 18, 2015 at 4:49 PM Martin Probst <notifications@github.com>
wrote:

> In src/test/java/com/google/javascript/cl2dts/MultiFileTest.java
> <https://github.com/angular/closure-to-dts/pull/102#discussion_r39911638>:
>
> > +import java.nio.file.Path;
> > +
> > +public class MultiFileTest {
> > +
> > +  // Repro for https://github.com/angular/closure-to-dts/issues/101
> > +  @Test
> > +  public void testPruneSymbolsProvidedInOtherFile() throws Exception {
> > +    String expected = DeclarationGeneratorTests.getTestFileText(input("prune_other/require.d.ts"));
> > +    assertThatProgram(
> > +        singletonList(input("prune_other/require.js")),
> > +        singletonList(input("prune_other/provide.js")))
> > +        .generatesDeclarations(false, expected);
> > +  }
> > +
> > +  private File input(String relativePath) {
> > +    Path testDir = FileSystems.getDefault().getPath("src", "test", "java");
>
> We could share the code to get the base directory, couldn't we?
>

Actually I think we might want to collapse the existing tests to share the
multi-file fixture, they just happen to have a single .js file.
Maybe in a followup change, this is getting big

PR #104, opened 2015-09-21
Add support for Arguments and NodeList.

rkirov:
Platform apis like Arguments and Nodelist need to be translated to
their corresponding apis in lib.d.ts

Upgrade closure to the newest released version.

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -480,18 +480,7 @@ public class DeclarationGenerator {
     }
 
     private void walkDefaultInterface(FunctionType ftype) {
-      for (String propName : ftype.getOwnPropertyNames()) {
-        if (propName.equals("prototype")) continue;
-        JSType pType = ftype.getPropertyType(propName);
-        // Here we assume the enum is exported and handled separately.
-        if (pType.isEnumType()) continue;
-        emit("var");
-        emit(propName);
-        emit(":");
-        visitType(pType);
-        emit(";");
-        emitBreak();
-      }
+      visitNamespaceLikeType(ftype);
     }
 
     private void visitTemplateTypes(ObjectType type) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -480,18 +480,7 @@ private void walk(TypedVar symbol, boolean isDefault) {
...
-        if (pType.isEnumType()) continue;
-        emit("var");
-        emit(propName);
-        emit(":");
-        visitType(pType);
-        emit(";");
-        emitBreak();
-      }
+      visitNamespaceLikeType(ftype);
     }
```
rkirov: This is a clean up for a previous PR (`visitNamespaceLikeType` was added by not used), not directly related with the change.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -643,6 +632,18 @@ public Void caseTemplatizedType(TemplatizedType type) {
             emit("[]");
             return null;
           }
+          // Arguments<?> and NodeList<?> in es3 externs are correspondinly
+          // IArguments and NodeList interfaces (not-parametrized) in lib.d.ts.
+          // New* are temporary work-arounds for upstream externs.
+          // TODO(rado): upgrade closure compiler and remove them.
+          if (type.getDisplayName().equals("Arguments") || type.getDisplayName().equals("NewArguments")) {
```
alexeagle: can we figure out whether this is the Arguments symbol we are interested in, so we don't take other user-defined types that happen to share the name?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -643,6 +632,18 @@ public Void caseTemplatizedType(TemplatizedType type) {
...
+            emit("IArguments");
+            return null;
+          }
+          if (type.getDisplayName().equals("NodeList") || type.getDisplayName().equals("NewNodeList")) {
```
alexeagle: seems like we should probably extract a configuration for this map

rkirov: Are you suggesting data driven config, or just code-wise extract it. Because I don't think we know the nature of the beast yet, to create a nice data DSL. These examples might be a bit deceiving. think about `Array.<?>` vs `any[]`.

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: Unless I am mistaken `displayName` is qualified, so there is little chance that would overlap as custom classes will be `foo.bar.Arguments`. 

Also, I can't seem to find an api to get the originating file of the type declaration. I guess I can add some JSDoc heuristics, but then we are one reformating away from braking.

PR #105, opened 2015-09-21
fix dedups fields from prototype chain.

rkirov:
It appears that @lends has a way to create duplicate fields.
Adds a verification that deduping is not needed if prototype fields
are just added (without @lends), but conservatively we always do
deduping.

### DIFF
--- a/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
@@ -831,8 +831,8 @@ public class DeclarationGenerator {
         emit("[key: string]: any;");
         emitBreak();
       }
-      // Methods.
-      visitProperties(prototype, false);
+      // Prototype fields (mostly methods).
+      visitProperties(prototype, false, ((ObjectType) instanceType).getOwnPropertyNames());
       // Statics.
       if (processStatics) {
         visitProperties(type, true);
@@ -843,7 +843,13 @@ public class DeclarationGenerator {
     }
 
     private void visitProperties(ObjectType objType, boolean isStatic) {
+      visitProperties(objType, isStatic,  Collections.<String>emptySet());
+    }
+
+    private void visitProperties(ObjectType objType, boolean isStatic, Set<String> skipNames) {
       for (String propName : objType.getOwnPropertyNames()) {
+        if (skipNames.contains(propName)) continue;
+
         if ("prototype".equals(propName) || "superClass_".equals(propName)
             // constructors are handled in #visitObjectType
             || "constructor".equals(propName)) {
--- a/src/test/java/com/google/javascript/cl2dts/lends.js
+++ b/src/test/java/com/google/javascript/cl2dts/lends.js
@@ -1,0 +1,19 @@
+goog.provide('lends');
+goog.provide('lends.A');
+
+/**
+ * @constructor
+ */
+lends.A = function() {
+  /** @type {string} */
+  this.a = "";
+};
+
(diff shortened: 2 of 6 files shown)


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -829,7 +829,13 @@ private void visitObjectType(ObjectType type, ObjectType prototype, Boolean proc
     }
 
     private void visitProperties(ObjectType objType, boolean isStatic) {
+      visitProperties(objType, isStatic, null);
```
mprobst: Why not always pass `Collections.emptySet()` here, so that you don't have to handle `null` below? It's usually better to avoid `null` and favour empty collections instead (fewer code paths, less pondering on nullness, etc).

src/test/java/com/google/javascript/cl2dts/lends.js
```diff
@@ -0,0 +1,19 @@
...
+  /** @type {string} */
+  this.a = "";
+};
+
+goog.object.extend(lends.A, /** @lends {lends.A} */ {b: 1});
+
+// lends manages to avoid the deduping of fields, and as far as closure
+// is concerned there is a field a on the prototype and on the instance.
+// Also the presense of @lends extends the type signature of A.prototype.
+goog.object.extend(lends.prototype.A, /** @lends {lends.A.prototype} */ {a: 1, c: true});
```
mprobst: is this really `lends.prototype.A` on the LHS, and not `lends.A.prototype`? That seems super confusing to me (but then again, that wouldn't be new...).

Assuming it should be the latter, this looks to me as if `a` should end up on `lends.A` and on `lends.A.prototype`, but maybe Closure syntax is sort of overlapping here, confusing `lends.A` the instance type (vs `lends.A` the static container) and `lends.A.prototype`, which ends up mostly being the instance type?

It might be worth raising this on the mailing list. Your fix still seems reasonable.

src/test/java/com/google/javascript/cl2dts/provide_single_class.js
```diff
@@ -6,6 +6,11 @@ goog.provide('foo.bar.Baz.NestedEnum');
 foo.bar.Baz = function() {
   /** @type {string} */
   this.field = 'a';
+  // Surprisingly, this doesn't throw a type error.
```
mprobst: maybe not just `this`, but `declaring a field on the prototype and in the constructor` to make it a bit clearer?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: good point, it was a bad optimizing habit :)

src/test/java/com/google/javascript/cl2dts/provide_single_class.js
rkirov: done.

src/test/java/com/google/javascript/cl2dts/lends.js
rkirov: that is correct `lends.A.prototype` is the correct usecase.

You confusion is very warranted. In fact half of the documentation for @lends is spent trying to clarify that point. https://developers.google.com/closure/compiler/docs/js-for-compiler?hl=en

Basically, @lends works beyond the type system, and it works against the `raw` JS objects and the keys we expect on them. In that case `lends.A` would tell closure that `a` is a static method. So working as expected, though expected is quite odd.

Overall @lends is quite a confusing feature, and really only makes sense when you try to type `goog.object.extend` method of dynamically building constructors.

PR #108, opened 2015-09-23
replace IThenable (closure extern) with PromiseLike (lib.d.ts)

rkirov:


src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -647,6 +647,9 @@ public Void caseTemplatizedType(TemplatizedType type) {
             emit("NodeList");
             return null;
           }
+          if (type.getDisplayName().equals("IThenable")) {
```
mprobst: Java 7 allows string literals in switch statements:
http://docs.oracle.com/javase/7/docs/technotes/guides/language/strings-switch.html

Would it make sense to tackle these as cases based on display name?
```java
switch (type.getDisplayName()) {
  case "Arguments":
  case "NewArguments":
    ...
```

src/test/java/com/google/javascript/cl2dts/types_with_externs.js
```diff
@@ -92,4 +92,10 @@ typesWithExterns.id = function(x) { return x; }
...
  */
-typesWithExterns.topLevelFunction = function() {};
\ No newline at end of file
+typesWithExterns.topLevelFunction = function() {};
+
+/**
+ * @interface
+ * @extends {IThenable}
+ */
+typesWithExterns.extendsIThenable = function() {};
```
mprobst: Nit: Closure style is UpperCamelCase for types.

mprobst:
LGTM

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
rkirov: Refactored. Btw, this is still quite yuki special case code. We talked with Alex about even bigger refactoring to have these guys be driven by a configuration file. The plan is to collect a bit more examples first.

mprobst: Yeah, it's a bit ugly, but then again in my experience a bunch of special cases in code don't get nicer just because you move them out to a config file, you're just making the code slightly harder to understand through the indirection. Assuming we never have to actually configure this, I'd rather have this straight forward hard coded.

PR #109, opened 2015-09-23
Forward

rkirov:


src/test/java/com/google/javascript/cl2dts/DeclarationGeneratorTests.java
```diff
@@ -67,7 +67,7 @@ static File getGoldenFile(final File input) {
   static String getTestFileText(final File input) throws IOException {
     String text = Files.asCharSource(input, Charsets.UTF_8).read();
     // Strip test comments in .d.ts golden files starting with '//!!'.
-    return text.replaceAll("^\\s*//!!.*\\n", "");
+    return text.replaceAll("[ ]*//!!.*\\n", "");
```
mprobst: I think `\\s` was correct, the problem you ran into is probably the anchoring. If you switch the regexp to multiline mode using `"(?m)^\\s*//!!.*\\n"` this should be good.

Though actually I cannot see a reason not to strip the comments on all trailing lines, even if there was non-whitespace before.

src/test/java/com/google/javascript/cl2dts/MultiFileTest.java
```diff
@@ -26,6 +30,18 @@ public void shouldPruneProvidesFromNonrootFile() throws Exception {
         .generatesDeclarations(false, expected);
   }
 
+  @Test
+  public void shouldResolveForwardDeclares() throws Exception {
+    String expected = DeclarationGeneratorTests.getTestFileText(input("forward.d.ts"));
+    ArrayList<File> roots = new ArrayList<>();
```
mprobst: Why not `ImmutableList.of(input(...), input(...))`?

src/test/java/com/google/javascript/cl2dts/ProgramSubject.java
```diff
@@ -58,7 +58,7 @@ public void generatesDeclarations(Boolean withExterns, String expected) {
 
     String actual = new DeclarationGenerator(opts)
         .generateDeclarations(sourceFiles, NO_EXTERNS, new Depgraph(roots))
-        .replaceAll("^\\s*//!!.*\\n", "");
+        .replaceAll("[ ]*//!!.*\\n", "");
```
mprobst: Whoops. Could we move this to a shared constant somewhere?

src/test/java/com/google/javascript/cl2dts/shouldResolveForwardDeclares/forward.js
```diff
@@ -0,0 +1,11 @@
+goog.provide('forward.A');
```
mprobst: Shouldn't there be a `goog.forwardDeclare('forward.D');` here?

src/test/java/com/google/javascript/cl2dts/DeclarationGeneratorTests.java
rkirov: Problem was missing multi-line and `\n` being part of `\s`. Didn't know about `(?m)` that fixes it nicely.

src/test/java/com/google/javascript/cl2dts/MultiFileTest.java
rkirov: because my bag of Java tricks is not that large (remember this is my first Java project :))

src/test/java/com/google/javascript/cl2dts/shouldResolveForwardDeclares/forward.js
rkirov: So turns out that is not needed for closure to consider the type as present when going through forward.js even if it hasn't seen `forward.D`. I am guess that is a feature of closure?

Then it begs the question what is `goog.forwardDeclare` needed for?

mprobst: My understanding is that `goog.forwardDeclare` is useful if your code mentions another type (`forward.D`), but when your end user compiles this into a JSBinary, the forwarded type may or may not be part of the compilation unit.

It's a hack that allows users to make sure they definitely don't compile some code in, and don't have to trust JSCompiler's code pruning to work (which it sometimes doesn't, due to code that has side effects on load).

I think an appropriate test would be to have a `goog.forwardDeclare('forward.D');`, but not have a `forward.D` on the compilation path at all.

rkirov: That makes sense. I added a single file test with `goog.forwardDeclare` in addition to what we had.

I was unfortunately using the term forwardDeclare in different situation, which I renamed to `namedTypes`. The situation is the following:

`a.js` and `b.js` are part of the compilation. Class `A` in `a.js` uses type `B` from `b.js` but doesn't goog.require it. Both files are passed to closure. Turns out that if the order that those files are passed matters for the internal representation of types. If `a.js` is seen first type `B` is what internally is called a `namedType`. Outputting the named type without resolving is problematic for generic types, which don't need to be concertized in closure, but do need to be in TS. 

The fix in this PR is to follow the resolved type for namedTypes instead of just outputting the name.

mprobst: >
> I was unfortunately using the term forwardDeclare in different situation,
> which I renamed to namedTypes. The situation is the following:
>
> a.js and b.js are part of the compilation. Class A in a.js uses type B
> from b.js but doesn't goog.require it. Both files are passed to closure.
> Turns out that if the order that those files are passed matters for the
> internal representation of types. If a.js is seen first type B is what
> internally is called a namedType. Outputting the named type without
> resolving is problematic for generic types, which don't need to be
> concertized in closure, but do need to be in TS.
>
Interesting. That seems like incorrect behaviour, care to file an internal
bug against Closure Compiler?

mprobst: Makes sense, but this does sound like a bug in JSCompiler, being unable to use generic types before they are being declared, but also not receiving a warning about it (just JSCompiler things, I guess). Maybe file an issue with them (internal bug tracker)?

mprobst: Makes sense, but this does sound like a bug in JSCompiler, being unable to use generic types before they are being declared, but also not receiving a warning about it (just JSCompiler things, I guess). Maybe file an issue with them (internal bug tracker)?

src/main/java/com/google/javascript/cl2dts/DeclarationGenerator.java
```diff
@@ -69,6 +69,10 @@
 
   private static final Logger logger = Logger.getLogger(DeclarationGenerator.class.getName());
   private static final String INTERNAL_NAMESPACE = "ಠ_ಠ.cl2dts_internal";
+
+  // Comments in .d.ts and .js golden files starting with '//!!' are stripped.
+  public static final String GOLDEN_FILE_COMMENTS_REGEXP = "(?m)^\\s*//!!.*\\n";
```
alexeagle: more idiomatic (and performant) to hold a Pattern object here
public static final Pattern BLAH = Pattern.compile(string)

PR #114, opened 2015-10-12
Fix dependency on set ordering

### LINKED ISSUE #83: Unstable iteration order
When running tests on OSX with JDK8, the order of declarations in the .d.ts files differs from the golden files.

mprobst:
Fixes #83.


mprobst:
@rkirov in case you have time to look at something :-)

PR #124, opened 2015-10-25
fix: emit correct symbols from default namespace.

mprobst:
When generating in the default namespace (""), make sure not to strip
a character from encountered symbols.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/refer_from_default_ns.js
+++ b/src/test/java/com/google/javascript/clutz/refer_from_default_ns.js
@@ -1,0 +1,10 @@
+goog.provide('fn');
+goog.provide('fn.String');
+
+/** @return {fn.String} */
+fn = function() {
+  return new fn.String();  
+};
+
+/** @constructor */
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/refer_from_default_ns.js
```diff
@@ -0,0 +1,10 @@
+goog.provide('fn');
+goog.provide('fn.String');
+
+/** @return {fn.String} */
```
rkirov: not sure how this test is broken without your changes. Namespace should be empty to trigger the bug, but in the test it looks like 'fn'?

rkirov:
LGTM, just make sure the added test fails without the change.

src/test/java/com/google/javascript/clutz/refer_from_default_ns.js
mprobst: It does - the issue is that fn is essentially a `var`, so the namespace while visiting `fn` is `''`.

PR #125, opened 2015-10-25
feat: support constructor function types.

### LINKED ISSUE #121: Generate construct signatures for Closure's constructor-typed functions
Closure allows specifying a construct signature like this:
```js
/** @type {function(new:X, string, number)} */
var x;
```

We should translate this to a constructable interface:
```ts
var x: {new(a: string, b: number): X};
```

mprobst:
Fixes #121.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -436,10 +436,10 @@ public class DeclarationGenerator {
         // Closure represents top-level functions as classes because they might be new-able.
         // This happens through externs es3.js which has Function marked as constructor.
         // See https://github.com/angular/closure-to-dts/issues/90
-        boolean ordinaryFunctionAppearingAsClass = type.isConstructor() &&
-            type.getDisplayName().equals("Function");
+        boolean ordinaryFunctionAppearingAsClass =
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -435,23 +435,35 @@ private void walk(TypedVar symbol, boolean isDefault) {
...
-        if (type.isOrdinaryFunction() || ordinaryFunctionAppearingAsClass) {
+        if (ftype.isOrdinaryFunction() || ordinaryFunctionAppearingAsClass) {
           emit("function");
           emit(getUnqualifiedName(symbol));
           visitFunctionDeclaration(ftype);
           emit(";");
           emitBreak();
           return;
+        } else if (!ftype.isNominalConstructor()) {
+          // A top-level field that has a specific constructor function type.
```
rkirov: this is identical with code at line 515, but it might not be worth refactoring to reuse it. At least reuse `visitTypeDeclaration` instead of `visitType`

rkirov:
LGTM.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Refactored to extract that.

PR #126, opened 2015-10-26
fix: do not emit empty templatized args.

### LINKED ISSUE #123: Uint8Array < > (without type arguments)
`Uint8Array` should be a non-templatized subtype of the templatized type `ArrayLike<number>`.

We currently generate `Uint8Array < >` for it.

mprobst:
In Closure, types that filled all template args from a super type are
still represented by a templatized type.

Fixes #123.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/empty_type_args.d.ts
+++ b/src/test/java/com/google/javascript/clutz/empty_type_args.d.ts
@@ -1,0 +1,31 @@
+declare namespace ಠ_ಠ.clutz_internal.empty_type_args {
+  interface ITemplated < T > {
+  }
+}
+declare module 'goog:empty_type_args.ITemplated' {
+  import alias = ಠ_ಠ.clutz_internal.empty_type_args.ITemplated;
+  export default alias;
+}
+declare namespace ಠ_ಠ.clutz_internal.empty_type_args {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/empty_type_args.d.ts
```diff
@@ -0,0 +1,31 @@
+declare namespace ಠ_ಠ.clutz_internal.empty_type_args {
+  interface ITemplated < T > {
+  }
+}
+declare module 'goog:empty_type_args.ITemplated' {
+  import alias = ಠ_ಠ.clutz_internal.empty_type_args.ITemplated;
+  export default alias;
+}
+declare namespace ಠ_ಠ.clutz_internal.empty_type_args {
+  class NoMoreTemplateArgs implements ITemplated < number > {
```
alexeagle: I don't understand which line here was absent before the fix. Was there no `<number>` here?

mprobst: Yeah, our golden masters are getting a bit wordy.

This is all just part of the buildup - the problem was that when `NoMoreTemplateArgs` was used, it was printed as `NoMoreTemplateArgs < >`, which is bad syntax in TS.

PR #127, opened 2015-10-26
feat: mismatched inherited static properties.

### LINKED ISSUE #120: TypeScript bails on non-overriding `static` methods
In Closure, it is legal to (non-)override "static" methods:

```js
/** @constructor */
function Parent() {}
/** @param {string} a */
Parent.foo = function(a) {};

/** @constructor */
function Child() {}
/** @param {number} a */
Parent.foo = function(a) {};
```

Note that `Parent.foo`'s signature is incompatible with `Child.foo`'s signature. That's fine - for Closure, these are different methods. For TypeScript, the "class static" side of a child class has to be a subtype of the "class static" side of a Parent class. E.g.:

```ts
class Parent {
	static x(a: string) {}
}

class Child extends Parent {
	static x(a: number) {}
}
```

Fails to compile for that reason.

One possible, though slightly incorrect, way to fix that is to generate overloads:

```ts
class Parent {
	static x(a: string) {}
}

class Child extends Parent {
	static x(a: number);
	static x(a: string);
	static x(a: string|number) {}
}
```

This would allow calling `Child.x('string')`, which is wrong, but better than straight out failing to compile.

mprobst:
In TypeScript, the "static" side of classes must correctly inherit from
any of its parents. In Closure, the methods are completely separate and
do not have to match. To paper over the difference, emit overloads so
that TypeScript allows the inheritance.

This pretends functions exist that do not, and it will fail for
functions that only differ in return type, but it should be good enough
for most code, for now.

Fixes #120.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -436,10 +436,10 @@ public class DeclarationGenerator {
         // Closure represents top-level functions as classes because they might be new-able.
         // This happens through externs es3.js which has Function marked as constructor.
         // See https://github.com/angular/closure-to-dts/issues/90
-        boolean ordinaryFunctionAppearingAsClass = type.isConstructor() &&
-            type.getDisplayName().equals("Function");
+        boolean ordinaryFunctionAppearingAsClass =
+            ftype.isConstructor() && "Function".equals(ftype.getDisplayName());
 
-        if (type.isOrdinaryFunction() || ordinaryFunctionAppearingAsClass) {
+        if (ftype.isOrdinaryFunction() || ordinaryFunctionAppearingAsClass) {
           emit("function");
           emit(getUnqualifiedName(symbol));
           visitFunctionDeclaration(ftype);
@@ -447,12 +447,21 @@ public class DeclarationGenerator {
           emitBreak();
           return;
         }
-        if (type.isConstructor()) {
+
+        if (!ftype.isNominalConstructor()) {
+          // A top-level field that has a specific constructor function type.
+          // <code>/** @type {function(new:X)} */ foo.x;</code>
+          visitVarDeclaration(symbol, ftype);
+          return;
+        }
+
+        if (ftype.isConstructor()) {
+          // "proper" class constructor
           emit("class");
-        } else if (type.isInterface()) {
+        } else if (ftype.isInterface()) {
           emit("interface");
         } else {
-          checkState(false, "Unexpected function type " + type);
+          checkState(false, "Unexpected function type " + ftype);
         }
         emit(getUnqualifiedName(symbol));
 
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -900,32 +900,79 @@ private void visitProperties(ObjectType objType, boolean isStatic, Set<String> s
...
+     *   class Child extends Parent { static foo(a: number); }
+     * </pre>
+     *
+     * To fix, the code below re-emits (by recursion) any properties that have a colliding type.
+     * 
+     * This is necessarily incomplete, and will fail for differences that cannot be dispatched on,
+     * e.g. different return types. It also pretends that functions exist that do not.
+     */
+    private void emitStaticOverloads(String propName, ObjectType objType, JSType propertyType) {
+      if (!objType.isFunctionType()) {
```
rkirov: Technically, we can get into the same issue with static fields. It is ok to not bother with fields for now, but document this is something that we might have to revisit.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -900,32 +900,79 @@ private void visitProperties(ObjectType objType, boolean isStatic, Set<String> s
...
+      }
+      emit(";");
+      emitBreak();
+      if (isStatic) {
+        emitStaticOverloads(propName, objType, propertyType);
+      }
+    }
+
+    /**
+     * Emit overloads for static methods from super classes that do not match the current function.
```
rkirov: `do not match the signature of the current property`.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -900,32 +900,79 @@ private void visitProperties(ObjectType objType, boolean isStatic, Set<String> s
...
+     */
+    private void emitStaticOverloads(String propName, ObjectType objType, JSType propertyType) {
+      if (!objType.isFunctionType()) {
+        return;
+      }
+      ObjectType superType = getSuperType((FunctionType) objType);
+      if (superType == null || superType.getConstructor() == null) {
+        return;
+      }
+      superType = superType.getConstructor();
```
rkirov: This closure API is quite odd, but why not have a new var `superTypeCtor`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: ops, misread as `!propertyType.isFunctionType`. I think your code *should* work with fields too (famous last works), can you add a test for those too.

rkirov:
LGTM, just add a test for fields too.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: That doesn't actually work - you can never declare two properties with the same name on the same class, you cannot *overload* properties. You could get close using setters, but you still cannot overload based on return value for getters.

I've added an explicit check and a more explanatory error for this.

PR #132, opened 2015-10-30
fix: inner classes, enums and interfaces are emitted.

rkirov:
Previously, we expected an explicit goog.provide for inner classes,
enums and interfaces, but according to style guide such requirers are
forbidden.

Now we do a second pass over the properties of default exported classes
and emit the inner properties that are not static methods or fields into
a namespace.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -287,11 +287,28 @@ public class DeclarationGenerator {
       treeWalker.walkDefaultInterface((FunctionType) symbol.getType());
       emitNamespaceEnd();
     }
+    // extra walk required for inner classes and inner enums. They are allowed in closure,
+    // but not in TS, so we have to generate a namespace-class pair in TS.
+    if (isDefault && hasNestedTypes(symbol.getType())) {
+      emitNamespaceBegin(symbol.getName());
+      treeWalker.walkInnerClassesAndEnums((ObjectType) symbol.getType(), symbol.getName(), provides);
+      emitNamespaceEnd();
+    }
     // skip emitting goog.require declarations for value empty namespaces, as calling typeof
     // does not work them.
     if (treeWalker.valueSymbolsWalked > 0) {
       emitGoogRequireSupport(namespace, isDefault ? symbol.getName() : namespace);
     }
+  }
+
+  private boolean hasNestedTypes(JSType type) {
+    if (!type.isConstructor()) return false;
+    FunctionType ftype = (FunctionType) type;
+    for (String name : ((FunctionType) type).getOwnPropertyNames()) {
+      JSType propType = ftype.getPropertyType(name);
+      if (propType.isConstructor() || propType.isEnumType() || propType.isInterface()) return true;
+    }
+    return false;
   }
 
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -294,6 +301,16 @@ public int compare(TypedVar o1, TypedVar o2) {
     }
   }
 
+  private boolean isClassWithInner(JSType type) {
```
mprobst: maybe call it `hasNestedTypes`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -425,9 +442,10 @@ private void emitComment(String s) {
 
   private ObjectType getSuperType(FunctionType type) {
     ObjectType proto = type.getPrototype();
-    return proto == null ? null :
-        proto.getImplicitPrototype().getDisplayName().equals("Object") ? null :
-        proto.getImplicitPrototype();
+    if (proto == null) return null;
+    ObjectType implicitProto = proto.getImplicitPrototype();
+    if (implicitProto == null) return null;
+    return implicitProto.getDisplayName().equals("Object") ? null : implicitProto;
```
mprobst: Display name occasionally ends up being null. Call `"Object".equals`?

mprobst: Is there a constant somewhere for the Object proto that we could compare to?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -554,6 +532,49 @@ private void walk(TypedVar symbol) {
       }
     }
 
+    private void visitClassAndInterface(String name, FunctionType ftype) {
```
mprobst: `visitClassOrInterface`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1105,5 +1124,18 @@ private void visitFunctionParameters(FunctionType ftype, boolean emitTemplatized
...
       emit(")");
     }
+
+    public void walkInnerClassesAndEnums(ObjectType type, String namespace, Set<String> provides) {
+      for (String propName : getSortedPropertyNames(type)) {
+        if (provides.contains(namespace + '.' + propName)) continue;
+        JSType pType = type.getPropertyType(propName);
+        if (pType.isEnumType()) {
+          visitEnumType((EnumType) pType);
+        }
```
mprobst: else if to make clear these are mutually exclusive?

mprobst:
LGTM

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: not sure how to get a hold of the cannonical Object object from closure.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

Issue #131, opened 2015-10-30
Prevent accidental structural typing


mprobst:
We could skip emitting private fields in `.d.ts` files. This allows better caching in build systems, but could also reduce the number of potentially problematic "overridden static field as wrong type" issues.

We should still emit a `private noStructuralTyping_: any;` field for every Closure `class`, to make sure classes don't accidentally get type compatible where they should not.

rkirov:
We already skip private properties, for example:

https://github.com/angular/clutz/blob/f97f4b01219b50959cf1115ea4615a1c8b7cf2ae/src/test/java/com/google/javascript/clutz/fn_params.js#L36

and the output

https://github.com/angular/clutz/blob/f97f4b01219b50959cf1115ea4615a1c8b7cf2ae/src/test/java/com/google/javascript/clutz/fn_params.d.ts

noStructuralTyping for ones with private is good idea, so keeping the issue open.

mprobst:
Odd, I thought I saw a case where we emit private static class fields.

In any case, will send a PR for the other issue.

PR #133, opened 2015-11-02
feat: prevent structural typing w/ private fields.

mprobst:
When a type has private fields, add special fields called
`private no(ClassSide)?StructuralTyping_: any;` to avoid accidentally
applying structural typing to a type that does not support that.

Also move all tests for `@private` annotations into their own file, so later
maintainers have a chance to find that code.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -321,7 +321,7 @@ public class DeclarationGenerator {
     // but not in TS, so we have to generate a namespace-class pair in TS.
     if (isDefault && hasNestedTypes(symbol.getType())) {
       emitNamespaceBegin(symbol.getName());
-      treeWalker.walkInnerClassesAndEnums((ObjectType) symbol.getType(), symbol.getName(), provides);
+      treeWalker.walkInnerClassesAndEnums((ObjectType) symbol.getType(), symbol.getName());
       emitNamespaceEnd();
     }
     return treeWalker.valueSymbolsWalked;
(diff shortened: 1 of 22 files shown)


mprobst:
Fixes #131.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -964,8 +965,21 @@ private void visitProperties(ObjectType objType, boolean isStatic, Set<String> s
             || "constructor".equals(propName)) {
           continue;
         }
+        if (isPrivate(objType.getOwnPropertyJSDocInfo(propName))) {
+          hasPrivateProps = true;
+          continue;
```
rkirov: It would be safe to continue into the `visitProperty` call, which already has private property check. That way if we need to modify whether we emit private properties we have to check any only one place. `visitProperties` is only concerned with emitting the no*Typing props.

rkirov:
lgtm

mprobst:
PTAL, I've changed this quite a bit in response to your comment about where `private` should be checked.

Closure does not allow structural typing for classes, so we should really emit the `private noStructuralTyping_: any;` for any `class` (not `interface`) that does not extend another `class` - not only for the ones where we filter `private` fields. Ideally we'd also do something for interfaces, but TypeScript (sensibly) doesn't allow `private` properties on them.

rkirov:
Looks even better. Closure authored classes should not be structurally casted into at TS, as that would break closure compilation, so indeed all closure classes should be forbidden.

There is a bigger question about how would sickle deal with TS code that uses structural typing, but I see you have https://github.com/angular/sickle/issues/5 for that.

PR #135, opened 2015-11-04
emitting static overloads skips over private and compatible fields.

rkirov:
Instead of throwing errors, we skip over if one (or all) of the
overlapping definitions are private. Also we support overloading when
the types are compatible.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1048,7 +1048,7 @@ public class DeclarationGenerator {
       }
       emit(";");
       emitBreak();
-      if (isStatic) {
+      if (isStatic && !isPrivate(objType.getOwnPropertyJSDocInfo(propName))) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1080,7 +1080,9 @@ private void emitStaticOverloads(String propName, ObjectType objType, JSType pro
       }
       // "I, for one, welcome our new static overloads." - H.G. Wells.
       JSType superPropType = superTypeCtor.getPropertyType(propName);
-      if (!superPropType.equals(propertyType)) {
+      if (!superPropType.isSubtype(propertyType)) {
+        // If the super field is private we have no issue.
```
mprobst: Maybe add `because private fields do not get emitted`? TypeScript itself would bail.

mprobst: Nit: prefer 3rd person voice in comments.

mprobst:
LGTM.

PR #145, opened 2015-11-08
fix: always generate absolute names.

### LINKED ISSUE #144: Repeated package names can shadow "higher up" packages
```ts
declare namespace clutz.foo {
  class Bar {}
}
declare namespace clutz.x.foo {
  // No class Bar here.
}
declare namespace clutz.x.whatever {
  // Class X cannot find "foo.Bar", because it is looking in `clutz.x.foo`, not in `clutz.foo`.
  class X extends foo.Bar {}
}
```

mprobst:
This makes it much less likely that a type or package gets shadowed by another
type or package with the same name somewhere in the namespace hierarchy.

There is a remaining corner case around re-declared global names like `Element`
or `Object`. The fix would be to alias *all* global names, but that risks
aliasing global names from Closure that do not exist in TypeScript.

Fixes #144.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
+++ b/src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
@@ -1,7 +1,7 @@
 declare namespace ಠ_ಠ.clutz_internal.nested.bar {
   type Enum = number ;
   var Enum : {
-    A : nested.baz.Enum ,
+    A : ಠ_ಠ.clutz_internal.nested.baz.Enum ,
   };
 }
 declare namespace ಠ_ಠ.clutz_internal.goog {
@@ -14,7 +14,7 @@ declare module 'goog:nested.bar.Enum' {
 declare namespace ಠ_ಠ.clutz_internal.nested.baz {
   type Enum = number ;
   var Enum : {
-    A : Enum ,
+    A : ಠ_ಠ.clutz_internal.nested.baz.Enum ,
   };
 }
 declare namespace ಠ_ಠ.clutz_internal.goog {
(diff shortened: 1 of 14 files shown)


mprobst:
Aliasing all global names would mean having:

```ts
declare namespace ಠ_ಠ.clutz_internal.global_aliases {
	type AliasedObject = Object;
	type AliasedElement = Element;
}
declare namespace ಠ_ಠ.clutz_internal {
	type Object = ಠ_ಠ.clutz_internal.global_aliases.AliasedObject;
	type Element = ಠ_ಠ.clutz_internal.global_aliases.AliasedElement;
}
```

That'd allow us to always slap "ಠ_ಠ.clutz_internal" in front of any type, and thus always be correct - we'd move the entire Closure namespace into clutz_internal. But I wonder how we could reliably generate a list of all the required aliases?

src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
```diff
@@ -14,7 +14,7 @@ declare module 'goog:nested.bar.Enum' {
 declare namespace ಠ_ಠ.clutz_internal.nested.baz {
   type Enum = number ;
   var Enum : {
-    A : Enum ,
+    A : ಠ_ಠ.clutz_internal.nested.baz.Enum ,
```
rkirov: in this case you can guarantee the `Enum` is what you want unambiguously, but then again I don't see much value in producing the cleanest .d.ts possible.

mprobst: Yeah, I considered special casing e.g. types from the same namespace, but I think it's not worth it right now.

rkirov:
As discussed moving lib.d.ts to ಠ_ಠ.clutz_internal.global_aliases is doable and we have done something similar with closure.lib.d.ts and Error. What can be tricky is that we have type-var pairs like (Error and ErrorConstructor) in lib.d.ts. It might have to be semi-manual work.

A whole separate concern is that a lot of the lib.d.ts types are not even usable by clutz generated code without a manual mapping from closure types Arguments -> IArguments etc, so doing all of lib.d.ts would be a wasted effort.

Let's revisit, when the need arises.

PR #146, opened 2015-11-10
feat: emit JSDoc.

mprobst:
Not only to show off in the demo :-)

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -14,6 +14,7 @@ import com.google.common.base.Functions;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
+import com.google.common.base.Splitter;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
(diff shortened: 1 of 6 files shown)


mprobst:
/CC @alexeagle we might want to show this in some future demo.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -586,6 +590,32 @@ private void walk(TypedVar symbol) {
...
+    private void maybeEmitJsDoc(JSDocInfo docs, boolean ignoreParams) {
+      if (docs == null) return;
+      String desc = docs.getBlockDescription();
+      if (desc == null) return;
+      emit("/**");
+      emitBreak();
+      if (desc != null) {
+        for (String line : Splitter.on('\n').split(desc)) {
+          emit(" *");
+          if (!line.isEmpty()) emit(line);
```
alexeagle: why avoid emitting empty string?

src/test/java/com/google/javascript/clutz/types.d.ts
```diff
@@ -6,7 +6,13 @@ declare namespace ಠ_ಠ.clutz_internal.types {
   var e : any [] ;
   var f : ( ( ) => string ) [] ;
   var functionAndUnion : (a : number , b : any ) => any ;
+  /**
+   * marked const to appear in `compiler.getTopScope().getAllSymbols()`
+   */
   var inferrednum : number ;
```
alexeagle: I assume we just lose line-comments like
`var thing; // thing is an important stuff`

alexeagle:
Indeed I need this for awesome demo

alexeagle:
LGTM

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Because it causes emit to emit a trailing space, and I'm a pedant.

src/test/java/com/google/javascript/clutz/types.d.ts
mprobst: My mental model is that `// ...` is an implementation comment, and we should only emit documentation comments.

PR #147, opened 2015-11-10
fix: handle empty union types.

### LINKED ISSUE #141: null type is not emitted in `function(null=)`
example https://github.com/google/closure-library/blob/28d9db61f5dc639c010be74e4d61682121d2dbd7/closure/goog/soy/soy.js#L116

mprobst:
Union types can end up having no legal branch, e.g. with `null=`. Emit an `any`
type for those cases - `any` is TypeScript's closest representation for a
Null Type / Bottom Type.

Fixes #141.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/null_type.js
+++ b/src/test/java/com/google/javascript/clutz/null_type.js
@@ -1,0 +1,7 @@
+goog.provide('nulltypes');
+
+/**
+ * @param {null} a
+ * @param {null=} b
+ */
+nulltypes = function(a, b) {};
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1011,6 +1011,10 @@ public class DeclarationGenerator {
           return !input.isNullType() && !input.isVoidType();
         }
       });
+      if (alts.size() == 0) {
+        emit("any");
+        return;
+      }
       if (alts.size() == 1) {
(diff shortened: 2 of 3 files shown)


src/test/java/com/google/javascript/clutz/null_type.js
```diff
@@ -0,0 +1,7 @@
+goog.provide('lefunc');
+
+/**
+ * @param {null} a
+ * @param {null=} b
+ */
+lefunc = function(a, b) {};
```
rkirov: doesn't test the change, ie. union type.

src/test/java/com/google/javascript/clutz/null_type.js
```diff
@@ -0,0 +1,7 @@
+goog.provide('lefunc');
```
rkirov: 'le func' is this french for 'the funk'? Probably want `nullTypes`.

src/test/java/com/google/javascript/clutz/null_type.js
rkirov: duh, `null=` is `null|undefined`.

rkirov:
LGTM.

src/test/java/com/google/javascript/clutz/null_type.js
mprobst: :-)

src/test/java/com/google/javascript/clutz/null_type.js
mprobst: gnagnagnagna. Done.

PR #148, opened 2015-11-10
feat: handle private types in public APIs.

### LINKED ISSUE #140: non-private namespace var with a type of a private class errors
example https://github.com/google/closure-library/blob/28d9db61f5dc639c010be74e4d61682121d2dbd7/closure/goog/events/actioneventwrapper.js#L49

we emit the var but not the class (thus type).

mprobst:
Surprisingly, Closure allows private types to be used in public APIs.
Emit an empty object type literal for them, i.e. a type that allows no
use but being passed around and is incompatible with everything else.

Fixes #140.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -302,7 +302,10 @@ public class DeclarationGenerator {
       });
 
       ObjectType objType = (ObjectType) symbol.getType();
-      for (String property : objType.getPropertyNames()) {
+      // Can be null if the symbol is provided, but not defined.  
+      Set<String> propertyNames =
+          objType != null ? objType.getPropertyNames() : Collections.<String>emptySet();
+      for (String property : propertyNames) {
         if (!isPrivate(objType.getOwnPropertyJSDocInfo(property))) {
           desiredSymbols.add(symbol.getName() + "." + property);
         }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -732,6 +735,13 @@ private void visitType(JSType typeToVisit) {
       // See also JsdocToEs6TypedConverter in the Closure code base. This code is implementing the
       // same algorithm starting from JSType nodes (as opposed to JSDocInfo), and directly
       // generating textual output. Otherwise both algorithms should produce the same output.
+      if (isPrivate(typeToVisit.getJSDocInfo())) {
```
rkirov: Always find it crazy that we need to dance around JSDocInfo to get the visibility of a symbol. (no action required).

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -732,6 +735,13 @@ private void visitType(JSType typeToVisit) {
...
+        // TypeScript does not allow public APIs that expose non-exported/private types. Just emit
+        // an empty object literal type for those, i.e. something that cannot be used for anything,
+        // except being passed around.
+        emit("{}");
```
rkirov: How about adding `interface PrivateType {}` to closure.lib.d.ts and used here, so when this errors out the users will get a more descriptive message. Typewise, AFAIKT `{}` is equivalent to empty interface.

src/test/java/com/google/javascript/clutz/private_type.js
```diff
@@ -0,0 +1,7 @@
+goog.provide('private_type.user');
```
rkirov: nit: underscore namespaces are against Closure style guide (probably guilty of few of those in our tests myself).

rkirov:
LGTM

src/test/java/com/google/javascript/clutz/private_type.js
mprobst: Done.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Done, that's a good idea. Also added some JSDoc on that type.

PR #149, opened 2015-11-10
feat: support renaming aliased enums.

### LINKED ISSUE #139: enum redirection does not work when going from 'foo.bar.enum' to 'foo.enum'
For example https://github.com/google/closure-library/blob/28d9db61f5dc639c010be74e4d61682121d2dbd7/closure/goog/ui/button.js#L65

mprobst:
Enum names are the names of the symbol they get assigned to, not the
names of the underlying type they refer to.

Fixes #139.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -527,11 +527,11 @@ public class DeclarationGenerator {
     }
 
     private String getUnqualifiedName(TypedVar symbol) {
-      return lastPart(symbol.getName());
+      return getUnqualifiedName(symbol.getName());
     }
 
-    private String getUnqualifiedName(JSType type) {
-      return lastPart(type.getDisplayName());
+    private String getUnqualifiedName(String name) {
+      return lastPart(name);
     }
 
     private String lastPart(String input) {
@@ -576,7 +576,7 @@ public class DeclarationGenerator {
       } else {
         maybeEmitJsDoc(symbol.getJSDocInfo(), /* ignoreParams */ false);
         if (type.isEnumType()) {
-          visitEnumType((EnumType) type);
+          visitEnumType(symbol.getName(), (EnumType) type);
           return;
         }
         // Closure keeps type None for the symbol which became the type alias.
@@ -697,7 +697,7 @@ public class DeclarationGenerator {
       emitBreak();
     }
 
-    private void visitEnumType(EnumType type) {
+    private void visitEnumType(String symbolName, EnumType type) {
       // Enums are top level vars, but also declare a corresponding type:
       // <pre>
       // /** @enum {ValueType} */ var MyEnum = {A: ..., B: ...};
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -672,10 +672,10 @@ private void visitEnumType(EnumType type) {
       // var MyEnum: {A: MyEnum, B: MyEnum, ...};
       // </pre>
       // TODO(martinprobst): Special case number enums to map to plain TS enums?
-      visitTypeAlias(type.getElementsType().getPrimitiveType(), getUnqualifiedName(type));
+      String unqualifiedName = getUnqualifiedName( symbolName );
```
rkirov: funky whitespace

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -672,10 +672,10 @@ private void visitEnumType(EnumType type) {
...
+      visitTypeAlias(type.getElementsType().getPrimitiveType(), unqualifiedName);
       emit("var");
-      // TS `type` declarations accept only unqualified names.
```
rkirov: I thought the comment was useful, did you remove it intentionally?

src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
```diff
@@ -1,14 +1,14 @@
 declare namespace ಠ_ಠ.clutz_internal.nested.bar {
-  type Enum = number ;
-  var Enum : {
+  type HahaEnum = number ;
+  var HahaEnum : {
     A : ಠ_ಠ.clutz_internal.nested.baz.Enum ,
```
rkirov: This is very odd the items are of `Enum` type and not `HahaEnum`, and I think that's why I had the comment "The output is only correct for enums with the same name." before you added fully qualified type output.

However, it does seem to work because they are all `number` after dealiasing, so lets go with it.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -523,11 +523,11 @@ private String getAbsoluteName(ObjectType objectType) {
     }
 
     private String getUnqualifiedName(TypedVar symbol) {
-      return lastPart(symbol.getName());
+      return getUnqualifiedName(symbol.getName());
     }
 
-    private String getUnqualifiedName(JSType type) {
```
rkirov: OOC, what was the bug here. Type was not qualified to begin with?

rkirov:
LGTM

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: I keep forgetting I don't have `clang-format` in Java land. We should consider hooking up `google-java-format` at some point.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Ah I see, that referred to the line two above. Makes sense then.

src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
mprobst: Yeah, this was conscious on my side. But it turns out we don't have to explicitly emit the `getMembersType` value, as we obviously just aliased the correct type. So I'm using the unqualified name now.

Re enum only works for same name, you might also have run into accidentally aliasing the type at the time.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: No, I'm just unifying all calls to end up on the same call path, so that you cannot e.g. change `getUnqualifiedName` below, and then one method was unused - this change is mostly unrelated.

The actual problem was as mentioned above that the enum that was aliased was in a higher package than the lower enum, which due to the lookup mechanism of TypeScript caused some aliasing snafu.

PR #150, opened 2015-11-10
feat: support files that provide non-defined symbols.

mprobst:
This is sometimes used just for the require/provide markers, e.g. to support
AutoJsDeps.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/undefined_provide.d.ts
+++ b/src/test/java/com/google/javascript/clutz/undefined_provide.d.ts
@@ -1,0 +1,6 @@
+declare namespace ಠ_ಠ.clutz_internal.undefined_provide {
+}
+declare module 'goog:undefined_provide' {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/undefined_provide.d.ts
```diff
@@ -0,0 +1,6 @@
+declare namespace ಠ_ಠ.clutz_internal.undefined_provide {
+}
+declare module 'goog:undefined_provide' {
+  import alias = ಠ_ಠ.clutz_internal.undefined_provide;
+  export = alias;
+}
```
rkirov: At first I thought this is not emitting `goog.require` shim, but with our change to have default `goog.require` be empty interface, the default type is correct. Carry on.

rkirov:
LGTM.

DavidSouther:
LGTM

PR #151, opened 2015-11-11
rename clutz_internal to clutz.

rkirov:
Look-of-disapproval provides enough to show this is internal namespace.
Also the namespace shows up in inline IDE docs, so we need something
short.


mprobst:
LGTM.

PR #157, opened 2015-11-13
Private enums in type positions

mprobst:
Fixes #156.

PR #158, opened 2015-11-13
Add a test case for private nested classes.

### LINKED ISSUE #155: private inner classes should not be emitted

mprobst:
Does not fix #155.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/privates.js
+++ b/src/test/java/com/google/javascript/clutz/privates.js
@@ -1,4 +1,5 @@
 goog.provide('priv');
+goog.provide('priv2.PublicClass');
 
 /** @private */
 priv.field = 12;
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/privates.js
```diff
@@ -23,3 +23,6 @@ priv.PublicClass.prototype.publicField;
 
 /** @private @return {number} */
 priv.PublicClass.prototype.method_ = function() { return this.field; };
+
+/** @constructor @private */
+priv.PublicClass.PrivateNestedClass_ = function() {};
```
rkirov: `goog.provide('priv.PublicClass');` to see the bug

mprobst: Fixed.

mprobst:
Note that this is now based on #157.

rkirov:
LGTM

PR #159, opened 2015-11-17
Set input language to ES6

rictic:
This enables handling and understanding the broadest set of valid input JS.

Needs the setLanguageOut call because the compiler doesn't want to produce ES6.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -50,6 +50,8 @@ public class Options {
     options.setClosurePass(true);
     options.setCheckGlobalNamesLevel(CheckLevel.ERROR);
     options.setCheckGlobalThisLevel(CheckLevel.ERROR);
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -50,6 +50,8 @@ public CompilerOptions getCompilerOptions() {
     options.setClosurePass(true);
     options.setCheckGlobalNamesLevel(CheckLevel.ERROR);
     options.setCheckGlobalThisLevel(CheckLevel.ERROR);
+    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6);
```
rkirov: Not sure if clutz can handle any ES6 module usage well. But this will enable other ES6 features that need .d.ts declarations, which I think is your usecase.

rkirov:
LGTM, I will merge it. Thanks for contributing!

rkirov:
merged in 8b96bad8a7c63627e4411b2314b89b3ce81a04a2

PR #174, opened 2015-11-26
feat: support abstract methods.

### LINKED ISSUE #173: Support abstract classes
```ts
interface Bar { foo(): string; }
abstract class Foo implements Bar {
  abstract foo(): string;
}
```

mprobst:
Abstract methods in Closure are actually just methods that assign the
special value goog.abstractMethod into a property. The property then
has the type "Function", i.e. takes any parameters and returns any.

This change does not generate a TypeScript abstract class, but just
emits code exactly representing this pattern. We might have to revisit
later on, emitting proper abstract classes might be nicer for users.

Fixes #173.


rkirov:
Comments for this one are left on https://github.com/angular/clutz/pull/177, ops :)

rkirov:
LGTM, like the other PR.

PR #176, opened 2015-11-26
feat: handle nested enum type name collisions.

### LINKED ISSUE #154: namespaces must be covariant with respect to classes
```
class A {}
class B extends A {}

declare namespace A {
	var foo : {a:number};
}
declare namespace B {
	var foo : {};
}
```
is error in TS, but in closure A.foo and B.foo are not required to follow any type relation.

mprobst:
Nested types need to correctly inherit the "class static side" in TS.
Closure code however commonly re-defines incompatible enums in
subclasses, leading to TS side errors.

This change only fixes the problem for enums. It's unclear if we can
fix it for inherited `class`es, as that'd mean overloading a class ctor
function. However it turns out enums are the only common issue in the
actual, real world Closure code base.

Semi-fixes #154.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -57,7 +57,15 @@ import org.kohsuke.args4j.CmdLineException;
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
@@ -82,7 +90,7 @@ public class DeclarationGenerator {
   private final List<JSError> errors = new ArrayList<>();
   private StringWriter out = new StringWriter();
   private final Options opts;
-  private final Set<String> privateEnums = new HashSet<>();
+  private final Set<String> privateEnums = new LinkedHashSet<>();
 
   DeclarationGenerator(Options opts) {
     this.opts = opts;
@@ -655,7 +663,7 @@ public class DeclarationGenerator {
       } else {
         maybeEmitJsDoc(symbol.getJSDocInfo(), /* ignoreParams */ false);
         if (type.isEnumType()) {
-          visitEnumType(symbol.getName(), (EnumType) type);
+          visitEnumType(symbol.getName(), (EnumType) type, Collections.<JSType>emptySet());
           return;
         }
         if (isTypedef(type)) {
@@ -774,7 +782,7 @@ public class DeclarationGenerator {
       emitBreak();
     }
 
-    private void visitEnumType(String symbolName, EnumType type) {
+    private void visitEnumType(String symbolName, EnumType type, Iterable<JSType> intersectTypes) {
       // Enums are top level vars, but also declare a corresponding type:
       // <pre>
       // /** @enum {ValueType} */ var MyEnum = {A: ..., B: ...};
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1222,31 +1244,22 @@ private void visitProperty(String propName, ObjectType objType, boolean isStatic
...
-        if (isPrivateProperty(superTypeCtor, propName)) return;
-        if (!propertyType.isFunctionType() || !superPropType.isFunctionType()) {
-          errors.add(JSError.make(currentSymbol.getNode(), CLUTZ_OVERRIDDEN_STATIC_FIELD,
-              objType.getDisplayName(), propName));
-        } else {
-          emit("/** WARNING: emitted for non-matching super type's static method. "
-              + "Only the first overload is actually callable. */");
-          emitBreak();
+      for (JSType superPropType : incompatibleSuperTypes) {
+        if (!propertyType.isSubtype(superPropType)) {
```
rkirov: move this check into the TODO in collectIncompatibleSuperTypes. If the type is a subtype, it is *compatible* and not incompatible.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1314,16 +1327,48 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
...
+      if (!type.isFunctionType()) return;
+      ObjectType superType = getSuperType(type.toMaybeFunctionType());
+      if (superType == null) return;
+      FunctionType superTypeCtor = superType.getConstructor();
+      if (superTypeCtor == null) return;
+      if (superTypeCtor.hasOwnProperty(propName) && !isPrivateProperty(superTypeCtor, propName)) {
+        // If this field is public, but the super field is private, and thus not emitted, it's ok.
+        JSType superPropType = superTypeCtor.getPropertyType(propName);
+        // TODO(martinprobst): Should we actually check whether all accumulator types are subtypes,
+        // instead of just checking type equality?
```
rkirov: yes, otherwise we are not incompatible.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1314,16 +1327,48 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
         String qualifiedName = innerNamespace + '.' + propName;
         if (provides.contains(qualifiedName)) continue;
         JSType pType = type.getPropertyType(propName);
+
+        Set<JSType> incompatibleTypes = new LinkedHashSet<>();
+        incompatibleTypes.add(pType);
```
rkirov: this odd dance needs a comment.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1314,16 +1327,48 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
         String qualifiedName = innerNamespace + '.' + propName;
         if (provides.contains(qualifiedName)) continue;
         JSType pType = type.getPropertyType(propName);
+
+        Set<JSType> incompatibleTypes = new LinkedHashSet<>();
```
rkirov: I think we should be calling visitProperty here instead of repeating the same code, which your change makes painfully clear. Add a TODO for me if you agree that there is duplicated code.

rkirov:
LGTM

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Ack.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Extracted.

PR #178, opened 2015-11-26
feat: support untyped provides.

### LINKED ISSUE #172: Generated i18n defines not emitted for non @type'd symbols
```js
goog.provide('goog.i18n.CompactNumberFormatSymbols_af');
/**
 * Compact number formatting symbols for locale af.
 */
goog.i18n.CompactNumberFormatSymbols_af = {
  COMPACT_DECIMAL_SHORT_PATTERN: {...}
};
```

Generates the module, but no namespace, class, or other symbol.

mprobst:
Surprisingly, these behave differently from undefined provides, in that there
is no symbol at all for the provide (even though it is assigned!). Just emit
an empty namespace.

Fixes #172 for the immediate issue, but still would need to export the actual type shape at some point.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/untyped_provide.js
+++ b/src/test/java/com/google/javascript/clutz/untyped_provide.js
@@ -1,0 +1,10 @@
+goog.provide('untyped.Provide');
+// Surprisingly this behaves differently from undefined_provide.js, there's no symbol at all.
+
+/**
+ * Has no type, but interesting structure.
+ */
+untyped.Provide = {
+  SOME_FIELD: 'x',
+  SOME_OTHER_FIELD: 'y'
+};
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/untyped_provide.js
```diff
@@ -0,0 +1,10 @@
+goog.provide('untyped.Provide');
+// Surprisingly this behaves differently from undefined_provide.js, there's no symbol at all.
+
+/**
+ * Has no type, but interesting structure.
+ */
+untyped.Provide = {
+  SOME_FIELD: 'x',
+  SOME_OTHER_FIELD: 'y'
```
rkirov: This guys are still unaccessible in TS, how about we investigate emitting any for them (doesn't need to be in this PR).

rkirov:
LGTM, maybe keep issue open if we should follow up with traversing object props.

src/test/java/com/google/javascript/clutz/untyped_provide.js
mprobst: Good point, I'll re-open.

PR #180, opened 2015-11-26
feat: simplify & fix interface walking.

### LINKED ISSUE #179: Classes nested in interfaces only create var
```js
goog.provide('some.Interface');

/** @interface */
some.Interface = function() {};

/** @constructor */
some.Interface.SomeClass = function() {};
```

Generates:

```ts
declare namespace clutz.some.Interface {
  var SomeClass: { new(): clutz.some.Interface.SomeClass };
}
```

Which breaks, because there's no type to refer to later on, only the `var SomeClass` symbol.

mprobst:
Interfaces containing static properties can simply be treated like
namespaces with no default export (`isDefault == false`).

Fixes #179.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -380,9 +380,8 @@ public class DeclarationGenerator {
     // static methods. TS does not support static methods on interfaces, thus we create a new
     // namespace for them.
     if (isDefault && isInterfaceWithStatic(symbol.getType())) {
-      emitNamespaceBegin(symbol.getName());
-      treeWalker.walkDefaultInterface((FunctionType) symbol.getType());
-      emitNamespaceEnd();
+      declareNamespace(symbol.getName(), symbol, /* isDefault, prevents infinite recursion */ false,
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -368,9 +368,7 @@ public int compare(TypedVar o1, TypedVar o2) {
     // static methods. TS does not support static methods on interfaces, thus we create a new
     // namespace for them.
     if (isDefault && isInterfaceWithStatic(symbol.getType())) {
-      emitNamespaceBegin(symbol.getName());
-      treeWalker.walkDefaultInterface((FunctionType) symbol.getType());
-      emitNamespaceEnd();
+      declareNamespace(symbol.getName(), symbol, false, compiler, provides, isExtern);
```
rkirov: yey, recursion. You can add comment isDefault false guarantees termination, so that some further refactoring doesn't get us into infinite loops.

rkirov:
LGTM.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #181, opened 2015-11-27
feat: only fail for actual Clutz issues.

mprobst:
Clutz used to fail for every WARNING ever reported. That's generally
not a good idea for arbitrary Closure code, where users might be
ignoring many warnings.

This changes Clutz' error handling logic to only report WARNINGS in
debug mode, and to only specifically report errors that are likely
going to cause issues for Clutz' type conversions.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -79,42 +108,23 @@
...
   private final Options opts;
-  private Set<String> privateEnums = new HashSet<>();
+  private final Compiler compiler;
+  private final ClutzErrorManager errorManager;
+  private StringWriter out = new StringWriter();
+  private final Set<String> privateEnums = new HashSet<>();
 
   DeclarationGenerator(Options opts) {
     this.opts = opts;
+    this. compiler = new Compiler();
```
rkirov: extra space after '.'

no auto-format in eclipse?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -387,16 +363,16 @@ public int compare(TypedVar o1, TypedVar o2) {
...
     if (utype != null) {
-      return isNativeObjectType(utype.restrictByNotNullOrUndefined(), compiler);
+      return isNativeObjectType(utype.restrictByNotNullOrUndefined());
     }
-    return isNativeObjectType(type, compiler);
+    return isNativeObjectType(type);
   }
 
   // TODO(rado): refactor so that the native object type is a final field.
-  private boolean isNativeObjectType(JSType type, Compiler compiler) {
```
rkirov: Great, this was really bothering me. Much cleaner now, to have the lifetime of Generator match the lifetime of the compiler.

src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -48,8 +49,15 @@
   public CompilerOptions getCompilerOptions() {
     final CompilerOptions options = new CompilerOptions();
     options.setClosurePass(true);
+
+    // All diagnostics are WARNINGs (or off) and thus ignored unless debug == true.
+    // Only report issues (and fail for them) that are specifically causing problems for Clutz.
+    // The idea is to not do a general sanity check of Closure code, just make sure Clutz works.
+    // Report missing types as errors.
```
rkirov: Sounds like we are becoming more permissive, allowing wrong Closure code. Wouldn't those errors sooner or later end up in a js_binary rule and break the compilation?

src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -85,20 +88,13 @@ public void generatesDeclarations(String expected) {
...
-          expectedDiagnostics, unexpectedRes);
-    } catch (DeclarationGeneratorException e) {
-      assertThat(e.getMessage()).contains(expectedDiagnostics);
-    }
+  StringSubject diagnosticStream() {
+    return assertThat(parse(true));
   }
 
-  private String parse() throws AssertionError {
+  private String parse(boolean forDiagnostics) throws AssertionError {
```
rkirov: I find it odd that the same method returns stderr or proper output depending on flag. Why not introduce a new method parseAndReturnErr.

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
```diff
@@ -0,0 +1,49 @@
...
+      reportClutzMissingTypes = false;
+      this.report(CheckLevel.ERROR, JSError.make(error.sourceName, error.getLineNumber(),
+          error.getCharno(), DeclarationGenerator.CLUTZ_MISSING_TYPES));
+    }
+    super.report(level, error);
+  }
+
+  @Override
+  public void generateReport() {
+    // Don't do anything - only generate a report.
```
rkirov: confusing comment, can't generate a report by doing nothing?!?

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.google.javascript.jscomp.CheckLevel;
+import com.google.javascript.jscomp.JSError;
+import com.google.javascript.jscomp.MessageFormatter;
+import com.google.javascript.jscomp.PrintStreamErrorManager;
+
+import java.io.PrintStream;
+
+/**
+ * An error manager that delays reporting though {@link #doGenerateReport()}, allows failing fast,
```
rkirov: through

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
```diff
@@ -0,0 +1,49 @@
...
+ * and that adds special, Clutz-specific errors.
+ */
+final class ClutzErrorManager extends PrintStreamErrorManager {
+  private final boolean debug;
+  boolean reportClutzMissingTypes = true;
+
+  /**
+   * @param failEarly if true, directly fail on the first reported error (but not warning!).
```
rkirov: no such param? also looks like fail early didnt make the cut.

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+  @Override
+  public void report(CheckLevel level, JSError error) {
+    // Ignore warnings in non-debug mode.
+    if (!debug && level == CheckLevel.WARNING) return;
+
+    if (reportClutzMissingTypes
+        && error.description.contains("Bad type annotation. Unknown type")) {
+      // Prepend an error that hints at missing externs/dependencies.
+      reportClutzMissingTypes = false;
```
rkirov: why not report all occurrences? We are trading off noisy output for need to rerun to see more errors.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: The problem is that the Eclipse formatter would wreck the whole file if run on it, and there's no good option to only run it on changed lines. I wish google-java-format had an Eclipse plugin already...

src/main/java/com/google/javascript/clutz/Options.java
mprobst: As discussed offline, projects have different settings, and we just used to treat even every warning as an error. Also, we're really not in the business of telling people whether their Closure code is ok, that's best left to the actual JS binary rule.

src/test/java/com/google/javascript/clutz/ProgramSubject.java
mprobst: Yeah, I thought so too, but figured it's a private API. I changed this to return a two element `String[]`; as long as this is purely private to `ProgramSubject` I think that ought to be enough.

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
mprobst: Fixed the comment. The point is that we need to control the time when the report is generated, i.e. after clutz has run. That should be removed when/if we refactor Clutz to be a compiler pass.

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
mprobst: Done.

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
mprobst: Yeah, removed it again as it doesn't seem to be useful enough. So far it has been easy enough to reduce issues to small reproducible cases and attach a debugger to it, so no real need to run it in "throw a stack trace" mode.

src/main/java/com/google/javascript/clutz/ClutzErrorManager.java
mprobst: We still report the error itself, we just *also* report a more specific, more readable message. That message should only appear once.

Added a `fall through` comment to make that more obvious.

PR #189, opened 2015-12-09
do not throw on out-of-order sources.

rkirov:
JS compiler complains about late provides, but still emits correct type
information, so we turn off that check.

Fixes bug in error-checking on all unit tests (wrong assertion), and a
few js code changes that were necessitated by that.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/ProgramSubject.java
+++ b/src/test/java/com/google/javascript/clutz/ProgramSubject.java
@@ -81,7 +81,7 @@ class ProgramSubject extends Subject<ProgramSubject, ProgramSubject.Program> {
 
   void generatesDeclarations(String expected) {
     String[] parseResult = parse();
-    assertThat(parseResult[1].isEmpty());
+    assertThat(parseResult[1]).isEmpty();
     String actual = parseResult[0];
     String stripped =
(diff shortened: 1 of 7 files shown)


src/test/java/com/google/javascript/clutz/shouldWorkWithOutOfOrderProvides/index.js
```diff
@@ -0,0 +1,11 @@
...
+goog.require('dep.D');
+
+/**
+ * @constructor
+ */
+main.A = function() {
+};
+
+/** @param {dep.D} a*/
+main.A.prototype.fn;
```
mprobst: I think this should have an `= function(a) {};`, shouldn't it?

src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -81,7 +81,7 @@ ProgramSubject withExterns(Externs newExterns) {
 
   void generatesDeclarations(String expected) {
     String[] parseResult = parse();
-    assertThat(parseResult[1].isEmpty());
+    assertThat(parseResult[1]).isEmpty();
```
mprobst: ಠ_ಠ new fangled hipster assertion libraries. Filed an issue to use error prone.

mprobst:
LGTM.

Issue #186, opened 2015-12-09
Support goog.module


rkirov:
One problem with is that scoped symbols appear in a synthetic `$jscomp.scope` namespace, which cannot be emitted as is.

PR #193, opened 2015-12-10
feat: add a test for goo.module.

mprobst:
Related to #186.


mprobst:
@rkirov I can't quite reproduce what exactly the issue is - this all looks fine to me. Do you have a better repro?

rkirov:
Sorry, the issue turned out to be a lot more specific (see https://github.com/angular/clutz/tree/goog-module-default). Needs default export that introduces a new type.

Your tests is pretty good, so I am ok with submitting as it is and fixing in a separate PR.

PR #195, opened 2015-12-18
Emit class A (statics) and class A_Instance per closure class A.

### LINKED ISSUE #184: Fix static inheritance mismatches by separating static and instance inheritance chain
In Closure's type system, there are no classes, but rather just constructor functions. In contrast to TypeScript classes, constructor functions do not inherit static properties from their parents. Clutz has to generate actual classes and cannot use the `interface X {}; var X: { new(): X; ... statics from x ... }` trick, because Closure exposes nested classes within constructor functions. The nested classes must be represented in TS namespaces, as they are accessible as properties on their containing class, and the `var`s in above trick collide with namespace declarations.

A reasonable fix for this situation is separating the static hierarchy and the instance hierarchy, with only the instance hierarchy inheriting, and the static hierarchy extending the non-static counterparts. This matches the actual Closure inheritance semantics. The exception are ES6 classes that currently have no correct representation in Closure's type system, but are also very rarely used.

E.g. for the three classes inheriting in chain: `Child -> Parent -> GrandParent`, the effective inheritance chain would be (with `_I` representing the instance fields of the respective class, and `A == Ancestor == GrandParent`):

```
+-----+    +---+
| A_I <----+ A |
+--^--+    +---+
   |
+--+--+    +---+
| P_I <----+ P |
+--^--+    +---+
   |
+--+--+    +---+
| C_I <----+ C |
+-----+    +---+
```

And the corresponding TS declarations:

```ts
declare namespace x {
  class GrandParent extends GrandParent_Instance {
    static gpStatic: string;
  }
  class GrandParent_Instance {
    gpField: string;
  }
  class Parent extends Parent_Instance {
    static parentStatic: string;
  }
  class Parent_Instance extends GrandParent_Instance {
    parentField: string;
  }
  class Child extends Child_Instance {
    static childStatic: string;
    constructor(x: number);  // ctors are part of the static aspect of a class
  }
  class Child_Instance extends Parent_Instance {
    childField: string;
  }
}

declare namespace x.GrandParent {
  class Z {}  
}

declare namespace x.Parent {
  class Z {}  // Look, no collision!
}

var p = new x.Parent();
```

rkirov:
This fixes all sorts of static overwrite issues and removes a number of
cludgy solutions (like intersection types on enums, overloading function
declarations).

Closes #184

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -68,6 +68,8 @@ import javax.annotation.Nullable;
  */
 public class DeclarationGenerator {
 
+  static final String INSTANCE_CLASS_SUFFIX = "_Instance";
+
   public static void main(String[] args) {
     Options options = null;
     try {
@@ -94,10 +96,6 @@ public class DeclarationGenerator {
     System.exit(0);
   }
 
-  static final DiagnosticType CLUTZ_OVERRIDDEN_STATIC_FIELD =
-      DiagnosticType.error("CLUTZ_OVERRIDDEN_STATIC_FIELD",
-          "Found a statically declared field that does not match the type of a parent field "
-              + "with the same name, which is illegal in TypeScript.\nAt {0}.{1}");
   static final DiagnosticType CLUTZ_MISSING_TYPES = DiagnosticType.error("CLUTZ_MISSING_TYPES",
       "A dependency does not compile because it is missing some types. This is often caused by "
           + "the referenced code missing dependencies or by missing externs in your build rule.");
@@ -656,20 +654,7 @@ public class DeclarationGenerator {
       } else {
         maybeEmitJsDoc(symbol.getJSDocInfo(), /* ignoreParams */ false);
         if (type.isEnumType()) {
-          Set<JSType> incompatibleTypes = Collections.emptySet();
-
-          // Recovering the constructor type (if any) for the parent object, in case the enum is
-          // an nested enum.
-          JSType parentType = typeRegistry.getType(getNamespace(type.getDisplayName()));
-          ObjectType namespaceType = parentType != null ? parentType.toMaybeObjectType() : null;
-          FunctionType constructorType = namespaceType != null ? namespaceType.getConstructor() : null;
-
-          if (constructorType != null) {
-             incompatibleTypes = getIncompatibleSuperTypes(constructorType,
-                 getUnqualifiedName(symbol), type);
-          }
-
-          visitEnumType(symbol.getName(), (EnumType) type, incompatibleTypes);
+          visitEnumType(symbol.getName(), (EnumType) type);
           return;
         }
         if (isTypedef(type)) {
@@ -711,6 +696,33 @@ public class DeclarationGenerator {
     }
 
     private void visitClassOrInterface(String name, FunctionType ftype) {
+      // TypeScript classes (like ES6 classes) have prototypal inheritance on the static side,
+      // which means that if A extends B, A.foo and B.foo cannot be of incompatible types.
+      // Closure (ES5) classes have no such restriction.
+      // To avoid collisions on the static side, emit all statics in the original class name A,
+      // and emit instance fields in another class A_Instance. As such:
+      // class A extends A_Instance { <static fields and methods> }
+      // class B extends B_Instance { <static fields and methods> }
+      // class A_Instance extends B_Instance { <instance fields and methods> }
+      // class B_Instance { <instance fields and methods> }
+      //
+      // Emit original name class before Instance in order to match the already emitted JSDoc.
+      boolean emitInstance = !ftype.isInterface();
+      if (emitInstance) {
+        emit("class");
+        emit(name);
+        visitTemplateTypes(ftype);
+        emit("extends");
+        emit(name + INSTANCE_CLASS_SUFFIX);
+        visitTemplateTypes(ftype);
+        emit("{");
+        indent();
+        emitBreak();
+        visitProperties(ftype, true);
+        unindent();
+        emit("}");
+        emitBreak();
+      }
       if (ftype.isConstructor()) {
         // "proper" class constructor
         emit("class");
@@ -719,7 +731,7 @@ public class DeclarationGenerator {
       } else {
         checkState(false, "Unexpected function type " + ftype);
       }
-      emit(name);
+      emit(emitInstance ? name + INSTANCE_CLASS_SUFFIX : name);
 
       visitTemplateTypes(ftype);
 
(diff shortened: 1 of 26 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -68,6 +68,8 @@
  */
 public class DeclarationGenerator {
 
+  public static final String INSTANCE_CLASS_SUFFIX = "_Instance";
```
mprobst: I presume we won't run into collisions with actual user code with this, as `_` violates Closure style, but might be something we have to keep in mind for the one obscure case that always comes up.

mprobst: Does this have to be `public`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -711,6 +700,33 @@ private void maybeEmitJsDoc(JSDocInfo docs, boolean ignoreParams) {
     }
 
     private void visitClassOrInterface(String name, FunctionType ftype) {
+      // Typescript classes (like ES6 classes) have prototypal inheritance on the static side,
```
mprobst: Nit: `Type_S_cript`

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -711,6 +700,33 @@ private void maybeEmitJsDoc(JSDocInfo docs, boolean ignoreParams) {
...
+      // Closure (ES5) classes have no such restriction.
+      // To avoid collisions on the static side, emit all statics in the original class name A,
+      // and emit instance fields in another class A_Instance. As such:
+      // class A extends A_Instance { <static fields and methods> }
+      // class B extends B_Instance { <static fields and methods> }
+      // class A_Instance extends B_Instance { <instance fields and methods> }
+      // class B_Instance { <instance fields and methods> }
+      //
+      // Emit original name class before Instance in order to match the already emitted JSDoc.
+      boolean emitInstance = !ftype.isInterface();
```
mprobst: Good catch for the `interface` case.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -750,7 +766,8 @@ private void visitClassOrInterface(String name, FunctionType ftype) {
           }
         }
       }
-      visitObjectType(ftype, ftype.getPrototype(), !ftype.isInterface());
+      visitObjectType(ftype, ftype.getPrototype());
+
```
mprobst: nit empty line

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -863,6 +869,10 @@ private void visitTypeAsPrimary(JSType type) {
     }
 
     private void visitType(JSType typeToVisit) {
+      visitType(typeToVisit, false);
```
mprobst: Nit: maybe add an inline `/* extendingInstanceClass */` for the parameter name in call sites? It's a bit hard to read like this.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -737,7 +753,7 @@ private void visitClassOrInterface(String name, FunctionType ftype) {
       // Class extends another class
       if (getSuperType(ftype) != null) {
         emit("extends");
-        visitType(getSuperType(ftype));
+        visitType(getSuperType(ftype), emitInstance);
```
mprobst: I think the only place where we'd ever pass `true` here is for classes extending base classes. In that case, we always know it's a nominal type that we're extending (I don't think you can extend arbitrary type expressions in Closure - can you?).

So I wonder if the code would be cleaner if we'd just directly call the code to emit a nominal type name + possible generic type instantiation here (and also handle the referenced type indirection), instead of calling the more generic `visitType` and then always passing the `emitInstance` flag around when calling `visitType`.

Not sure what's better, but might be worth taking a look.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1163,10 +1175,7 @@ private void visitObjectType(ObjectType type, ObjectType prototype, Boolean proc
       }
       // Prototype fields (mostly methods).
       visitProperties(prototype, false, ((ObjectType) instanceType).getOwnPropertyNames());
-      // Statics.
-      if (processStatics) {
-        visitProperties(type, true);
-      }
+      // Statics are handled in another class.
```
mprobst: ... `in the INSTANCE_CLASS_SUFFIX class`?

src/test/java/com/google/javascript/clutz/ctor_func.d.ts
```diff
@@ -1,5 +1,7 @@
 declare namespace ಠ_ಠ.clutz.ctor_func {
-  class Ctor < T > {
+  class Ctor < T > extends Ctor_Instance < T > {
```
mprobst: Nit: something we could consider in a later commit is checking if a class has any static fields at all, and if not, merge the instance and static classes into just `Ctor` here. I presume that at some point, users will read these `.d.ts` files to try and make sense of their code, so less noise might be nicer.

mprobst:
Nice!

LGTM with a bunch of nits and a possible refactoring.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: Agreed. I did a quick check against the closure ui libs and looks good.

Removed 'public'.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: ack.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: Good catch, tried doing it and found a bug with templatized classes.
At that point we have two cases visited in visitType, so it is not so simple as copying them over. So I am leaving a comment for now.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/test/java/com/google/javascript/clutz/ctor_func.d.ts
rkirov: agree.
It will be a bit tricker, because now if Foo superclasses Ctor, it can't blindly expect _Instance to be there (i.e. we need some global tracking of what _Instance classes exist).

There is another venue of potential optimization - if there is no superclassing of Foo (even with statics) we dont need Foo_Instance.

#196

PR #201, opened 2016-01-04
feat: fix goog.module with default export.

### LINKED ISSUE #186: Support goog.module

rkirov:
Since default export does not have a fixed symbol name at the export
site, Closure introduces synthetic $jscomp namespace, which needs to be
cleaned-up before emitting.

Closes #186

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -30,6 +30,7 @@ import com.google.javascript.rhino.InputId;
 import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.JSDocInfo.Visibility;
 import com.google.javascript.rhino.Node;
+import com.google.javascript.rhino.StaticSourceFile;
 import com.google.javascript.rhino.jstype.EnumElementType;
 import com.google.javascript.rhino.jstype.EnumType;
 import com.google.javascript.rhino.jstype.FunctionType;
@@ -59,6 +60,8 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1112,6 +1124,35 @@ public Void caseProxyObjectType(ProxyObjectType type) {
...
     }
 
+    // When importing a default export it introduces a synthetic type under $jscomp.scope.
+    // import A from './module' -> type.name is $jscomp.scope.A.
+    private String visitSyntheticModuleTypes(ObjectType type) {
+      // It appears that Closure does not have a convenient API to find the module name given the
+      // synthetically generated type for a default export. So we do this the hard way.
+      // From the constructor function, find the raw source file and parse it to extract the
+      // module name.
+      Node source = type.getConstructor().getSource();
```
mprobst: maybe extract a `getSourceCode(JSType)` method?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -944,8 +948,16 @@ public Void caseObjectType(ObjectType type) {
           } else if (type.isDict()) {
             emit("{[key: string]: any}");
           } else if (type.getReferenceName() != null) {
+            String name;
+            // If a class A is declared in goog.module and exported as default, closure
+            // assigns it a synthetic type $jscomp.scope.A.
+            if (type.getReferenceName().contains("$jscomp.scope")) {
```
mprobst: Add a comment on why/how/whether that is different for non-default exports? This seems a bit crazy, maybe it's a thing to raise with Closure? I'd expect to get wonky error messages for users, too (`cannot assign blah to type $jscomp.scope.A` - wtf?).

mprobst:
LGTM.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: The error actually looks like this:
`Property foo never defined on variable$$module$path$to$usage$file of type A$$module$path$to$definition$of$A`, so the path before `.` gets stripped, but the module path gets appended separately to the back.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #203, opened 2016-01-05
fix: do not emit namespaces that collide with static fields in externs.

### LINKED ISSUE #187: emitting namespaces as static fields
The following closure pattern 

```
/**
 * @constructor
 */
function Foo() {}

/** @type {Object}*/
Foo.foo;

/** @type {string}*/
Foo.foo.bar;
```

is currently emitted as

```
class Foo {
  static foo ..
}

namespace Foo.foo {
  var bar: ...
}
```
which is an error in TS.

Real world examples from jquery externs - https://github.com/google/closure-compiler/blob/master/contrib/externs/jquery-1.9.js - `ajaxSettings, fx, support`

rkirov:
Closes #187

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -276,13 +276,17 @@ public class DeclarationGenerator {
       String parentPath = getNamespace(symbol.getName());
       boolean isDefault = isDefaultExport(symbol);
 
+      // skip foo.bar.baz in the following three cases:
+      // * foo.bar is namespace
       if (visitedNamespaces.contains(parentPath)) {
         // Note that this class has been visited before continuing.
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -272,13 +272,17 @@ private void processExternSymbols() {
       String parentPath = getNamespace(symbol.getName());
       boolean isDefault = isDefaultExport(symbol);
 
+      // skip foo.bar.baz if:
+      // * foo.bar is namespace
+      // * foo.bar is class-like and baz is a static field.
+      // * foo is a class-like and foo.bar is a static field.
```
mprobst: maybe move the comments next to the if statements? Might also be interesting to have a note like "these are handled by xxx.", or at least "elsewhere".

mprobst:
LGTM.

PR #211, opened 2016-01-07
fix: do not add _instance for platform externs

### LINKED ISSUE #207: extending from a class in lib.d.ts
When A extends from a browser build-in type like XMLHttpRequest, we generate:
`A_Instance extends XMLHttpRequest_Instance`
which is wrong because `XMLHttpRequest_Instance` doesn't exist.

We should detect types that are not generated by clutz and not append _Instance suffix.

rkirov:
Closes #207


mprobst:
LGTM.

PR #212, opened 2016-01-07
fix: do not emit .prototype as a type.

### LINKED ISSUE #206: do not emit `A.prototype` types
In closure it appear that `A.prototype` is a valid type. In fact we emit it in type position for jquery.fn coming from this declaration

https://github.com/google/closure-compiler/blob/master/contrib/externs/jquery-1.9.js#L867

looks like it is not decided how to type that in a .d.ts:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/jquery/jquery.d.ts#L934

and this case is obscure enough. I suggest just emitting `any` for the `prototype` type.

rkirov:
This can happen very rarely (Foo.prototype is not a valid type symbol in
Closure either), but jquery externs manages to produce it by
setting:

jQuery.fn = jQuery.prototype.

Closes #206

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -982,6 +982,14 @@ public class DeclarationGenerator {
             } else {
               name = getAbsoluteName(type);
             }
+            // Under special conditions (see prototype_inferred_type.js) closure can infer
+            // the type be the prototype object. TypeScript has nothing that matches the shape
+            // of the prototype object (surprisingly typeof A.prototype is A). The best we can do is
+            // any.
+            if (name.endsWith(".prototype")) {
+              emit("any");
+              return null;
+            }
             emit(extendingInstanceClass ?
                 name + INSTANCE_CLASS_SUFFIX : name);
           } else {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -960,6 +960,14 @@ public Void caseObjectType(ObjectType type) {
             } else {
               name = getAbsoluteName(type);
             }
+            // Under special conditions (see prototype_inferred_type.js) closure can infer
+            // the type be the prototype object. Typescript has nothing that matches the shape
```
mprobst: nit: Type[S]cript.

src/resources/third_party_externs.js
```diff
@@ -122,9 +122,4 @@ var ClassNamespace = function() {};
...
  * @const
  */
-ClassNamespace.constObj = {};
-
-/**
- * @type {number}
- */
-ClassNamespace.constObj.num;
\ No newline at end of file
+ClassNamespace.constObj = {};
```
mprobst: Is this change related?

mprobst:
LGTM.

src/resources/third_party_externs.js
rkirov: nope, thanks for catching it.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #213, opened 2016-01-07
fix: introduce PrivateInterface and PrivateClass

### LINKED ISSUE #191: Private type handling must be extended to support private interfaces too
example - https://github.com/google/closure-compiler/blob/master/contrib/externs/jquery-1.9.js#L1411

we emit ``` class Foo implements PrivateType``` but PrivateType is a type alias of void, not an interface.

rkirov:
We already emit PrivateType, but for extend/implement positions an
actual class/interface is needed.

Closes #191


mprobst:
LGTM.

PR #215, opened 2016-01-16
fix: skip emitting private default provides.

rkirov:
Private default provides are not common, but when going through externs
we output each class and interface as a default provide.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -338,7 +338,12 @@ public class DeclarationGenerator {
     emitNamespaceBegin(namespace);
     TreeWalker treeWalker = new TreeWalker(compiler.getTypeRegistry(), provides);
     if (isDefault) {
-      treeWalker.walk(symbol);
+      if (isPrivate(symbol.getJSDocInfo())) {
+        emit("// skipping emitting private default export " + symbol.getName() + ".");
+        emitBreak();
+      } else {
+        treeWalker.walk(symbol);
+      }
     } else {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -338,7 +338,12 @@ private int declareNamespace(String namespace, TypedVar symbol, boolean isDefaul
     emitNamespaceBegin(namespace);
     TreeWalker treeWalker = new TreeWalker(compiler.getTypeRegistry(), provides);
     if (isDefault) {
-      treeWalker.walk(symbol);
+      if (isPrivate(symbol.getJSDocInfo())) {
+        emit("// skipping emitting private default export " + symbol.getName() + ".");
```
mprobst: why is this a `default` export?

src/resources/third_party_externs.js
```diff
@@ -127,4 +127,10 @@ ClassNamespace.constObj = {};
 /**
  * @type {number}
  */
-ClassNamespace.constObj.num;
\ No newline at end of file
+ClassNamespace.constObj.num;
```
mprobst: Please create a second test file for this, this is getting way too long. OK to do in a separate PR.

mprobst:
LGTM.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: as discussed offline, this is because we generate default module export for this symbol.

src/resources/third_party_externs.js
rkirov: will need some new infrastructure. will do in another PR

PR #216, opened 2016-01-19
fix: default exported private values.

rkirov:
Second attempt, completely skipping the value, results in broken code.
Instead emit a field of PrivateType.


mprobst:
LGTM.

Issue #205, opened 2016-01-05
don't skip closure browser apis externs that are not matched by lib.d.ts


rkirov:
for example - https://github.com/google/closure-compiler/blob/master/externs/browser/fileapi.js which is Chrome only and discontinued - http://caniuse.com/#search=filesystem

go through https://github.com/google/closure-compiler/tree/master/externs/browser and add unmatched files to a whitelist.

rkirov:
the emitted list of browser apis is tightened in https://github.com/angular/clutz/commit/4606e52931d7322fb70644e06b72548258aaa490

PR #219, opened 2016-01-21
fix: restrict the platform externs list down.

rkirov:
This is more of a temporary fix. A better solution will follow when we tackle #205.


mprobst:
LGTM.

PR #231, opened 2016-01-30
Declare COMPILED in closure.d.ts

mprobst:
Unlike most Closure defines, `COMPILED` is not namespaced on `goog` (c.f. `goog.DEBUG` and
similar). This special cases `COMPILED` to just be a non-namespaced global variable.

### DIFF
--- a/src/resources/closure.lib.d.ts
+++ b/src/resources/closure.lib.d.ts
@@ -28,3 +28,9 @@ declare namespace ಠ_ಠ.clutz.goog {
 // Closure's goog namespace is accessible as a global symbol without the need for
 // an explicit goog.require, during the closure compiler pass.
 declare var goog: typeof ಠ_ಠ.clutz.goog;
+/**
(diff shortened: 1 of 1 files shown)


src/resources/closure.lib.d.ts
```diff
@@ -28,3 +28,5 @@ declare namespace ಠ_ಠ.clutz.goog {
 // Closure's goog namespace is accessible as a global symbol without the need for
 // an explicit goog.require, during the closure compiler pass.
 declare var goog: typeof ಠ_ಠ.clutz.goog;
+/** Global variable indicating whether the JavaScript code has been compiled. */
```
rkirov: add that it is defined in goog.base, and not explicitly provided.

rkirov:
LGTM

src/resources/closure.lib.d.ts
mprobst: Done.

PR #233, opened 2016-02-03
fix(scope): support class/interface alias.

rkirov:
Instead of producing two incompatible types for the alias, now we emit a
"type alias, constructor var" pair.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -701,7 +701,14 @@ public class DeclarationGenerator {
           visitVarDeclaration(symbol, ftype);
           return;
         }
-        visitClassOrInterface(getUnqualifiedName(symbol), ftype);
+        // The class/interface symbol might be an alias for another symbol.
+        // Since closure inlines all aliases before this step, check against
+        // the type name.
+        if (isAliasedClassOrInterface(symbol, ftype)) {
+          visitClassOrInterface(getUnqualifiedName(symbol), ftype);
+        } else {
+          visitClassOrInterfaceAlias(getUnqualifiedName(symbol), ftype);
+        }
       } else {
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -701,7 +701,14 @@ private void walk(TypedVar symbol) {
           visitVarDeclaration(symbol, ftype);
           return;
         }
-        visitClassOrInterface(getUnqualifiedName(symbol), ftype);
+        // The class/interface symbol might be an alias for another symbol.
+        // Since closure inlines all aliases before this step, check against
+        // the type name.
+        if (symbol.getName().equals(ftype.getDisplayName())) {
```
mprobst: maybe extract a method `isAlias(symbol, ftype)` to make the code more readable?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -730,6 +737,24 @@ private void walk(TypedVar symbol) {
...
+    private void visitClassOrInterfaceAlias(String unqualifiedName, FunctionType ftype) {
+      String typeName = Constants.INTERNAL_NAMESPACE + "." + ftype.getDisplayName();
+      emit("type");
+      emit(unqualifiedName);
+      emit("=");
+      emit(typeName);
+      emit(";");
+      emitBreak();
+      // TS type aliases are only useful in type positions.
+      // To emulate closure alias semantics, introduce also an aliased constructor
```
mprobst: What if the aliased type is an `@interface`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: good catch, added test and fix.

mprobst:
LGTM

PR #234, opened 2016-02-04
fix(interface alias): support templatized types.

rkirov:
Type aliases need to explicitly list their type parameters:

```typescript
class B<T> {}
type A = B;  // wrong
type A<T> = B<T>  // correct
```

### DIFF
--- a/src/test/java/com/google/javascript/clutz/goog_scope_templatized.js
+++ b/src/test/java/com/google/javascript/clutz/goog_scope_templatized.js
@@ -1,0 +1,16 @@
+goog.provide("aliasT.iboom");
+goog.provide("aliasT.I");
+goog.provide("aliasT.I2");
+
+goog.scope(function() {
+  /**
+   * @interface
+   * @template T
+   */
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/clutz/goog_scope_templatized.js
```diff
@@ -0,0 +1,14 @@
+goog.provide("aliasT.iboom");
+goog.provide("aliasT.I");
+
+goog.scope(function() {
+  /**
+   * @interface
+   * @template T
+   */
+  var I = function() { }
+  /** @const */ aliasT.I = I;
```
mprobst: Maybe add a test case for multiple aliasing indirections?

mprobst:
LGTM

src/test/java/com/google/javascript/clutz/goog_scope_templatized.js
rkirov: done

PR #237, opened 2016-02-06
fix: provided functions can be used as namespaces too.

rkirov:
Merge walkInnerSymbol and interface static logic. Emit auxilary
(non-provided) namespace only when non-empty.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -341,7 +341,6 @@ public class DeclarationGenerator {
     TreeWalker treeWalker = new TreeWalker(compiler.getTypeRegistry(), provides, isExtern);
     if (isDefault) {
       if (isPrivate(symbol.getJSDocInfo())) {
-
         treeWalker.emitPrivateValue(symbol);
       } else {
         treeWalker.walk(symbol);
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1382,21 +1352,63 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
           privateEnums.add(innerNamespace + '.' + propName);
         }
       }
```
rkirov: I hate this pattern, but this is the only sure way to emit only when needed. Otherwise the logic is too complicated. Any ideas how to clean it up?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1382,21 +1352,63 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
...
+      boolean nonEmpty = false;
 
       for (String propName : getSortedPublicPropertyNames(type)) {
         String qualifiedName = innerNamespace + '.' + propName;
         if (provides.contains(qualifiedName)) continue;
         JSType pType = type.getPropertyType(propName);
-
         if (pType.isEnumType()) {
+          if (!nonEmpty) {
+            emitNamespaceBegin(innerNamespace);
```
rkirov: I hate this pattern, but its the only foolproof way to guarantee that no empty namespaces get emitted. Any ideas how to clean it up?

mprobst: Couldn't you have some function to check `hasAnyNamespaceMembers(type)`, and inspect that? That duplicates the logic a bit, but not too terribly so?

In any case, I'd call the boolean `foundNamespaceMembers` or so and invert the logic. Then you have `if (!foundNamespaceMembers) { emitNamespace(); foundNamespaceMembers = true; }` which would make this a lot more readable.

mprobst:
LGTM.

Apart from the somewhat ugly namespace emit, this looks like a nice improvement.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: That pattern got us in trouble before, because "NamespaceMembers" is tricky to define (4 branches on type - class / interface / function / enum, then 2 branches on member - function / field). So we had a mismatch where we didn't catch a namespaceMember, but would have emitted it. Also I am still not confident I got it right (closure API is hard - see isNominalConstructor that returns true on structural interfaces ?!?), so rather keep it in one place.

Going with the rename and single lining the emit, for better readability.

PR #238, opened 2016-02-10
fix(type alias): count the type alias as type used.

rkirov:
Adds a test for goog.module + ES6 classes pattern.


calebegg:
LGTM. Thanks!

PR #240, opened 2016-02-11
do not emit empty namespaces.

### LINKED ISSUE #239: avoid emitting empty namespaces
For example when consuming 
https://github.com/google/closure-library/blob/master/closure/goog/i18n/datetimesymbols.js
which contains a large number of untyped goog.provided symbols, clutz produces:
```typescript
declare namespace ಠ_ಠ.clutz.goog.i18n.DateTimePatterns {}
declare module 'goog:goog.i18n.DateTimePatterns' {                               
   import alias = ಠ_ಠ.clutz.goog.i18n.DateTimePatterns;                           
  export default alias;                                                          
}
```

Typescript behaves a bit off when a namespace is empty:
```
namespace N {}
var bar = N;
```
errors with 'cannot find symbol N'. Thus the emitted module is unusable by users. We should be instead emitting:

```typescript
declare namespace ಠ_ಠ.clutz.goog.i18n {
  var DateTimePatterns: Object;
}
declare module 'goog:goog.i18n.DateTimePatterns' {                               
   import alias = ಠ_ಠ.clutz.goog.i18n.DateTimePatterns;                           
  export default alias;                                                          
}
```

A separate question is whether the type of DateTimePatterns should be 'any' instead of 'Object' if we don't want to restrict field access through []-bracket access.

rkirov:
Typescript does not emit empty namespaces, thus correspondingly they
don't typecheck as valid symbols and are not useful to users.

Closes #239

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2,9 +2,6 @@ package com.google.javascript.clutz;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.base.Predicates.equalTo;
-import static com.google.common.base.Predicates.not;
-import static com.google.common.collect.Iterables.any;
 import static com.google.javascript.rhino.jstype.JSTypeNative.ARRAY_TYPE;
 import static com.google.javascript.rhino.jstype.JSTypeNative.OBJECT_TYPE;
 import static java.nio.charset.StandardCharsets.UTF_8;
@@ -59,7 +56,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -315,14 +322,18 @@ private boolean isStaticFieldOrMethod(JSType type) {
...
-  private boolean isDefaultExport(TypedVar symbol) {
-    if (symbol.getType() == null) return false;
-    return !symbol.getType().isObject() ||
-        symbol.getType().isInterface() ||
-        symbol.getType().isInstanceType() ||
-        symbol.getType().isEnumType() ||
-        symbol.getType().isFunctionType() ||
-        isTypedef(symbol.getType());
+  private boolean isDefaultExport(JSType type) {
+    // goog.provides with zero properties are the empty object in closure.
```
mprobst: Maybe extract the comment with the code into its own method? This looks as if it was a comment on `isDefaultExport`, which is confusing.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -610,6 +621,21 @@ private ObjectType getSuperType(FunctionType type) {
     if (implicitProto == null) return null;
     return "Object".equals(implicitProto.getDisplayName()) ? null : implicitProto;
   }
+  private String getUnqualifiedName(TypedVar symbol) {
```
mprobst: nit: empty line above

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

rkirov:
PTAL

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -611,6 +621,22 @@ private ObjectType getSuperType(FunctionType type) {
...
 
+  private String getUnqualifiedName(TypedVar symbol) {
+    return getUnqualifiedName(symbol.getName());
+  }
+
+  private String getUnqualifiedName(String name) {
+    return lastPart(name);
+  }
+
+  private String lastPart(String input) {
```
martine: Is there a better name for this?  E.g. lastDottedComponent or something?  I don't like my name much either but lastPart is so vague.

PR #243, opened 2016-02-17
support optional props in record types.

### LINKED ISSUE #232: Nullable and undefine-able closure types should translate to key? in object literals
For example 

```javascript
/** 
 * @typedef {{
 *   foo: (boolean|undefined),
 * }}
 */
```
translates to 
``` {foo : boolean}```, but should be ```{foo?: boolean}```.

Similar logic was already implemented for optional function arguments.

### PR DESCRIPTION
**Pull request by rkirov**
In typescript the optionality is handled in the name of the property by
adding ? to the name, vs closures approach to keep it purely as a type
of T | undefined.

Closes #232

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2,6 +2,7 @@ package com.google.javascript.clutz;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.Iterables.any;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/main/java/com/google/javascript/clutz/DeclarationGenerator.java [hunk 1]
```diff
@@ -634,6 +635,12 @@ private String lastDottedPart(String input) {
     private final Set<String> provides;
     private int valueSymbolsWalked = 0;
     private final boolean isExtern;
+    private final Predicate<JSType> isUnknownType = new Predicate<JSType>() {
```
mprobst: The naming is a bit odd - unknown vs undefined vs void vs optional. Maybe call this `isVoidType`, and add a comment on why void means optional in unions when translated to TS?

PR #244, opened 2016-02-17
fix: visit fully qualified template types that are used but not provi…

rkirov:
…ded.

Allow for recursive discovery of types used but not provided up to
maximum of 5 iterations.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -111,6 +111,7 @@ public class DeclarationGenerator {
   private final ClutzErrorManager errorManager;
   private StringWriter out = new StringWriter();
   private final Set<String> privateEnums = new LinkedHashSet<>();
+
   /**
    * Aggregates all emitted types, used in a final pass to find types emitted in type position but
    * not declared, possibly due to missing goog.provides.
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -69,6 +69,14 @@
   public static final Pattern GOOG_MODULE_EXTRACT =
       Pattern.compile("goog.module\\(['\"](.*)['\"]\\);");
 
+  /**
+   * Closure types can be used but not provided. Clutz collects them and
+   * visits them in a separate pass. A new set of types can be discovered
+   * while visiting unprovided types, thus recusively iterating. This holds
+   * the maximum depth the tool will visit.
+   */
+  private int maxTypeUsedDepth = 5;
```
mprobst: * this variable is mutable state on what is essentially the global object, and manipulated at a far distance. That's a recipe for nasty bugs in the future, in my experience.
* `5` is obviously an arbitrary number, so Clutz might still fail for some code.
* there's an odd collaboration going on between `processUnprovidedTypes` and `emitTemplatizedTypes` below.

I wonder how prevalent this code pattern is, it seems quite questionable. Can't we fix the code that exhibits the issue?

Alternatively: could we just emit all symbols, unconditionally? That seems to be conceptually what JSCompiler does behind the scenes. Why did we filter those again?

rkirov: I thought java guys accepted the practical usefulness of mutable state :)

In all seriousness, indeed the pattern of collecting types and reemitting later stinks (hence the bugs I just caught). However, it is already in our codebase, so I want to push for this PR as a bug fix. Rolled in the two new fields into function local variables, so the opportunity for abuse is a lot less.

Switching to emitting all symbols is worth investigating and will clean up this particular issue, but might bring some other unknown problems. Issue #245

rkirov:
PTAL

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Ack. Though you forgot to drop the field here?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -254,19 +263,28 @@ private boolean isArrayLike(TypedVar symbol) {
...
-      if (!typesUsed.contains(name)) continue;
-      // skip provided symbols (as default or in an namespace).
-      if (provides.contains(name) || provides.contains(getNamespace(name))) continue;
-
-      // skip extern symbols (they have a separate pass).
-      CompilerInput symbolInput = this.compiler.getInput(new InputId(symbol.getInputName()));
-      if (symbolInput != null && symbolInput.isExtern()) continue;
-      declareNamespace(getNamespace(name), symbol, /* isDefault */ true,
-          Collections.<String>emptySet(), /* isExtern */ false);
+    int maxTypeUsedDepth = 5;
```
mprobst: Move the comment from above as an implementation comment down here (`// ...`)?

mprobst:
LGTM.

rkirov:
merged

Issue #226, opened 2016-01-28
Missing externs in lib.d.ts


mprobst:
```
Cannot find name 'Image'.
Cannot find name 'WebWorker'.
Cannot find name 'WorkerGlobalScope'.
Cannot find name 'RequestCache'.
Cannot find name 'RequestCredentials'.
Cannot find name 'WorkerGlobalScope'.
Cannot find name 'RequestCache'.
Cannot find name 'RequestCredentials'.
Cannot find name 'IDBDatabaseException'.
```

rkirov:
Fixed with https://github.com/angular/clutz/commit/4606e52931d7322fb70644e06b72548258aaa490

Issue #224, opened 2016-01-28
MockStorageInstance


mprobst:
```
Class 'MockStorage_Instance' incorrectly implements interface 'Storage'.
  Index signature is missing in type 'MockStorage_Instance'.
```

rkirov:
We try to be clever and reuse lib.d.ts Storage, but Closure's Storage diverges from lib.d.ts storage. Closure's 'Storage' doesn't have index signature (no way to even express that in closure AFAIK). But when we swap with TS Storage, MockStorage does not implement the interface any longer.

My proposed fix is detach closure types from TS types completely, even for platfom APIs, because as you see here, there are too many incompatible ways to type a platform API.

mprobst:
Completely detaching the platform APIs sounds like it could be a lot of
work, and lead to fragmentation - we'd have Closure APIs not necessarily
agreeing with what lib.d.ts proclaims. For compilation, we also have to
make sure Closure understands when code implements an external interface,
otherwise we risk Closure mangling all the external symbols. And last but
not least we don't currently have a way to communicate to Closure that
lib.d.ts is an extern, i.e. we cannot go from .d.ts to an externs file.

I agree that having two sets of platform APIs and hoping that we can make
them match is kind of crazy, but given these other difficulties, it might
be the lesser evil. Is there a more short term fix we can chose?

I think the equivalent of an index signature in Closure should be extending
`Object<T>` and putting @dict on the object, right?

Rado Kirov <notifications@github.com> schrieb am Do., 18. Feb. 2016 um
03:58 Uhr:

rkirov:
Agree, having the platform APIs typed twice in the same compilation unit is trouble. So is swapping the platform API types underneath closure code that has been written to work with them.

Issues with closures' minification pass seem orthogonal to the discussion at hand. For example, any code that relies on lib.d.ts's File.msClose property is broken because Closures' File interface does not have that property, irrelevant of what clutz does. If anything that points towards not using lib.d.ts at all and using clutz's generated definitions ones, which this PR attempts to do.

Check out the longer doc, I am going to send to the team.

rkirov:
Final solution - modify lib.d.ts to match closure's type shape. Only differences detected were Blob and Storage APIs.

Issue #170, opened 2015-11-26
TypeScript is missing some platform externs (or might use different names)


mprobst:
* [ ] IDBDatabaseException
* [ ] Image

mprobst:
Turns out that `IDBDatabaseException` was [removed from the spec](https://developer.mozilla.org/de/docs/Web/API/IDBDatabaseException), and `Image` is provided in a later TypeScript version, so we just have to upgrade.

rkirov:
fixed with https://github.com/angular/clutz/commit/4606e52931d7322fb70644e06b72548258aaa490

PR #246, opened 2016-02-18
emit all externs

### LINKED ISSUE #205: don't skip closure browser apis externs that are not matched by lib.d.ts

rkirov:
The strategy to not emit platform externs, because they are present in
lib.d.ts is error prone, because closure platform typings can widely
vary from lib.d.ts's versions. Even thought they target the same runtime
API, there are many design decisions (structuring interfaces, enums etc)
that cannot be easily reconciled within the tool.

This commit switches to emitting all closure externs, in order to allow
closures' types to typecheck against them.

Collisions are avoided by making sure all closure types are still in the
internal clutz namespace. That is closure Object interface is
ಠ_ಠ.clutz.Object, vs TS Object that is Object.

This drastic change required some reshuffling of the test
infrastructure. Removing the Syntax test that was a subset of
Usage test. Restrict to only single _with_externs test to run.

Fixes #205, #224, #226

BREAKING_CHANGE:
Previously, TS and Closure's Object were the same interface, thus one
can do:

var myObj: Object;
closureFunction(myObj);

Now this would be a type error because closureFunction expects ಠ_ಠ.clutz.Object.
The easiest workaround for now would be to do `closureFunction(<any> myObj);`

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -281,12 +281,11 @@ public class DeclarationGenerator {
       if (symbolInput == null || !symbolInput.isExtern() || symbol.getType() == null) {
         continue;
       }
-      if (isPlatformExtern(symbolInput.getName())) {
-        continue;
-      }
       // Closure treats all prototypes as separate symbols, but we handle them in conjunction with
       // parent symbol.
       if (symbol.getName().contains(".prototype")) continue;
+      // For is a reserved word.
+      if (symbol.getName().equals("Symbol.for")) continue;
 
       String parentPath = getNamespace(symbol.getName());
       boolean isDefault = isDefaultExport(symbol);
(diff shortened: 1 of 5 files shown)


rkirov:
Clearly this is very rough, but I want to hear your opinion before cleaning it up a bit. Also with its current state this gets us to 2 errors left of all of closure libs, so the hacks served their purpose to prove the concept.

mprobst:
See my comment on the issue, let's keep the discussions there.

rkirov:
I will reply on #224 tomorrow, but keep in mind that this fixes #205, #226 and #170 too.

rkirov:
Superceeded by #256

PR #248, opened 2016-02-19
do not emit invalid property names.

rkirov:
Invalid prop names can only be accessed with index operator, i.e.
foo['0'], which cannot be typed differently for each prop.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -412,6 +412,10 @@ public class DeclarationGenerator {
         String otherName = other.getName();
         if (desiredSymbols.contains(otherName) && other.getType() != null
             && !other.getType().isFunctionPrototypeType() && !isPrototypeMethod(other)) {
+          if (!isValidJSProperty(getUnqualifiedName(other))) {
+            emitComment("skipping property " + otherName + "because it is not a valid symbol.");
+            continue;
+          }
           // For safety we need to special case goog.require to return the empty interface by
           // default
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -412,6 +412,10 @@ public int compare(TypedVar o1, TypedVar o2) {
         String otherName = other.getName();
         if (desiredSymbols.contains(otherName) && other.getType() != null
             && !other.getType().isFunctionPrototypeType() && !isPrototypeMethod(other)) {
+          if (!isValidJSProperty(getUnqualifiedName(other))) {
+            emitComment("skipping property " + otherName + "because it is not valid symbol.");
```
martine: not *a* valid symbol?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -446,6 +450,12 @@ public int compare(TypedVar o1, TypedVar o2) {
     return treeWalker.valueSymbolsWalked;
   }
 
+  private boolean isValidJSProperty(String other) {
```
martine: "other" is a weird name.  Maybe "name" or "symbol"?

martine:
lgtm

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: duh, already spoiled by "extract method", it kept the name of the local var. done

PR #249, opened 2016-02-20
Fix extern namespace.

rkirov:
Externs do not use goog.require so namespaces need to be manually
declared with vars. Depending on the presence of @const there are two
different paths in clutz and for each one we have to make sure not to
generate a var typing that would clash with the namespace.


mprobst:
LGTM.

Are you also sending a CL to fix those files to use `@const`?

rkirov:
Miraculously, the test case with @const just works. We visit the declaration as a namespace (not as a default export), and emit nsConst.I as a member of the namespace. We cannot do anything remotely similar for UnknownTypes was they have no known members.

In all honesty, this ties up our code with closure's type inference, but I don't see a better way of doing it. My only hope is that when we switch to emitting all symbols some of this will be simplified. The test cases are probably the most useful in the wrong run :)

PR #251, opened 2016-02-25
fix: functions that are declared as fields and extended as methods.

rkirov:
Closure does not err in this pattern, but TS is more restrictive and
disallows methods shadowing properties from a base class. To emit
correct TS, we emit never emit a method if it has the same name as a
field on a base class.


martine:
Is it possible to just require people to fix their code?  It seems this will do the wrong thing in the original Closure source, won't it?  (E.g. if the superclass tries to access its field, it will suddenly get a function of the wrong type?)

martine:
LGTM otherwise

rkirov:
I doubt my ability to argue that this pattern is invalid. There is a line of reasoning in which everything is just a property and methods vs fields are artificial concepts. Note that the type could be exactly the same in superclass and subclass (/** @type {function():number} */). The difference is not in the type, but where does the function live. In the superclass it can be on the prototype, but in the subclass it is on the instance. TS does not consider that a true "subtype" relationship.

rkirov:
Added some more logic and test cases, PTAL.

martine:
LGTM++

PR #252, opened 2016-03-01
fixes issue with nested namespace.

### LINKED ISSUE #217: Private classes can have inner classes that are provided but not private
```javascript
goog.provide('foo.A');
goog.provide('foo.A.B');

/**
 * @constructor
 * @private
 */
foo.A = function() {}

/**
 * @constructor
 */
A.B = function() {}
```

produces 

```typescript
declare namespace foo {
  var A: PrivateType;
}
declare namespace foo.A {
  ... 
}
```
which is an error, because of duplicate var-namespace symbol.

Real example - https://github.com/googlei18n/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js#L37

rkirov:
When a foo.bar is provided and we have emitted 'var bar' in namespace
foo, TS does not allow declaring namespace foo.bar. Since all var bar
has no internal references except in module exports, we use an internal
__clutz_alias to avoid the collision.

Note that the module shim still has the correct name, thus the alias in
invisible to any users through ES6 modules.

Closes #217

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Constants.java
+++ b/src/main/java/com/google/javascript/clutz/Constants.java
@@ -7,4 +7,12 @@ public class Constants {
    * disturbed by the extended Unicode character.
    */
   static final String INTERNAL_NAMESPACE = "ಠ_ಠ.clutz";
+
+  /**
+   * When providing foo.bar and foo.bar.Klass, we cannot generate a 'var bar: T' declaration,
+   * because it collides with 'namespace foo.bar' declaration. We alias the first using a postfix.
+   * The alias does not affect the external module declaration, thus the user can still import the
+   * symbol using unaliased module name (i.e. import bar from 'goog:foo.bar').
+   */
+  static final String SYMBOL_ALIAS_POSTFIX = "__clutz_alias";
 }
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -181,7 +181,7 @@ public class DeclarationGenerator {
 
   public String produceDts(Depgraph depgraph) {
     // Tree sets for consistent order.
-    Set<String> provides = new TreeSet<>();
+    TreeSet<String> provides = new TreeSet<>();
     Set<String> transitiveProvides = new TreeSet<>();
     out = new StringWriter();
 
@@ -195,19 +195,25 @@ public class DeclarationGenerator {
       }
     }
 
+    Set<String> shadowedProvides = getShadowedProvides(provides);
+
     TypedScope topScope = compiler.getTopScope();
     for (String provide : provides) {
       TypedVar symbol = topScope.getOwnSlot(provide);
+      String emitName = provide;
+      if (needsAlias(shadowedProvides, provide, symbol)) {
+        emitName += Constants.SYMBOL_ALIAS_POSTFIX;
+      }
       if (symbol == null) {
         // Sometimes goog.provide statements are used as pure markers for dependency management, or
         // the defined provides do not get a symbol because they don't have a proper type.
-        emitNamespaceBegin(getNamespace(provide));
+        emitNamespaceBegin(getNamespace(emitName));
         emit("var");
-        emit(getUnqualifiedName(provide));
+        emit(getUnqualifiedName(emitName));
         emit(": any;");
         emitBreak();
         emitNamespaceEnd();
-        declareModule(provide, true);
+        declareModule(provide, true, emitName);
         continue;
       }
       // ArrayLike is defined in lib.d.ts, so we skip any type alias that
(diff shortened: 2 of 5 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -246,6 +251,19 @@ public String produceDts(Depgraph depgraph) {
     return out.toString();
   }
 
+  private Set<String> getShadowedProvides(Set<String> provides) {
+    Set<String> shadowedProvides = new TreeSet<>();
+    for (String child : provides) {
+      for (String base : provides) {
+        if (child.startsWith(base + ".") && !child.equals(base)) {
```
mprobst: how could child start with base + "." and at the same time equal it?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -246,6 +251,19 @@ public String produceDts(Depgraph depgraph) {
     return out.toString();
   }
 
+  private Set<String> getShadowedProvides(Set<String> provides) {
+    Set<String> shadowedProvides = new TreeSet<>();
```
mprobst: This looks a bit as if a `Trie` was the better data structure, but I guess it doesn't matter for O(100k) symbols?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -306,9 +327,17 @@ private void processExternSymbols() {
...
+      externSymbols.add(symbol);
+      externSymbolNames.add(symbol.getName());
+    }
+
+    Set<String> shadowedSymbols = getShadowedProvides(externSymbolNames);
 
+    for (TypedVar symbol : externSymbols) {
       String parentPath = getNamespace(symbol.getName());
       boolean isDefault = isDefaultExport(symbol);
+      String emitName = symbol.getName() +
```
mprobst: what about not using a ternary but rahter an `if (needsAlias) emitName += ...`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -334,6 +363,21 @@ private void processExternSymbols() {
     }
   }
 
+  private boolean needsAlias(Set<String> shadowedSymbols, String provide, TypedVar symbol) {
+    if (!shadowedSymbols.contains(provide)) return false;
+    // We emit var foo : any for provided but not declared symbols.
```
mprobst: avoid using 1st person speech, `Emit var foo...` instead of `We emit...`

src/main/java/com/google/javascript/clutz/Constants.java
```diff
@@ -7,4 +7,10 @@
    * disturbed by the extended Unicode character.
    */
   static final String INTERNAL_NAMESPACE = "ಠ_ಠ.clutz";
+
+  /**
+   * When providing foo.bar and foo.bar.Klass, we cannot generate a 'var bar: T' declaration,
+   * because it collides with 'namespace foo.bar' declaration. We alias the first using a postfix.
```
mprobst: add docs about how this is still usable from the external module declaration?

mprobst:
Generally LGTMs.

I do think we should think a bit about the potential performance issue with the quadratic `TreeSet` traversal. You could, for each symbol `foo`, see if the range in the `TreeSet` for `foo.` to `foo.\uFFFF` `isEmpty`.

https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#subSet(E,%20E)
https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#isEmpty()

That should make the algorithm `O(n log n)`, which could be substantial assuming we have `O(100k)` symbols here.

mprobst:
I would also like to take this moment and commemorate the occasion where I actually thought about the performance of code in big-O terms in my actual day to day work.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: ops, left over from times when + "." wasn't there. done

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: bad habits from paper writing.

src/main/java/com/google/javascript/clutz/Constants.java
rkirov: done

rkirov:
Good call to use subSet :) I was concerned about the quadratic loop too, but don't know the Java apis that well to replace it with a quick improvement (like subSet.isEmpty).

(btw, `O(100k)` is not valid notation mathematically, you mean `when n ~ 100k` ;))

mprobst:
LGTM.

PR #253, opened 2016-03-01
fix: support up to 52 args in a function.

rkirov:
Looking at you MsPointerEvent.initPointerEvent:
https://github.com/google/closure-compiler/blob/master/externs/browser/ie_event.js#L223

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1366,7 +1366,7 @@ public class DeclarationGenerator {
         Iterable<Node> parameterNodes = functionSource.getFirstChild().getNext().children();
         names = Iterables.transform(parameterNodes, NODE_GET_STRING).iterator();
       }
-      char pName = 'a'; // let's hope for no more than 26 parameters...
+      int paramCount = 0;
       while (parameters.hasNext()) {
         Node param = parameters.next();
         if (param.isVarArgs()) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1366,7 +1366,7 @@ private void visitFunctionParameters(FunctionType ftype, boolean emitTemplatized
         Iterable<Node> parameterNodes = functionSource.getFirstChild().getNext().children();
         names = Iterables.transform(parameterNodes, NODE_GET_STRING).iterator();
       }
-      char pName = 'a'; // let's hope for no more than 26 parameters...
+      char pName = 'a'; // let's hope for no more than 52 parameters...
```
mprobst: Should we just number them as `p1`, `p2`, etc at this point?

rkirov: I thought this is pretty clever extension, without doing the full work :) But sure I will do the little interview problem to implement the proper name.next() :)

mprobst: I think this is just `"p" + (counter++)` ?

rkirov: oh I thought you meant gen'ing 'ab','ac', etc. That's a great fallback, will do.

rkirov:
PTAL

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1375,7 +1375,14 @@ private void visitFunctionParameters(FunctionType ftype, boolean emitTemplatized
         if (names != null && names.hasNext()) {
           emitNoSpace(names.next());
         } else {
-          emitNoSpace("" + pName++);
+          String pName;
+          if (paramCount < 26) {
```
mprobst: I don't understand why you bother with this and not just generate p1, p2, p3, p4, and so on? What's so awesome about 'a', 'b', 'c' in comparison to that? It complicates the code substantially over just `String pName = "p" + (++paramCount);`

rkirov: 4 lines vs 1 line, hardly adding more complexity :)

I think 'a','b','c' reads better than 'p1','p2','p3'

mprobst:
Your choice. LGTM.

PR #254, opened 2016-03-08
Do not emit extern enum values for externs.

rkirov:


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -43,6 +43,7 @@
 import com.google.javascript.rhino.jstype.UnionType;
 import com.google.javascript.rhino.jstype.Visitor;
 
+import com.sun.codemodel.internal.JStatement;
```
martine: Is this needed?

src/test/java/com/google/javascript/clutz/extern_enum.externs.js
```diff
@@ -0,0 +1,4 @@
+var ns = {};
+
+/** @enum */
+ns.Enum = {A: 0}
```
martine: It seems you still emitted a value here... ?

rkirov: The goal was not to emit the value separately too. Previously, we emitted:

declare namespace ಠ_ಠ.clutz.ns.Enum {
  var A: number;
}
along what you see in extern_enum.d.ts

martine:
LGTM

PR #255, opened 2016-03-09
fix shadowing logic.

rkirov:
Previously, "foo.bar" was considered shadowed by "foo.barBuz", which is
incorrect. It is only shadowed by "foo.bar.buz".

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -256,7 +256,7 @@ public class DeclarationGenerator {
   private Set<String> getShadowedProvides(TreeSet<String> provides) {
     Set<String> shadowedProvides = new TreeSet<>();
     for (String provide : provides) {
-      if (!provides.subSet(provide + ".", provide + "\uFFFF").isEmpty()) {
+      if (!provides.subSet(provide + ".", provide + ".\uFFFF").isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -256,7 +256,7 @@ public String produceDts(Depgraph depgraph) {
   private Set<String> getShadowedProvides(TreeSet<String> provides) {
     Set<String> shadowedProvides = new TreeSet<>();
     for (String provide : provides) {
-      if (!provides.subSet(provide + ".", provide + "\uFFFF").isEmpty()) {
+      if (!provides.subSet(provide + ".", provide + ".\uFFFF").isEmpty()) {
```
rkirov: I don't think this deserves a whole test of the type we have been writing. But maybe this is a good reason to add unittests, what do you think?

martine: Sometimes for this sort of thing I'll just sneak it in to an existing regression test.  You could write a unit test for this, but sometimes that sort of unit test will only check for exactly this bug and not whatever the next bug is.

I guess, use your judgement.  :)

rkirov: yeah, I just added extra case to an existing test.

PR #256, opened 2016-03-09
emit every extern except es.*.

rkirov:
Closure platform types significantly differ from TS lib.d.ts types.
After this change we would emit more of closure's view of the platform
in the clutz namespace, which guarantees that exisiting closure code
will not make clutz choke.

The downside is that ts code interfacing with closure can suffer from
this different type view. As a compromise we don't emit ES types -
Array, Object, etc, which are most likely to be passed between TS and
Closure code, and have more consistent types than all of the DOM APIs.

As a byproduct of this change there can be only a single
_with_platform test. Due to how we run Usage tests all go into the same
compilation unit and more than one _with_platform cause repeated
symbols.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -70,6 +70,13 @@
   public static final Pattern GOOG_MODULE_EXTRACT =
       Pattern.compile("goog.module\\(['\"](.*)['\"]\\);");
 
+  public static final Predicate<ObjectType> IARRAYLIKE_PREDICATE = new Predicate<ObjectType>() {
```
mprobst: It's uncommon to have `_TYPENAME` in names in Java. Maybe `IS_IARRAYLIKE`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -117,7 +124,7 @@ public String apply(Node input) {
    * Aggregates all emitted types, used in a final pass to find types emitted in type position but
    * not declared, possibly due to missing goog.provides.
    */
-  private final Set<String> typesUsed = new LinkedHashSet<>();
+  private final Set<String> typesUsed = new TreeSet<>();
```
mprobst: Why?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -381,6 +396,15 @@ private void processExternSymbols() {
     }
   }
 
+  private void sortSymbols(List<TypedVar> symbols) {
+    Collections.sort(symbols, new Comparator<TypedVar>() {
```
mprobst: Take a look at Guava's `Ordering` utility.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -341,14 +353,17 @@ private void processExternSymbols() {
...
+    Iterator<TypedVar> it = externSymbols.iterator();
+    while (it.hasNext()) {
+      if (enumElementSymbols.contains(it.next().getName())) {
+        it.remove();
       }
     }
 
     Set<String> shadowedSymbols = getShadowedProvides(externSymbolNames);
 
+    sortSymbols(externSymbols);
```
mprobst: Why does this need to be sorted?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -409,14 +433,18 @@ private boolean isDefaultExport(TypedVar symbol) {
...
-  // Replace with a list of externs that *only* contain symbols in lib.d.ts.
-  private boolean isPlatformExtern(String name) {
-    name = name.replace("externs.zip//", "");
-    // mostly matching what is in https://github.com/google/closure-compiler/tree/master/externs.
-    return Pattern.matches("javascript/externs/[^/]*.js", name) || name.startsWith("es")
-        || name.startsWith("w3c") || name.startsWith("ie_") || name.startsWith("browser");
+  // Only emit language externs. There are a number of differences between types in externs and
+  // lib.d.ts for browser APIs, so the safest approach is to reemit the closure types for as much
+  // as possible. However, that creates tension at the interface between .ts and closure code,
+  // so we reuse TS types for language (es3-6) features.
```
mprobst: Give an example? `Array` I guess?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -409,14 +433,18 @@ private boolean isDefaultExport(TypedVar symbol) {
...
+  private boolean isLanguageExtern(String filename, String objectName) {
+    // Defined in w3c_dom1, but special cased in Closure internals which makes its type signature
+    // very odd. Looks like we can get away with swapping with TS's Window interface.
+    if (objectName.equals("Window")) return true;
+    // Defined in es6.js, but technically part of HTML spec. Missing from lib.d.ts.
+    if (objectName.equals("Transferable")) return false;
+
+    return Pattern.matches(".*/es.*.js", filename);
```
mprobst: This needs a comment.

mprobst: is this a filename, or a path? filename wouldn't contain the `/`, right?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -929,7 +960,9 @@ private void visitClassOrInterface(String name, FunctionType ftype) {
         emit("implements");
         emitCommaSeparatedInterfaces(it);
       }
-      visitObjectType(ftype, ftype.getPrototype());
+
+      boolean implementsIArrayLike = any(ftype.getOwnImplementedInterfaces(), IARRAYLIKE_PREDICATE);
```
mprobst: Only the own implemented interfaces (immediate implemented?), or also transitively?

src/test/java/com/google/javascript/clutz/DeclarationSyntaxTest.java
```diff
@@ -31,7 +31,7 @@
   public static void setUpTsc() throws Exception {
     if (!TSC.toFile().exists()) {
       System.err.println("Installing typescript...");
-      runChecked("npm", "install", "typescript@1.6.2");
+      runChecked("npm", "install", "typescript@1.8.0");
```
mprobst: should this be a separate change?

src/test/java/com/google/javascript/clutz/types_externs_with_platform.js
```diff
@@ -3,6 +3,7 @@ goog.provide('typesWithExterns.A');
 goog.provide('typesWithExterns.B');
 goog.provide('typesWithExterns.C');
 goog.provide('typesWithExterns.ArrayLike');
+goog.provide('type_renaming_with_externs');
 
 /**
```
mprobst: Do you have a test for this?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: Our tests run on JDK7, 8 and it seems that closure returns compiler.topSymbols() in different order between the two versions. We have only one golden file so the option is introduce two goldens or sort here.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: Adding fixed iteration order to make our tests run on JDK7 and 8 (see comment below). Turns out this was not the offending set, but I still see no reason not to use TreeSet everywhere.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: changed to getAllImplementedInterfaces()

src/test/java/com/google/javascript/clutz/DeclarationSyntaxTest.java
rkirov: sure

src/test/java/com/google/javascript/clutz/types_externs_with_platform.js
rkirov: This is just an existing test moved to a single _with_platform.js, due to how we run our Usage tests.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

rkirov:
removed es6.js from the excluded list. That means, we would generate clutz types from closure and not reuse lib.d.ts. PTAL

src/test/java/com/google/javascript/clutz/types_externs_with_platform.js
```diff
@@ -3,6 +3,7 @@ goog.provide('typesWithExterns.A');
 goog.provide('typesWithExterns.B');
```
mprobst: Could you split up `types_externs_with_platform.d.ts` into a part that's just the giant golden file and a part that has the bits we care about? This is going to be a maintenance nightmare otherwise.

src/test/java/com/google/javascript/clutz/types_externs_with_platform_usage.ts
```diff
@@ -1,4 +1,5 @@
 import {elementMaybe, Error} from 'goog:typesWithExterns';
 
-var el: Element = elementMaybe();
+//!! TS Element disagrees with Closure's Element because of missing msContentZoomFactor field.
+var el: Element = <any> (elementMaybe());
```
mprobst: As discussed offline, making the DOM incompatible seems like a very bad outcome. We should reconsider how hard it'd be to make `lib.d.ts` and Closure's externs match.

rkirov:
closing due to preferred approach of changing lib.d.ts to match closure externs and emit the rest of closure: see  https://github.com/angular/clutz/commit/4606e52931d7322fb70644e06b72548258aaa490

PR #258, opened 2016-03-10
emit an index signature for classes implementing IArrayLike.

rkirov:
The parametrized type of IArrayLike is not handled (emit as any).

Fixes a bug where interfaces picked up static methods from backing
function assignment.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -65,7 +65,12 @@ import javax.annotation.Nullable;
  * A tool that generates {@code .d.ts} declarations from a Google Closure JavaScript program.
  */
 public class DeclarationGenerator {
-
+  private static final Predicate<? super ObjectType> IS_IARRYLIKE = new Predicate<ObjectType>() {
+    @Override
+    public boolean apply(@Nullable ObjectType input) {
+      return input != null && input.getDisplayName().equals("IArrayLike");
+    }
+  };
   static final String INSTANCE_CLASS_SUFFIX = "_Instance";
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1708,6 +1719,11 @@ public Void caseFunctionType(FunctionType type) {
     }
   }
 
+  private boolean isAFunctionPrototypeProp(String propName) {
```
mprobst: drop the `isA`, just `is`? Sounds more Java-ish.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1708,6 +1719,11 @@ public Void caseFunctionType(FunctionType type) {
...
+    return propName.equals("apply") || propName.equals("call") ||
```
mprobst: This is non-complete, right? Maybe add a comment?

Also, wouldn't an alternative implementation be checking if the declaration location is the Function prototype? That's probably a lot harder, so just asking here.

Also: might be easier to write as a switch statement on propName? Java supports Strings in switches now, AFAIK.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: This is the complete list, that I was seeing emitted. My guess would be those should be filtered, because we asked for get*Own*PropertyNames, but somehow they still appear.

mprobst:
LGTM

PR #260, opened 2016-03-11
Add Event to the list of globalSymbolAliases.

rkirov:
Clutz will emit GlobalEvent instead of Event, where GlobalEvent is an
type alias for Event.

This allows the following pattern to work:

namespace goog {
  class Event extends Event {}
}


rkirov:
Didn't add a test because this is equivalent to the pattern for Error, which has a test.

PR #262, opened 2016-03-14
Fix field-method mismatch for inferred properties.

rkirov:
It is possible that this is also a bug in closure, to consider the
inferred property as member of the derived class, and not part of the subclass.
In any case, clutz should generate a valid TS.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/extending_function_field.d.ts
+++ b/src/test/java/com/google/javascript/clutz/extending_function_field.d.ts
@@ -1,0 +1,42 @@
+declare namespace ಠ_ಠ.clutz.a {
+  function messesWithB (b : ಠ_ಠ.clutz.ns.B ) : void ;
+}
+declare namespace ಠ_ಠ.clutz.goog {
+  function require(name: 'a.messesWithB'): typeof ಠ_ಠ.clutz.a.messesWithB;
+}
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/extending_function_field.d.ts
```diff
@@ -0,0 +1,31 @@
...
+  class A extends A_Instance {
+  }
+  class A_Instance {
+    private noStructuralTyping_: any;
+    fn : ( ...a : any [] ) => any ;
+  }
+  class B extends B_Instance {
+  }
+  class B_Instance extends ಠ_ಠ.clutz.ext.fn.Base.A_Instance {
+    //!! IHMO, it is a bug in closure that this property even appears.
```
mprobst: nit: IMHO

src/test/java/com/google/javascript/clutz/extending_function_field.js
```diff
@@ -0,0 +1,30 @@
+goog.provide('ext.fn.Base');
+
+/**
+ * @constructor
+ */
+ext.fn.Base = function() {};
```
mprobst: nit: rename `ext.fn.Base` to just a static method `screwUpTypeInference`.

mprobst:
LGTM

PR #267, opened 2016-04-07
Avoid ConcurrentModificationException.

rehmsen:
It would probably be ideal to additionally test this, but the method itself is private and the calling method does not seem to have any tests yet. Please advice. 

Also, it seems subideal to have to iterate linearly through `externSymbols`, but that was already the case and I did not want to make this change larger than necessary.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -382,9 +382,10 @@ public class DeclarationGenerator {
 
     // Enum values like Enum.A will appear as stand-alone symbols, but we do not need to emit them.
     externSymbolNames.removeAll(enumElementSymbols);
-    for (TypedVar symbol : externSymbols) {
-      if (enumElementSymbols.contains(symbol.getName())) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -382,11 +382,13 @@ private void processExternSymbols() {
 
     // Enum values like Enum.A will appear as stand-alone symbols, but we do not need to emit them.
     externSymbolNames.removeAll(enumElementSymbols);
+    List<TypedVar> externSymbolsToRemove = new ArrayList<>();
```
mprobst: What about using the iterator explicitly to remove elements?

```java
Iterator<> it = externSymbols.iterator();
while (it.hasNext()) {
  if (eES.contains(it.next().getName()) {
    it.remove();
  }
}
```

mprobst:
BTW I'm fine to do this without an integration test that exercises this, but it seems like we should improve test coverage here (@rkirov).

rehmsen:
Thanks for the suggestion. I changed it to use iterators and squashed my commits.

PR #271, opened 2016-04-11
Allow conversion of platform externs.

jDramaix:
Add an option allowPlatformExterns that allows clutz to explicitly convert platform externs.


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -43,6 +43,10 @@
       usage = "run faster by skipping the externs parsing (useful for tests)")
   boolean skipParseExterns;
 
+  @Option(name = "--allowPlatformExterns",
```
mprobst: Maybe `--emitPlatformExterns`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1254,27 +1255,31 @@ private Void emitTemplatizedType(TemplatizedType type, boolean extendingInstance
...
-        case "IThenable":
-          templateTypeName = "PromiseLike";
-          break;
-        case "IArrayLike":
-          templateTypeName = "ArrayLike";
-          break;
-        default:
-          break;
+
+      if (!opts.allowPlatformExterns) {
```
mprobst: Could you add a test for this behaviour?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -979,8 +979,9 @@ private void visitClassOrInterface(String name, FunctionType ftype) {
           // TypeScript does not allow public APIs that expose non-exported/private types.
           emit(Constants.INTERNAL_NAMESPACE + ".PrivateClass");
         } else {
-          Visitor<Void> visitor = new ExtendsImplementsTypeVisitor(
-              emitInstance && !isDefinedInPlatformExterns(superType));
+          boolean emitInstanceForObject =  emitInstance &&
```
mprobst: whitespace is a bit off here

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: +1 let me know if you need help with the testing infra.

It would be quite a large output, because if you don't pass externs the behavior is indistingushable, but if you do the output will contain all externs. Maybe you can think of a way to mock out the externs to only a few files instead of all platform externs. If not, I will be ok with a largish output for now.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
jDramaix: Done

src/main/java/com/google/javascript/clutz/Options.java
jDramaix: Done

jDramaix:
I have a question for you guys. If I want to convert only platform externs, do I have to pass the externs files as arguments or in the --externs parameters ?

1) java com.google.javascript.clutz.DeclarationGenerator  --emitPlatformExterns  es3.js es5.js es6.js

2) java com.google.javascript.clutz.DeclarationGenerator --externs es3.js es5.js es6.js  --emitPlatformExterns  empty.js

The solution 1) needs another code modification because clutz adds by default built-in externs 
The solution 2) looks weird because you have to pass an empty js file.

jDramaix:
I've added a test covering type renaming.
PTAL

jDramaix:
I have to fix DeclarationSyntaxTest first

src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -43,6 +43,10 @@
       usage = "run faster by skipping the externs parsing (useful for tests)")
   boolean skipParseExterns;
 
+  @Option(name = "--emitPlatformExterns",
+      usage = "emits platform externs.")
```
rkirov: clarify a bit more "emits platform externs, instead of omitting them in favor of TS lib.d.ts".

src/test/java/com/google/javascript/clutz/type_renaming_with_emit_platform_externs.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * @fileoverview Test that during the conversion of platform externs, the types below
+ * aren't renamed
+ * @externs
```
rkirov: this is confusion, the file is not an extern, but you are adding @externs annotation.

src/test/java/com/google/javascript/clutz/DeclarationSyntaxTest.java
```diff
@@ -37,29 +65,65 @@ public static void setUpTsc() throws Exception {
...
+  }
+
+  private void doTestDeclarationUsage(FilenameFilter filenameFilter, boolean noLib) throws Exception {
+    List<File> inputs = DeclarationGeneratorTests.getTestInputFiles(filenameFilter);
     final List<String> tscCommand =
         Lists.newArrayList(TSC.toString(), "--noEmit", "--skipDefaultLibCheck", "--noImplicitAny",
             "-m", "commonjs");
-    tscCommand.add("src/resources/closure.lib.d.ts");
+
+    if (noLib) {
```
rkirov: repeated code, extract to helper.

src/test/java/com/google/javascript/clutz/type_renaming_with_emit_platform_externs.js
rkirov: I think it will be cleaner if this file was "intentionally empty", and the ".externs.js" contains the current contents.

src/resources/min-lib.d.ts
```diff
@@ -0,0 +1,19 @@
+// Typescript compiler requires the types below to be defined.
```
rkirov: why is this needed? Clutz always emits within a o_o.clutz namespace, so there should be no collision with `lib.d.ts` and your output.

I want to keep the invariant that clutz' output has to always be consumed together with 'closure.lid.d.ts' and 'lib.d.ts'. Otherwise, we will get into inconsistencies in the output. For example currently we output `clutz.PrivateType` alias for private types, and that is defined in 'closure.lib.d.ts'.

rkirov:
I prefer option 2) because it is more uniform with the current usage and avoids clutz making the call on what is extern and what is not. It shouldn't be too hard to support not needing `empty.js` at all (in a separate pr).

LGTM, in general, just a few nitpicks.

src/resources/min-lib.d.ts
jDramaix: You're right I was confused by the fact that all d.ts files are passed to the compiler and as I redefine NodeList as parametrized type, the compilation of others d.ts file failed. 
Separate the tests based on the fact they use lib.d.ts or not is enough to avoid conflict

src/main/java/com/google/javascript/clutz/Options.java
jDramaix: done

src/test/java/com/google/javascript/clutz/type_renaming_with_emit_platform_externs.js
jDramaix: agreed, done

src/test/java/com/google/javascript/clutz/DeclarationSyntaxTest.java
jDramaix: code removed

rkirov:
squashed and merged.

PR #274, opened 2016-04-16
Handle reserved words gracefully.

### LINKED ISSUE #270: Support variables named like reserved keywords
I encountered this when trying to use Clutz on _ng.material.components.switch_. The *.d.ts produced by Clutz contains:

```
declare namespace ಠ_ಠ.clutz.ng.material.components {
  var switch : any;
}
```

And then the Typescript compiler throws

> java/com/google/developers/console/web/crash/common/_common_ts_js_typings.d.ts:155302:7: Variable declaration expected.

rkirov:
Closes #270

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -99,7 +99,20 @@ public class DeclarationGenerator {
       "window.js");
 
   /**
-   * List of global platfom platform symbols that are redirected through an alias in closure.lib.d.ts
+   * Reserved words in ES6.1.
+   * Thanks to https://mathiasbynens.be/notes/reserved-keywords.
+   */
+  private static final Set<String> reservedJsWords = ImmutableSet.of(
+      "arguments", "await", "break", "case", "catch", "class", "const",
+      "continue", "debugger", "default", "delete", "do", "else", "enum", "eval",
+      "export", "extends", "false", "finally", "for", "function", "if",
+      "implements", "import", "in", "instanceof", "interface", "let", "new",
+      "null", "package", "private", "protected", "public", "return", "static",
+      "super", "switch", "this", "throw", "true", "try", "typeof", "var",
+      "void", "while", "with", "yield");
+
+  /**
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -99,6 +99,19 @@ public boolean apply(@Nullable ObjectType input) {
...
+   * Thanks to https://mathiasbynens.be/notes/reserved-keywords.
+   */
+  private static final Set<String> reservedJsWords = ImmutableSet.of(
+      "do", "if", "in", "for", "let", "new", "try", "var", "case", "else",
+      "enum", "eval", "null", "this", "true", "void", "with", "await", "break",
+      "catch", "class", "const", "false", "super", "throw", "while", "yield",
+      "delete", "export", "import", "public", "return", "static", "switch",
+      "typeof", "default", "extends", "finally", "package", "private",
+      "continue", "debugger", "function", "arguments", "interface",
+      "protected", "implements", "instanceof");
```
martine: Alphabetize?  ¯\_(ツ)_/¯

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1107,7 +1190,7 @@ private void visitTypeAsPrimary(JSType type) {
       }
     }
 
-    private void visitType(JSType typeToVisit) {
+    void visitType(JSType typeToVisit) {
```
martine: Is this necessary?

martine:
As discussed offline, LGTM.

(We don't always want to do this transformation because namespaces are open, and we can't use assignments to build up the namespaces out of objects because this is d.ts file.)

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: nope, it's a inner class, so I can access all methods. forgot my java (never knew it anyways).

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: sure. done.

PR #279, opened 2016-04-21
Upgrade closure-compiler to the last version.

jDramaix:


jDramaix:
The test fails because there is a new check in the closure compiler that checks that externs definition for Symbol has been provided.

I've sent a mail to the closure team to ask if it's possible to disable this check.

Otherwise two solutions:
- we don't skip default externs file parsing for tests
- when we skip externs parsing, we add a minimal externs file that define `Symbol`

jDramaix:
AFAIC, the second option is the beste one.

jDramaix:
I've implemented the second one, PTAL

src/test/java/com/google/javascript/clutz/types_externs_with_platform.d.ts
```diff
@@ -132,3 +132,16 @@ declare namespace ಠ_ಠ.clutz.namespace.subNamespace {
 declare namespace ಠ_ಠ.clutz.namespace.subNamespace {
   var fieldB : number ;
 }
+declare namespace ಠ_ಠ.clutz.window {
```
jDramaix: I'm not really sure why I had to add this part...

rkirov: I have seen this behavior too. Under some conditions (externs present), closure starts doubling all global symbols - foo and "window.foo". I am only guessing, but I think es6_min, triggered this behavior. In any case, I replaced this output with a check to not emit anything starting with 'window.'

rkirov:
I made a few trivial changes (added a few comments), squashed and merged this on master. Thanks for contributing!

PR #285, opened 2016-05-27
Best management of IndexSignature.

jDramaix:
The IndexSignature is supported in closure by implementing IObject interface. IObject has two template types, the first one determines the type of the key and the second one determines the type of the returned value.

Note that `IArrayLike<T>` extends `IObject<number, T>` but closure compiler considers these two interfaces as being independents and a Type in closure can implements both interfaces. We have to process them separately.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -59,6 +59,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -977,7 +978,6 @@ public class DeclarationGenerator {
       }
     }
 
-
     /**
      * Used to differentiate a function with a constructor function type, from ordinary ones.
      * <code> @type {function(new:X)} foo.x;</code>
@@ -1561,7 +1561,7 @@ public class DeclarationGenerator {
       // IArrayLike<T> extends IObject<number, T>. Normally only looking for IObject interface
       // should be enough. But the closure compiler seems to process these two interfaces as if they
       // were independent. A type can even implements both.
-      for (ObjectType implementedInterface : type.getAllImplementedInterfaces()) {
+      for (ObjectType implementedInterface : getAllDirectlyImplementedInterfaces(type)) {
         String displayName = implementedInterface.getDisplayName();
         if ("IObject".equals(displayName)) {
           List<JSType> iObjectTemplateTypes = implementedInterface.getTemplateTypes();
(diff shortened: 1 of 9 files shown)


jDramaix:
The index signature is only defined on the top class in the hierarchy that implements `IArrayLike` or `IObject` (or any interfaces extending these interfaces) not on the sub classes.

Tests are covering that.

jDramaix:
fix issue 278

src/test/java/com/google/javascript/clutz/index_signature_with_platform.js
```diff
@@ -0,0 +1,66 @@
+goog.provide('index_signature');
+
+/**
+ * @constructor
+ * @return {!index_signature.SomeType}
```
rkirov: no need to add @return, @constructor is enough.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1559,6 +1583,30 @@ private void visitObjectType(FunctionType type, ObjectType prototype,
...
+
+      for (ObjectType implementedInterface : type.getOwnImplementedInterfaces()) {
+        addRelatedInterfaces(implementedInterface, interfaces);
+      }
+      return interfaces;
+    }
+
+    private void addRelatedInterfaces(ObjectType instance, Set<ObjectType> interfaces) {
+      FunctionType constructor = instance.getConstructor();
+      if (constructor != null && constructor.isInterface() && !interfaces.contains(instance)) {
```
rkirov: `interfaces` is a set already, I don't see the point of checking `contains` before adding, just add it.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1559,6 +1583,30 @@ private void visitObjectType(FunctionType type, ObjectType prototype,
       emitBreak();
     }
 
+    /**
+     * Returns all interfaces implemented by a class and any
+     * superclasses for any of those interfaces.
```
rkirov: `superclass` -> `superinterface`

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
jDramaix: it's for avoiding the for loop if the interface has already been processed

src/test/java/com/google/javascript/clutz/index_signature_with_platform.js
jDramaix: Done

jDramaix:
PTAL

src/resources/closure.lib.d.ts
```diff
@@ -20,6 +20,8 @@ declare namespace ಠ_ಠ.clutz {
    * type position PrivateType is used.
    */
   interface PrivateInterface {}
+
+  interface IObject<KEY1, VALUE> {}
```
rkirov: I was wondering why can't we make this guy ```IObject<KEY1, VALUE> { [key: KEY1]: VALUE}``` and greatly simplify the logic, but TS does not support generic types for index signatures.

http://www.typescriptlang.org/play/#src=interface%20IObject%3CKEY1%2C%20VALUE%3E%20%7B%0A%09%5Bk%3A%20KEY1%5D%20%3A%20VALUE%3B%0A%7D

I will add a comment to remind myself that as I merge the cl into master.

rkirov:
LGTM.

rkirov:
merged

Issue #288, opened 2016-06-14
gradle bulid fails because it depends on npm, but clutz has no package.json


jjudd:
Hi! Trying out clutz. As suggested by the readme, I did

```
git clone git@github.com:angular/clutz.git
cd clutz
gradle build
```

This fails because npm cannot install typescript. There are errors like
```
com.google.javascript.clutz.DeclarationSyntaxTest > testDeclarationUsage FAILED
    java.io.IOException: Cannot run program "node_modules/typescript/bin/tsc": error=2, No such file or directory
        at java.lang.ProcessBuilder.start(ProcessBuilder.java:1048)
        at com.google.javascript.clutz.DeclarationSyntaxTest.runChecked(DeclarationSyntaxTest.java:129)
        at com.google.javascript.clutz.DeclarationSyntaxTest.doTestDeclarationUsage(DeclarationSyntaxTest.java:119)
        at com.google.javascript.clutz.DeclarationSyntaxTest.testDeclarationUsage(DeclarationSyntaxTest.java:100)
```

Adding a minimal `package.json` at the root fixes the issue.

```
{
  "name": "clutz",
  "version": "0.0.0"
}
```

An alternate workaround is building without tests `gradle build -x test`

Not sure how we want to go about fixing this. I'll make a PR for the package.json, but if we want to do something else, then what should that be?

PR #289, opened 2016-06-14
Adding a package.json, so that gradle build succeeds.

jjudd:
Some of the tests depend on the npm TypeScript package being installed.
There is no package.json, so npm fails to install TypeScript. This adds
a very minimal pakage.json, so that npm can install TypeScript and gradle
build can succeed.

Resolves: #288


rkirov:
merged, thanks for the contribution!

PR #293, opened 2016-06-20
Adding missing comma in package.json

jjudd:


rkirov:
ops, that was my bad, modifying your PR as submitting.

PR #300, opened 2016-06-30
Ignore private annotation for constructors

lambertjamesd:
I ended up adding an `isConstructor` check in many places where there is an `isPrivate` check. I also made it so it doesn't emit the constructor if it is private. It doesn't prevent calling new on a private class but I felt that was the logic thing to do until typescript supports private constructors.


mprobst:
@lambertjamesd is it possible that your local git client doesn't have your email address properly set up? I think that's why googlebot complains.

rkirov:
merged

PR #308, opened 2016-07-25
Emit optional properties if Closure annotations have undefined in union type

### LINKED ISSUE #306: Generate optional properties from `<type>|undefined`
In our closure typedefs, we use `type|undefined` to represent an object property that is optional.
An example is the "parent" property here:

```
goog.provide('organization.OrganizationData');

/** @record */
organization.OrganizationData = function() {};

/** @export {string} */
organization.OrganizationData.prototype.key;

/** @export {string} */
organization.OrganizationData.prototype.name;

/** @export {!organization.OrganizationData|undefined} */
organization.OrganizationData.prototype.parent;
```

When clutz runs over this file, it just copies over the "| undefined" and produces:
```
  interface OrganizationData {
    key : string ;
    name : string ;
    parent : ಠ_ಠ.clutz.organization.OrganizationData | undefined ;
  }
```
When I create a variable of type OrganizationData, typescript insists that I include the optional properties as well. I think it would be better for clutz to special-case this and output optional properties - like "parent?" - instead.

polybuildr:
Resolve #306.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1816,7 +1816,13 @@ public class DeclarationGenerator {
       if (isStatic) emit("static");
       emit(propName);
       if (!propertyType.isFunctionType() || forcePropDeclaration) {
-        visitTypeDeclaration(propertyType, false, false);
+        UnionType unionType = propertyType.toMaybeUnionType();
+        boolean isOptionalProperty = false;
+        if (unionType != null && any(unionType.getAlternates(), isVoidType)) {
+          emit("?");
+          isOptionalProperty = true;
+        }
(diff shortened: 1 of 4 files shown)


polybuildr:
This works for my use case, @gmoothart does it work fine for yours too?

I'm not sure this is the proper approach, though or at feel that least that some further changes might need to be made.

polybuildr:
The tests are failing since the output no longer matches, but I'll wait for a review before I update them to match.

rkirov:
looks great, thanks for contributing! 

Note that you only need to change `general_with_platform.d.ts` for all the platform typings changes. Also separately you should add a small test just for this feature.

polybuildr:
I think I fixed the failing test and added a test case to check for this specifically. Should be ready for review now. :) Thanks for the help @rkirov!

rkirov:
lgmt, merged

PR #337, opened 2016-09-02
Replace getType with getToken since it will be deprecated

tadeegan:


tadeegan:
Messed up the email on the commits.

PR #347, opened 2016-09-22
feat: use system clang-format unless testing.

mprobst:
When running for real, gents should use the local clang-format that the user is using.
Only during tests we should use the node_modules installed version.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -50,6 +50,11 @@ dependencies {
   testCompile 'com.google.truth:truth:0.25'
 }
 
+tasks.withType(Test) {
+  systemProperty 'gents.clangFormat',
+                 System.getProperty('gents.clangFormat', 'node_modules/.bin/clang-format')
+}
+
 test {
   testLogging {
     exceptionFormat = 'full'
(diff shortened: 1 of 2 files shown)


gmoothart:
I wonder if this is best. the system (google3) clang-format could differ from the version supplied by npm.

I think it might be better to just require you to `npm install clang-format` if you want to run the gents tests. Then both local dev and jenkins are relying on the same clang.

rehmsen:
Is it not okay for the clang formats to differ?

When actually using gents, you want to have the clang format that you use
for all your other code, so if you have a special version installed, that
is what you want.

When running the tests, you don't really care how it's formatted, it just
has to match the expectation in the test case and thus be stable. So it is
better to isolate from your system configuration and use any one fixed
version. Actually, it would probably be good to not install latest but some
fixed version, to not break the tests when a new clang format is released
(which I believe happened with my PR that first passed and layer failed the
build). You could than update that manually every now and then and I update
the tests along with it.

mprobst:
@gmoothart the intent of this change is to do just that, always use the `npm` supplied `clang-format` when running tests, and only use the system `clang-format` when running for real.

The only problem is that the build file of course only affects `gradle` runs on the command line, so when you run the tests from your IDE, you'll have to manually set the `gents.clangFormat` property in your IDE config. I'm not aware of a way to have e.g. Eclipse pick that up from gradle, sadly :-/

gmoothart:
Ah yes, I misunderstood. This is fine

Issue #345, opened 2016-09-21
fields without type information should still be promoted to TS class fields


gmoothart:
Fields defined in a constructor that don't have an explicit type annotation are skipped by gents. There are two issues here:

- If we cannot infer a type we should just type to `any`, but we should add the class field anyway
- If the type is easily inferred (like the example below) we should do so.

Example:
```
/**
 * @constructor
 */
var A = function() {
  /** @const */
  this.w = 5;
};
```

should be converted to:
```
class A {
  w: number = 5;
}
```

but currently is:
```
class A {
  constructor() { this.w = 5; }
}
```

dpurp:
This is partially fixed; however, the type of the outputted property value is currently always set to `any`.

mprobst:
We'd have to run type inference for that, and prefer using types determined
by inference over the syntactical types. That might end up being tricky -
it'd require figuring out that the current syntactical symbol correlates to
some field of a type. That's messy, in particular for goog.module. Might
not be worth it?
-- 

Martin

gmoothart:
But this is a very restricted case, we only promote field initializers if the assignment is to a literal: https://github.com/angular/clutz/blob/52f0713977411a64e4fdedb71bd4e5e82d5c3bdb/src/main/java/com/google/javascript/gents/TypeConversionPass.java#L423

It seems like it shouldn't be that hard to construct the right JSTypeExpression if we know that the rhs is a NUMBER, STRING, etc.

gmoothart:
One option is to just omit the type annotation entirely and fall back on TypeScript inference. i.e. `private n = 5;` instead of today's `private n: any = 5;`

dpurp:
Note that the type annotations appear to be added on the ClosureCompiler side of the fence.

https://github.com/google/closure-compiler/blob/005278687e8d51db7aee409c7631e93a20d3578b/src/com/google/javascript/jscomp/CodeGenerator.java#L262

PR #351, opened 2016-09-22
emit all constructor fields as class fields

gmoothart:
Even fields which don't have type information should be emitted as class fields.

Partially addresses #345

Migrating from phabricator

### DIFF
--- a/src/main/java/com/google/javascript/gents/ClassConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ClassConversionPass.java
@@ -109,27 +109,29 @@ public final class ClassConversionPass implements CompilerPass {
     public void visit(NodeTraversal t, Node n, Node parent) {
       if (n.isExprResult()) {
         ClassMemberDeclaration declaration = ClassMemberDeclaration.newDeclarationOnThis(n);
-        // Ignore field declarations without a type annotation
-        if (declaration != null &&
-            declaration.jsDoc != null &&
-            declaration.jsDoc.getType() != null) {
-          Node fnNode = NodeUtil.getEnclosingFunction(n);
-          String fnName = fnNode.getParent().getString();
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/gents/ClassConversionPass.java
```diff
@@ -109,27 +109,29 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-              // Names and types must be equal
-              if (declaration.memberName.equals(param.getString()) && type.equals(paramType)) {
-                // Add visibility directly to param if possible
-                moveAccessModifier(declaration, param);
-                n.detachFromParent();
-                compiler.reportCodeChange();
-                return;
+
+        if (declaration != null) {
+          // TODO(gmoothart): in many cases we should be able to infer the type from the rhs
```
gmoothart: I played around with this a little but couldn't get anything working. I couldn't figure out how to convert (for example) a rhs that is a NUMBER to a JSTypeExpression needed by the code below.

There are also cases like
```
this.a = getFoo();
```
where we know the return type of getFoo. And in this case too I couldn't figure out how to get the AST to give me what I wanted.

mprobst:
clang-format strikes again! Looks good otherwise.

src/main/java/com/google/javascript/gents/ClassConversionPass.java
mprobst: That'd be a totally different API to obtain types, I think.

One thing you could consider with this code here is to restructure it to use early exits by extracting a method. I.e. check `if declaration == null) return;`, then `if (declaration.jsDoc == null) ... skip other part ...`, and so on.

gmoothart: added an early exit.

PR #353, opened 2016-09-23
gents accepts a json file with an extern -> type mapping

### LINKED ISSUE #336: gents extern to typing mappings
gents translates externs into typings very literally:

```
/**
 * @return {!angular.Directive}
 */
const fabMenuOptionDirective = function() {

```
becomes

`function fabMenuOptionDirective(): angular.Directive {`

when the type should be angular.IDirective (or ng.IDirective). It would be nice if gents maintained a map of common externs and their typing equivalent.

gmoothart:
gents now allows the user to pass a json file with an extern/typing map
for the cases where there are spelling differences between the two.

fixes #336

### DIFF
--- a/src/main/java/com/google/javascript/gents/Options.java
+++ b/src/main/java/com/google/javascript/gents/Options.java
@@ -99,7 +99,7 @@ public class Options {
     }
   }
 
-  Options(String[] args) throws CmdLineException, FileNotFoundException {
+  Options(String[] args) throws CmdLineException {
     CmdLineParser parser = new CmdLineParser(this);
     parser.parseArgument(args);
     srcFiles.addAll(arguments);
@@ -109,7 +109,12 @@ public class Options {
       throw new CmdLineException(parser, "No files were given");
     }
 
-    externsMap = getExternsMap();
+    try {
+      externsMap = getExternsMap();
+    } catch (FileNotFoundException e) {
+      throw new CmdLineException(parser,
+          "externs file " + externsMapFile + " not found.", e);
+    }
   }
 
   Options() {
--- a/src/main/java/com/google/javascript/gents/Options.java
+++ b/src/main/java/com/google/javascript/gents/Options.java
@@ -41,6 +41,11 @@ public class Options {
       handler = StringArrayOptionHandler.class)
   List<String> externs = new ArrayList<>();
 
+  @Option(name = "--externsMap",
+      usage = "File mapping externs to their TypeScript typings equivalent. Formatted as json",
+      metaVar = "EXTERNSMAP")
+  String externsMapFile = null;
+
   @Argument
   List<String> arguments = new ArrayList<>();
 
@@ -87,5 +92,10 @@ public class Options {
   }
 
   Options() {
+    this.externsMapFile = externsMapFile;
+  }
+
+  Options(String externsMapFile) {
+    this.externsMapFile = externsMapFile;
   }
 }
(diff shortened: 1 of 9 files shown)


src/test/java/com/google/javascript/gents/test_externs_map.json
```diff
@@ -0,0 +1,4 @@
+{
+  "jQuery": "JQuery"
```
mprobst: I wonder if it makes sense to hard code this, instead of configuring it? It's probably going to be the same stuff for most projects? OTOH, there might be specifics e.g. for internal Google use. Both works, I guess.

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -80,11 +87,12 @@ public static void main(String[] args) {
   private final Options opts;
   private final Compiler compiler;
   private final GentsErrorManager errorManager;
+  private final Map<String, String> externsMap;
 
   final PathUtil pathUtil;
   final NameUtil nameUtil;
 
-  TypeScriptGenerator(Options opts) {
+  TypeScriptGenerator(Options opts) throws FileNotFoundException {
```
mprobst: Please construct it, passing in the file?

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -94,6 +102,14 @@ public static void main(String[] args) {
 
     this.pathUtil = new PathUtil(opts.root);
     this.nameUtil = new NameUtil(compiler);
+    if (opts.externsMapFile != null) {
+      Type mapType = new TypeToken<Map<String, String>>(){}.getType();
+      JsonReader reader = new JsonReader(new FileReader(opts.externsMapFile));
+      externsMap = new Gson().fromJson(reader, mapType);
+    }
```
mprobst: indent

src/test/java/com/google/javascript/gents/test_externs_map.json
gmoothart: Yeah, I think there are so many we'd need an extension point anyway. I was planning on adding a canonical mapping file to google3 and have convert_js_to_ts use it; but if you prefer  I could add the default file in gents itself.

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
gmoothart: I'm not sure what you're suggesting? I could do the json parsing in the Options class itself (and then it would be the one throwing FileNotFound), is that what you mean?

mprobst: Yes. The idea would be to isolate the TypeScriptGenerator from stuff like file reading and be done with that before it runs. That avoids doing work that can fail in the constructor, and also makes sure that all inputs are validated properly before the program runs.

gmoothart: Done.

src/main/java/com/google/javascript/gents/Options.java
```diff
@@ -75,7 +89,17 @@ private void setIdeMode(final CompilerOptions options) {
...
+    if (this.externsMapFile != null) {
+      Type mapType = new TypeToken<Map<String, String>>(){}.getType();
+      JsonReader reader = new JsonReader(new FileReader(externsMapFile));
+      return new Gson().fromJson(reader, mapType);
+    } else {
+      return ImmutableMap.of();
+    }
+  }
+
+  Options(String[] args) throws CmdLineException, FileNotFoundException {
```
mprobst: Nit/optional: maybe it'd be nicer for code calling into this to convert the `FileNotFoundException` into a `CmdLineException`, so it doesn't have to special case the catch?

gmoothart: much better!

PR #360, opened 2016-09-28
externsMap allows a type alias for 'any' to be supplied

### LINKED ISSUE #328: Convert unknown types to `OldNoType` instead of `any`
In gents, consider aliasing `any` with a type named something like `OldNoType` when typing information cannot be determined.

This will allow human reviewers to be able to find and manually set types in a followup pass, without having to determine whether the type was intended to be `any` in the first place.

gmoothart:
This supports the case where you want to be able to audit your uses of "any" after conversion, by aliasing them to something like `OldNoType`.

Also adds an extension point in TypescriptGeneratorTests to allow certain tests to be run with a different set of options from the other tests.

fixes #328

### DIFF
--- a/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
+++ b/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
@@ -108,7 +108,6 @@ public class GentsCodeGenerator extends CodeGenerator {
           add(anyTypeName);
           return true;
         }
-
         return false;
       default:
         return false;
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -98,6 +101,15 @@ boolean maybeOverrideCodeGen(Node n) {
...
         }
         return false;
+      case ANY_TYPE:
+        // Check the externsMap for an alias to use in place of "any"
+        String anyTypeName = externsMap.get("any");
+        if (anyTypeName != null) {
+          add(anyTypeName);
+          return true;
+        }
+
```
mprobst: nit: drop this line?

PR #364, opened 2016-09-29
Do not promote dotted member assigments to fields

### LINKED ISSUE #357: class field generation incorrectly interprets dotted member assignments as fields
Constructs like `this.w.bar = 5` result in a class field `bar`.

Input:
```
/**
 * @constructor
 */
var A = function() {
  this.w;
  this.w.bar = 'eh';
};
```

output:
```
class A {
  w: any;
  bar: any = 'eh';
}
```

expected output:
```
class A {
  w: any;
  constructor() {
    this.w.bar = 'eh';
  }
}
```

gmoothart:
this.foo.bar = 5 no longer creates a class field bar = 5

fixes #357

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -649,6 +649,9 @@ public final class TypeConversionPass implements CompilerPass {
       Node fullName = getFullName(n);
       // Node MUST start with "this." and be shallow, i.e. "this.foo".
       // "this.foo.bar" and other nestings are not declarations and are ignored.
+      // fullName is a binary tree and multiple parts are represented by GETPROP
+      // nodes recursively on the left (first) child, so a first child of THIS is
+      // sufficient to ensure the name is of the form "this.foo".
       if (!fullName.isGetProp() || !fullName.getFirstChild().isThis()) {
         return null;
       }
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/gents/singleTests/fields.js
```diff
@@ -11,6 +11,9 @@ var A = function(a) {
   let y = 1;
   /** @type {number} */
   this.z = y + 1;
+
+  this.w;
+  this.w.bar = 'bar';
```
mprobst: add a test that just accesses some global variable, `foo.baz = 1;`?

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -647,8 +647,9 @@ static ClassMemberDeclaration newDeclaration(Node n, Map<String, Node> classes)
     @Nullable
     static ClassMemberDeclaration newDeclarationOnThis(Node n) {
       Node fullName = getFullName(n);
-      // Node MUST start with "this."
-      if (!fullName.isGetProp() || !containsThis(fullName)) {
+      // Node MUST start with "this." and be shallow, i.e. "this.foo".
+      // "this.foo.bar" and other nestings are not declarations and are ignored.
+      if (!fullName.isGetProp() || !fullName.getFirstChild().isThis()) {
```
mprobst: Can you formulate the comment such that it's more obvious that if `getFirstChild().isThis()` is true, this can only be a single `this.x` step because `fullName` is a binary tree, i.e. there cannot be a third and fourth child?

gmoothart:
Comments addressed.

PR #365, opened 2016-09-29
Remove duplicate fields

### LINKED ISSUE #358: gents field generation creates duplicate members
if an untyped member is assigned multiple times, it creates multiple properties with the same name and different initialized values.

input:
```
/**
 * @constructor
 */
var A = function() {
  this.n = 5;
};

A.prototype.foo = function() {
  this.n = 12;
};
```

output:
```
class A {
  n: any = 5;
  n: any = 12;
  foo() {}
}
```

expected output:
```
class A {
  n: number = 5;

  foo() {
    this.n = 12;
  }
}
```

gmoothart:
This fixes a bug that would cause gents to generate duplicate fields and promote some assignments to field initializers inappropriately.

It changes behavior to only look in the constructor for fields. With some care to remove duplicates and leave the assignments in-place it would be possible to promote fields in other methods too.

fixes #358

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -1,6 +1,8 @@
 package com.google.javascript.gents;
 
 import com.google.common.base.Preconditions;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.google.javascript.jscomp.AbstractCompiler;
 import com.google.javascript.jscomp.CompilerPass;
 import com.google.javascript.jscomp.JSError;
@@ -13,6 +15,7 @@ import com.google.javascript.rhino.JSDocInfo.Visibility;
 import com.google.javascript.rhino.JSTypeExpression;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


gmoothart:
This branch is based off https://github.com/angular/clutz/pull/364; only review the second commit for this PR (is there a better way to do this in github?

src/test/java/com/google/javascript/gents/singleTests/fields.js
```diff
@@ -11,6 +11,11 @@ var A = function(a) {
   let y = 1;
   /** @type {number} */
   this.z = y + 1;
+
+  this.w;
+  this.w.bar = 'bar';
+
+  this.n = 13;
```
mprobst: can you add a test where the constructor mentions `this.n` twice, e.g. assigns `12` into it on the next line?

gmoothart: :cry: You are right, we need to de-duplicate anyway. I'd like to do that in a separate PR, though.

gmoothart: Actually, never mind. To do this right I need a different approach so I'm going to just do that directly.

gmoothart:
@mprobst PTAL; new approach.

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -133,44 +136,61 @@ public void visit(NodeTraversal t, Node n, Node parent) {
    * Converts fields declared internally inside a class using the "this" keyword.
    */
   private class FieldOnThisConverter extends AbstractPostOrderCallback {
+    // Map from class node to its field names
```
mprobst: `/** JSDoc */`

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -133,44 +136,61 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-          Node fnNode = NodeUtil.getEnclosingFunction(n);
-          String fnName = fnNode.getParent().getString();
+        Node fnNode = NodeUtil.getEnclosingFunction(n);
+        String fnName = fnNode.getParent().getString();
+
+        // TODO(gmoothart): in many cases we should be able to infer the type from the rhs if there
+        // is no jsDoc
+
+        // Handle parameter properties when we are in the constructor and have a
+        // declaration of the form this.name = name;
```
mprobst: add how you handle them? I.e. "by converting them to private/public constructor parameters" or something like that?

PR #367, opened 2016-09-30
fix: support transitive types shadowed by namespaces.

mprobst:
Previously, Clutz would ignore types that are contained in a namespace,
assuming they were just fields on that namespace. Now, Clutz will still
generate them if they are their own provide (from the list of transitive
provides).

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -51,9 +51,6 @@ import com.google.javascript.rhino.jstype.TemplateType;
 import com.google.javascript.rhino.jstype.TemplatizedType;
 import com.google.javascript.rhino.jstype.UnionType;
 import com.google.javascript.rhino.jstype.Visitor;
-
-import org.kohsuke.args4j.CmdLineException;
-
 import java.io.File;
 import java.io.IOException;
 import java.io.StringWriter;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -486,18 +483,24 @@ private void processUnprovidedTypes(Set<String> provides) {
       // AFAICT, there is no api for going from type to symbol, so iterate all symbols first.
       for (TypedVar symbol : compiler.getTopScope().getAllSymbols()) {
         String name = symbol.getName();
-        // skip unused symbols, symbols already emitted or symbols whose namespace is emitted.
+        String namespace = getNamespace(name);
+        // skip unused symbols, symbols already emitted or symbols whose namespace is emitted
+        // (unless the symbols have their own provide).
         if (!typesUsed.contains(name) || typesEmitted.contains(name) ||
-            typesEmitted.contains(getNamespace(name))) {
+            (!transitiveProvides.contains(name) && typesEmitted.contains(namespace))) {
```
rkirov: I have hunch that this will need even more iteration in the future, but I will leave it for a refactoring after we know clutz with strict deps works. The core issue is that at two different places we do the '(TypedVar symbol : compiler.getTopScope().getAllSymbols())' and decide whether to emit symbol or not, and the logic is not shared.

PR #368, opened 2016-09-30
feat: add empty lines between methods in classes.

dpurp:
Summary:
Adds an empty line between functions in a class.

### DIFF
--- a/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
+++ b/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
@@ -69,7 +69,7 @@ public class GentsCodeGenerator extends CodeGenerator {
       //   - the next node is further away than the next line
       //   - the next node is a class function (but the current node is not a comment)
       if (lineSpacing > 1 ||
-          (nextNode.getToken() == Token.MEMBER_FUNCTION_DEF ^ n.getToken() == Token.EMPTY)) {
+          (nextNode.getToken() == Token.MEMBER_FUNCTION_DEF && n.getToken() != Token.EMPTY)) {
         add("\n");
       }
     }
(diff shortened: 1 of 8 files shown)


src/test/java/com/google/javascript/gents/singleTests/access_modifiers.ts
```diff
@@ -10,10 +10,15 @@ class A {
   private e: number;
   // Method access
```
gmoothart: I'm curious if there is an easy way to add a space between fields and the constructor as well - what if you added before rather than after the method body? No need to hold up this PR if it isn't straightforward.

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -54,11 +56,20 @@ protected void add(Node n, Context ctx) {
...
     Node nextNode = n.getNext();
     if (nextNode != null) {
       int lineSpacing = nextNode.getLineno() - n.getLineno();
-      if (lineSpacing > 1) {
+
+      // Add an empty line if:
+      //   - the next node is further away than the next line
+      //   - the next node is a class function (but the current node is not a comment)
+      if (lineSpacing > 1 ||
+          (nextNode.getToken() == Token.MEMBER_FUNCTION_DEF ^ n.getToken() == Token.EMPTY)) {
```
gmoothart: One of the newline conditions here is "current node is empty and next token is not a member function" - i.e. add an newline after all empty tokens except those followed by a member function. That seems non-intuitive to me; why is it what we want?

dpurp: Otherwise, the output will have an extra new line after the `//inline comment`

See below:

Input:
```javascript
const F = goog.defineClass(null, {
  // inline comment

  /**
   * block comment
   */
  constructor: function() {
  },
});
```

Output without check:
```typescript
class F {
  // inline comment


  /**
   * block comment
   */
  constructor() {}
}
```

src/test/java/com/google/javascript/gents/singleTests/access_modifiers.ts
dpurp: Probably, but I'd rather do it in another PR.

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
dpurp: Not sure what I was thinking here, but that does indeed work.  (Updated accordingly)

Issue #329, opened 2016-08-23
Eliminate extraneous "export {name}" statements


obsc:
```js
goog.module("A.B");
var B = function() { };
exports = B;
```
is converted to
```js
let B = function() {};
export {B};
```
This should be simplified to
```js
export function B() {}
```

dpurp:
I'll take this one up.

gmoothart:
#370 added default exports, which partially addresses this.

PR #369, opened 2016-10-02
feat(#329): Add support for `export default` in gents.

dpurp:
This change adds support for 'default exports' to gents  #329 .

Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export


dpurp:
Closing this, so that I can squash the commits into a single sha

mprobst:
@dpurp This works, but you can also squash the commits and then force push to your review branch (just make sure not to force push to a branch in the shared repository).

PR #370, opened 2016-10-02
feat: Add support for `export default` in gents.

dpurp:
This change adds support for 'default exports' to gents.  See request
#329 .

Reference:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statem
ents/export


dpurp:
Tests fixed + rebased.

PR #377, opened 2016-10-05
fix: Support defaultExports in import rewriting

dpurp:
The import rewriter is not aware of default exports, which results in all exports being imported as a named export.

Supports #329

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
@@ -1,5 +1,6 @@
 package com.google.javascript.gents;
 
+import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.HashBasedTable;
 import com.google.common.collect.Table;
@@ -15,6 +16,7 @@ import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
  * Converts Closure-style modules into TypeScript (ES6) modules (not namespaces).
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -264,29 +294,32 @@ void convertRequireToImportStatements(Node n, String fullLocalName, String requi
 
     if (module.providesObjectChildren.get(requiredNamespace).size() > 0) {
```
gmoothart: should this be an else if? It doesn't look like it makes sense to hit both code paths.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -203,12 +205,16 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
 
   /**
    * Converts a Closure goog.require call into a TypeScript import statement.
-   *
+   * <p>
```
dpurp: It's been quite some time; this was added by google-java-format.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -242,17 +250,39 @@ void convertRequireToImportStatements(Node n, String fullLocalName, String requi
...
-      // import {value as localName} from "./file"
-      Node importSpec = new Node(Token.IMPORT_SPEC, IR.name(moduleSuffix));
-      // import {a as b} only when a =/= b
-      if (!moduleSuffix.equals(localName)) {
-        importSpec.addChildToBack(IR.name(localName));
+      @Nullable String defaultExportSymbol = module.exportedNamespacesToSymbols.get(EXPORTS);
+      String importedSymbol = module.importedNamespacesToSymbols.get(requiredNamespace);
+      boolean isDefaultExport = importedSymbol.equals(defaultExportSymbol);
+
+      Optional<Node> defaultExport;
```
dpurp: I actually see that as an anti-pattern, since static analyzers are unable to determine if it was set in each branch.

src/test/java/com/google/javascript/gents/multiTests/type_rewrite/import.ts
```diff
@@ -1,9 +1,9 @@
 import E from 'goog:keep.E';
 
 import * as A from './imported_module';
-import {A as X} from './imported_module';
+import X from './imported_module';
```
dpurp: We don't support the combined syntax, but I suspect that this still works

(e.g. `import defaultExport, {namedExports} from './file'`)

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: It'll break the existing test cases; though, those cases may not actually be valid.

In any case, since it changes existing functionality/support, it should be in its own PR.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
mprobst: By Google standards, it should be on the same line as the text, right?

```
/**
 * Foo bar baz.
 *
 * <p>Bam blergh.
 */
```

... not that I care.

dpurp:
I agree with that philosophy for new TypeScript files, but lack of support for default exports makes it difficult to convert components one at a time, since (at least in our codebase) we tend to use default exports everywhere in Closure (`exports = Foo`).

Without support for default exports, we'll have to update the import/goog.require line for every client depending on our component, since they are outside of gents' scope.

dpurp:
Withdrawing per discussions.

PR #378, opened 2016-10-05
fix: Don't rename default exported values.

dpurp:
Summary:
Currently, if the name of a value does not match the module's suffix,
gents assigns the default export to a new variable that does. Remove
this behavior since it is unnecessary.

This supports #329 . 

Assume the source is from `goog.module('package.A');`

The output would be:

Before:

```typescript

class Foo {}

export default const A = foo;
```

After:

```typescript

class Foo{}

export default Foo;
```


src/test/java/com/google/javascript/gents/multiTests/module_imports/export_default.ts
```diff
@@ -1,3 +1,3 @@
-export default class A {
```
mprobst: This is quite useful - it'll make sure that JavaScript VMs assign a name to the function, visible in stack traces etc. Same for the case above.

Is this causing a problem?

dpurp: It wasn't immediately obvious how to solve/fix.  

The jscomp code generator seems to decide whether to keep it or remove it based on if it's used in the file (there should be a test in here to verify this behavior)

Is this something that you would like to see fixed in this PR or as a future feature request?

dpurp: Actually, it turns out this was easy to do. (Done)

PR #383, opened 2016-10-11
feat: Rewrite the AST to directly `export` nodes when possible.

### LINKED ISSUE #329: Eliminate extraneous "export {name}" statements

dpurp:
Fixes #329


src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -118,8 +133,8 @@ public void visit(NodeTraversal t, Node n, Node parent) {
                 new Node(Token.EXPORT_SPEC, Node.newString(Token.NAME, localName))));
             parent.addChildAfter(export, n);
             // Registers symbol for rewriting local uses.
-            registerLocalSymbol(child.getSourceFileName(), exportedNamespace, exportedNamespace,
-                localName);
+            registerLocalSymbol(
```
mprobst: Maybe for the future, it'd be nicer if you sent one "reformat everything" CL, instead of reformatting files as you touch them. This makes it quite hard to review the code, and will cause further confusion in blame lines (why was this line changed as part of a commit that did X?).

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -134,16 +149,43 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+    private void directlyExportSymbol(Node namedNode, Node parent, String filename) {
+      if (!fileToModule.containsKey(filename)) {
+        return;
+      }
+      FileModule module = fileToModule.get(filename);
+      Node child = namedNode.getFirstChild();
+      String nodeName = checkNotNull(child.getQualifiedName());
+
+      ExportedSymbol exportedSymbol = ExportedSymbol.of(filename, nodeName, nodeName);
+      if (module.exportedNamespacesToSymbols.containsValue(nodeName)) {
```
mprobst: this is the `goog.provide` case, right? add a comment?

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -31,10 +38,13 @@
 
   private final Map<String, FileModule> fileToModule;
   private final Map<String, FileModule> namespaceToModule;
+  private final Set<ExportedSymbol> directlyExportedSymbols = new HashSet<>();
+  private final Map<ExportedSymbol, Node> exportsToNodes = new HashMap<>();
```
mprobst: to what kind of nodes? To the actual `export ...`, or to the node that defines the symbol? Or are those identical here? Please document.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -134,16 +149,43 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+        Node next = namedNode.getNext();
+        namedNode.detachFromParent();
+
+        Node export = new Node(Token.EXPR_RESULT, new Node(Token.EXPORT, namedNode));
+        nodeComments.moveComment(namedNode, export);
+        parent.addChildBefore(export, next);
+        directlyExportedSymbols.add(exportedSymbol);
+        compiler.reportCodeChange();
+      } else {
+        exportsToNodes.put(exportedSymbol, namedNode);
```
mprobst: add a comment what condition this represents? I think I'd also pull this up by inverting the if/else cases, and then doing an early exit from the function to reduce indentation and make it easier to understand ("first handle special case X and return, then do the rest...").

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -306,24 +348,56 @@ void convertRequireToImportStatements(Node n, String fullLocalName, String requi
    * For example,
    * convertExportAssignment(pre.fix = ..., "pre.fix", "name") <-> export const name = ...
    * convertExportAssignment(pre.fix.foo = ..., "pre.fix", "name") <-> name.foo = ...
+   * @param filename the name of the containing file
```
mprobst: please avoid JavaDoc that's redundant with the parameter name.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -134,16 +149,43 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+                exportedNamespace,
                 symbols.get(exportedNamespace));
           }
           break;
         default:
           break;
       }
     }
+
+    private void directlyExportSymbol(Node namedNode, Node parent, String filename) {
```
mprobst: please camel case as `fileName`, to be consistent with JSCompiler's API (here and everywhere).

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -347,4 +421,85 @@ void registerLocalSymbol(String sourceFile, String fullLocalName, String require
     typeRewrite.put(sourceFile, fullLocalName, localName);
     typeRewrite.put(sourceFile, requiredNamespace, localName);
   }
+
+  /** Metadata about an exported symbol. */
+  private static class ExportedSymbol {
```
mprobst: Nice!

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -347,4 +421,85 @@ void registerLocalSymbol(String sourceFile, String fullLocalName, String require
...
+      String localName = (rhs.getQualifiedName() != null) ? rhs.getQualifiedName() : exportedSymbol;
+
+      String exportedName;
+      if (exportedNamespace.equals(EXPORTS)) { // is a default export
+        exportedName = localName;
+      } else if (exportedNamespace.startsWith(EXPORTS)) { // is a named export
+        exportedName =
+            exportedNamespace.substring(EXPORTS.length() + 1, exportedNamespace.length());
+      } else { // exported via goog.provide
+        exportedName = "";
```
mprobst: I think I'd actually prefer to use `null` for this. Empty strings mostly behave like strings, so this might cause more confusion down the line.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -347,4 +421,85 @@ void registerLocalSymbol(String sourceFile, String fullLocalName, String require
...
+
+    /**
+     * The name that the symbol is exported under via named exports.
+     *
+     * <p>For example, {@code Bar} in: <code> exports {Bar}</code>.
+     *
+     * <p>If a symbol is directly exported, as in the case of <code>export class Foo {}</code>, the
+     * {@link #localName} and {@link #exportedName} will both be {@code Foo}.
+     */
+    final String exportedName;
```
mprobst: Document the `goog.provide` case?

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: These should only be lines that I touched, rather than the whole file.  I probably made a change then reverted it, but the changes were likely in different git sub-commits so it didn't quite trigger right.

That being said, I can send a "reformat cl," if you'd like the code base as a whole to match google-java-format.

mprobst:
>
> These should only be lines that I touched, rather than the whole file. I
> probably made a change then reverted it, but the changes were likely in
> different git sub-commits so it didn't quite trigger right.
>
Ah OK, maybe I didn't spot the actual diff in here.

> That being said, I can send a "reformat cl," if you'd like the code base
> as a whole to match google-java-format.
>
Yes, I think if people on this are going to use google-java-format
generally (and I think that's true), then we should do that.
-- 

Martin

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: It was actually `export const A = B;` but I looked into this more with a fresh head and was able to refactor the code, so we don't need to do the AST transformation here, and, instead, can consolidate it in one place.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: refactored away :D

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: ack

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: goog.provide now uses `exportedName`

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
dpurp: refactored

PR #384, opened 2016-10-12
feat: Support goog.scope transformations in gents.

### LINKED ISSUE #349: Support Closure files with `goog.scope`
I converted some .js files into TypeScript. I see that `goog.scope` is not removed in the generated file.

dpurp:
Summary:
Rewrite the AST, removing the `goog.scope` wrapper, so that the various module and class-based transformations can take place.

Partially Adresses #349

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -128,7 +128,7 @@ public final class CommentLinkingPass implements CompilerPass {
       StringBuilder sb = new StringBuilder();
       String sep = "";
       for (Comment c : group) {
-        String comment = filterTags(c.type, c.value);
+        String comment = filterCommentContent(c.type, c.value);
         if (!comment.isEmpty()) {
           sb.append(sep).append(comment);
           sep = "\n";
@@ -142,8 +142,8 @@ public final class CommentLinkingPass implements CompilerPass {
       forceAddCommentToGroup();
     }
 
-    /** Removes unneeded tags from the comment */
-    private String filterTags(Type type, String comment) {
+    /** Removes unneeded tags and markers from the comment */
+    private String filterCommentContent(Type type, String comment) {
       Pattern[] replacements = (type == Type.JSDOC) ? JSDOC_REPLACEMENTS : COMMENT_REPLACEMENTS;
       for (Pattern p : replacements) {
         comment = p.matcher(comment).replaceAll("${keep}");
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -23,7 +23,7 @@ public final class CommentLinkingPass implements CompilerPass {
    * These jsdocs delete everything except for the `keep` capture group
    * Some regexes contain an empty capture group for uniform handling.
    */
-  private static final Pattern[] COMMENT_REPLACEMENTS = {
+  private static final Pattern[] JSDOC_REPLACEMENTS = {
       Pattern.compile("@(constructor|interface|record)[ \t]*(?<keep>)"),
       Pattern.compile("@(extends|implements|type)[ \t]*(\\{.*\\})[ \t]*(?<keep>)"),
       Pattern.compile("@(private|protected|public|package|const)[ \t]*(\\{.*\\})?[ \t]*(?<keep>)"),
@@ -35,6 +35,9 @@ public final class CommentLinkingPass implements CompilerPass {
       Pattern.compile("(?m)^[ \t]*\\*\\s*\n(?<keep>)"),
       // Remove type annotation from @export
       Pattern.compile("(?<keep>@export)[ \t]*(\\{.*\\})")};
+
+  private static final Pattern[] COMMENT_REPLACEMENTS = {
+      Pattern.compile("//\\s*goog.scope\\s*(?<keep>)")};
 
   private final Compiler compiler;
   private final NodeComments nodeComments;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -140,8 +143,9 @@ private void linkCommentToNode(Node n) {
     }
 
     /** Removes unneeded tags from the comment */
-    private String filterTags(String comment) {
-      for (Pattern p : COMMENT_REPLACEMENTS) {
+    private String filterTags(Type type, String comment) {
```
mprobst: rename the method & change its comment?

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -0,0 +1,69 @@
...
+    // Extract the goog.scope contents, and add them to module being constructed.
+    Node blockOfScopeContents = n.getLastChild().getLastChild().getLastChild();
+    blockOfScopeContents.detachFromParent();
+
+    // Collect all of the children into a list before detaching them from the parent,
+    // since detaching breaks the "getNext()" chain.
+    List<Node> scopeChildren = new ArrayList<>();
+    @Nullable Node maybeChildToAdd = blockOfScopeContents.getFirstChild();
+    while (maybeChildToAdd != null) {
+      scopeChildren.add(maybeChildToAdd);
```
mprobst: optional: I think you can collapse these two loops, by always taking the first child here (both in the while condition and the body), and then removing that child and appending it at the insertion point.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    // Now that we have the contents of the goog.scope, rewrite the AST.
+
+    // Create a marker so that we know where to insert the goog.scope contents.
+    Node insertAfterThisNode = n;
+
+    for (Node scopeChild : scopeChildren) {
+      scopeChild.detachFromParent();
+      n.getParent().addChildAfter(scopeChild, insertAfterThisNode);
```
mprobst: Are comments going to survive this? See NodeComments. I think yes, but might be worth adding a test.

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -146,6 +146,9 @@ void generateTypeScript() {
     Node externRoot = compiler.getRoot().getFirstChild();
     Node srcRoot = compiler.getRoot().getLastChild();
 
+    RemoveGoogScopePass amodulePrePass = new RemoveGoogScopePass(compiler);
```
mprobst: amodule?

src/test/java/com/google/javascript/gents/singleTests/goog_scope.js
```diff
@@ -0,0 +1,27 @@
+goog.provide("A.B");
+goog.provide("A.B.C");
+
+goog.scope(function() {
+
```
mprobst: a common pattern is to alias "require" results here (`var LocalName = goog.something.LocalName;`). Maybe add a test case?

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.javascript.jscomp.NodeTraversal;
+import com.google.javascript.rhino.Node;
+import com.google.javascript.rhino.Token;
+import java.util.ArrayList;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * Compiler pass removing the {@code goog.scope} wrapper, so that the module and class
+ * transformations can take effect.
```
mprobst: Add a comment that this translation is legal, because we're turning all files into modules so `var`s in the `goog.scope` body remain local (not global)?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: Done.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
dpurp: Good point.

Unfortunately, this breaks the "aggressive" exporting.  I'd like to leave this as a TODO for a future cl, if possible.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
dpurp: Done. Thanks!

src/test/java/com/google/javascript/gents/singleTests/goog_scope.js
dpurp: Good point.

Unfortunately, this breaks the "aggressive" exporting.  I'd like to leave this as a TODO for a future cl, if possible.

src/test/java/com/google/javascript/gents/singleTests/goog_scope.ts
```diff
@@ -2,14 +2,14 @@ export class Foo {
...
-    return 4;
+  static foo(): string {
+    return 'this should create a static method on Foo, since it is NOT goog.provided';
   }
 
   // TODO: Aggressively export rather than create static methods/fields
-  static qux(): number {
-    return 4;
+  static qux(): string {
+    return 'this should be directly exported, rather than create a static method, ' +
```
mprobst: It isn't though, is it? I don't mind that (I think exporting a static function of a class is a rare corner case), but maybe make it clearer language wise that this is not intended behaviour?

dpurp: the TODO above was supposed to indicate this :)

PR #388, opened 2016-10-12
Files with the same basename don't clobber each other.

### LINKED ISSUE #348: Files with same basename clobber each other
a/service.js
b/service.js

will convert to

a/service.ts
b/service.ts

Except that the output from converting one of the files will be the same between both `.ts` outputs, losing information from the other `.js` file.

gmoothart:
Use the whole file path as a hash key, so that files in different
directories that have the same basename do not overwrite one another.

fixes #348


gmoothart:
I was hoping to skate by without the test :-)

It took some care to get the recursive files working in MultiTests

PR #389, opened 2016-10-13
chore: Fix errorprone and other warnings.

dpurp:


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -613,15 +612,25 @@ private void sortSymbols(List<TypedVar> symbols) {
   }
 
   private boolean needsAlias(Set<String> shadowedSymbols, String provide, TypedVar symbol) {
-    if (!shadowedSymbols.contains(provide)) return false;
+    if (!shadowedSymbols.contains(provide)) {
```
mprobst: FYI we actually diverged from the style guide kind of intentionally here. I guess the convenience is not worth the divergence.

Issue #376, opened 2016-10-04
gents outputs mis-alligned jsdoc comments


gmoothart:
example:
https://github.com/angular/clutz/blob/a39822fd3d74c8254ca0e943b5555746b7ac7627/src/test/java/com/google/javascript/gents/singleTests/static_methods.ts#L9

Seems like it would be best for clang to handle this.

dpurp:
I suspect this is from an over-eager regex trimming in `CommentLinkingPass`

dpurp:
Fixed with update to clang-format in https://github.com/angular/clutz/commit/7db988fa59a068cf2eb6da121aa461dec99b4b5e

PR #391, opened 2016-10-14
feat: Add better newline insertion.

dpurp:
Summary:
Changes the newline insertion heuristic to always add a newline before comments and specific tokens.

This removes the previous heuristic of using line numbering for inserting newlines when nodes differed by more than one line number, which broke with multi-line statements.

Supports #376

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -126,7 +126,7 @@ public final class CommentLinkingPass implements CompilerPass {
     /** Links a comment to the corresponding node */
     private void linkCommentToNode(Node n) {
       StringBuilder sb = new StringBuilder();
-      String sep = "";
+      String sep = "\n";
       for (Comment c : group) {
         String comment = filterCommentContent(c.type, c.value);
         if (!comment.isEmpty()) {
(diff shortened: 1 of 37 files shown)


src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -126,7 +126,7 @@ private void addCommentToGroup(Node n) {
     /** Links a comment to the corresponding node */
     private void linkCommentToNode(Node n) {
       StringBuilder sb = new StringBuilder();
-      String sep = "";
+      String sep = "\n";
```
mprobst: Can you add a test where the Closure code contains:

```
// Hello
// World
/* Hello */
/* World */
```

That is, two subsequent line/C style comments?

dpurp: See comments.ts

mprobst: Ah I see, the code review doesn't contain `comments.js` and only shows the changed parts there. Looks good.

src/test/java/com/google/javascript/gents/singleTests/classes.ts
```diff
@@ -50,32 +51,32 @@ class E extends C {
...
 
 class F {
   // inline comment
 
+
   /**
-   * block comment
-   */
+     * block comment
+     */
```
gmoothart: why is this worse?

dpurp: i suspect the unattached, inline comment is throwing off the indenter.

PR #394, opened 2016-10-18
feat: Rewrite aliases in goog.scope.

### LINKED ISSUE #349: Support Closure files with `goog.scope`

dpurp:
Replaces aliased names within goog.scope so that further class based transformations can take place.

Fixes #349

### DIFF
--- a/src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
+++ b/src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
@@ -113,7 +113,14 @@ public final class RemoveGoogScopePass extends AbstractTopLevelCallback implemen
   }
 
   private void maybeReasignAlias(Node assign) {
-    String alias = assign.getFirstFirstChild().getQualifiedName();
+    // Find the name of the deepest first child.
+    String alias = null;
+    for (Node child = assign.getFirstChild(); child != null; child = child.getFirstChild()) {
+      if (child.isName()) {
+        alias = child.getString();
+      }
+    }
+
     if (aliasToProvidedNamespace.containsKey(alias)) {
       String providedNamespace = aliasToProvidedNamespace.get(alias);
       String suffix = assign.getFirstChild().getQualifiedName().substring(alias.length());
--- a/src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
+++ b/src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
@@ -3,8 +3,13 @@ package com.google.javascript.gents;
 import com.google.javascript.jscomp.AbstractCompiler;
 import com.google.javascript.jscomp.CompilerPass;
 import com.google.javascript.jscomp.NodeTraversal;
+import com.google.javascript.jscomp.NodeUtil;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/javascript/gents/singleTests/goog_scope.js
```diff
@@ -19,14 +19,12 @@ const Foo = lorem.ipsum.Foo;
 
 /** @return {string} */
 Foo.foo = function() { 
-  return "this should create a static method on Foo, since it is NOT goog.provided"; 
+	return 'this is a static method on Foo, since it is NOT goog.provided';
```
gmoothart: fix indenting?

src/test/java/com/google/javascript/gents/singleTests/goog_scope.ts
```diff
@@ -1,22 +1,20 @@
...
+export function qux(): string {
+  return 'this is directly exported since it is goog.provided';
 }
-const Foo = Foo;
-Foo.Bar = class {
-  static bar(): boolean {
-    return false;
-  }
+
+export const num: number = 8;
```
gmoothart: Foo.num is not goog.Provide'd, so shouldn't it be static instead of exported?

src/test/java/com/google/javascript/gents/singleTests/goog_scope.js
dpurp: Done.

src/test/java/com/google/javascript/gents/singleTests/goog_scope.ts
dpurp: I don't quite understand the behavior here to be honest.  I found the code that does the "aggressive exporting," and disabling it provides a more consistent experience, but breaks a bunch of the existing tests (TODO added).

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -113,7 +113,14 @@ private Node maybeRecordAndRemoveAlias(Node assign) {
   }
 
   private void maybeReasignAlias(Node assign) {
-    String alias = assign.getFirstFirstChild().getQualifiedName();
+    // Find the name of the deepest first child.
+    String alias = null;
+    for (Node child = assign.getFirstChild(); child != null; child = child.getFirstChild()) {
+      if (child.isName()) {
+        alias = child.getString();
```
dpurp: `getQualifiedName()` calls `getString()`, so they are equivalent, but switching for consistency.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -113,7 +113,14 @@ private Node maybeRecordAndRemoveAlias(Node assign) {
   }
 
   private void maybeReasignAlias(Node assign) {
-    String alias = assign.getFirstFirstChild().getQualifiedName();
+    // Find the name of the deepest first child.
+    String alias = null;
+    for (Node child = assign.getFirstChild(); child != null; child = child.getFirstChild()) {
```
dpurp: `Foo.Bar` wasn't being directly exported

PR #398, opened 2016-10-20
fix: goog.provide only export content on the goog.provide line.

dpurp:
Summary:
The current "aggresive exporting" method for goog.provide leads to counter-intuitive exporting.  Change the implementation to only export namespaces directly specified in the `goog.provide`.


src/test/java/com/google/javascript/gents/multiTests/import_rewrite/export.ts
```diff
@@ -1,23 +1,17 @@
 export function B() {}
-
-export const x = 4;
-
-export const y = 8;
+obj.C.x = 4;
```
mprobst: this won't work as it doesn't define `obj.C`, does it? Or am I missing something?

src/test/java/com/google/javascript/gents/multiTests/provide_imports/export.ts
```diff
@@ -1,17 +1,15 @@
 export function B() {}
 
-export const x = 4;
-
-export const y = 8;
+export const C = {};
```
mprobst: this won't work because the object `C` in TypeScript never defines that it has a property `x`. We'd have to generate:

```
C = {
  x: ...,
};
```

dpurp: fixed

src/test/java/com/google/javascript/gents/multiTests/import_rewrite/export.ts
```diff
@@ -1,23 +1,20 @@
...
 
 export function D() {}
-
-export function foo() {}
+D.foo = function() {};
 
 export function E() {}
-
-export function bar() {}
+E.bar = function() {};
```
mprobst: is this legal TypeScript? E doesn't declare it has a `bar` property, does it?

dpurp: And I almost got away with it, too.  Darn you and your thorough review. :P

mprobst: Haha. I'll file an issue to set up a test that compiles the `.ts` code, we should find these in a structured way.

Issue #382, opened 2016-10-10
Add readonly modifier to @const class fields


gmoothart:
When creating fields for typescript classes, any members that are annotated `@const` should have the `readonly` TS modifier added.

input:

``` javascript
/**
 * @constructor
 * @param {number} a
 */
var A = function(a) {
  /** @const {number} */
  this.a = a;
  /** @const {boolean} */
  this.b = true;
};
```

output:

``` typescript
class A {
  b: boolean = true;
  constructor(public a: number) {
  }
}
```

desired output:

``` typescript
class A {
  readonly b: boolean = true; // This work is still outstanding.

  // Adding `readonly` here has been fixed.
  constructor(public readonly a: number) {
  }
}
```

dpurp:
I think we should be able to do add this in GentsCodeGenerator#maybeOverideCodeGen.

If I understand the AST correctly, we would need to check for a MEMBER_PROPERTY that points to CONST node in the switch.

vpanta:
I can't tell if this is supposed to be fixed (sorry, github newbie).

But we're running into the same issue, private variables marked /** @private @const */ are not getting the readonly attribute added to them by Gents. Is that expected?

rkirov:
It's only fixed for ctor arguments, not fields (see discussion on PR).

rkirov:
Fixed with https://github.com/angular/clutz/commit/d3bac1d94ef32c17134357f939d25eed2667af56

PR #399, opened 2016-10-20
Add readonly to const parameters properties in gents

gmoothart:
This updates constructor parameter properties to be marked `readonly` if the parameter is annotated with `@const` 

Partially addresses #382

### DIFF
--- a/src/test/java/com/google/javascript/gents/singleTests/parameter_prop.ts
+++ b/src/test/java/com/google/javascript/gents/singleTests/parameter_prop.ts
@@ -10,5 +10,5 @@ class A {
 class B {
   constructor(
       public a: number, public b: number, protected c: number,
-      private d: number) {}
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/gents/singleTests/parameter_prop.ts
```diff
@@ -10,5 +10,5 @@ class A {
 class B {
   constructor(
       public a: number, public b: number, protected c: number,
-      private d: number) {}
+      private d: number, public readonly e: number) {}
```
mprobst: can you add a test where we generate a regular field?

gmoothart: Regular fields don't work; this is just for the parameter properties case.

PR #400, opened 2016-10-20
fix: Only add empty export blocks when necessary

dpurp:
Summary:
The current implementation adds empty export blocks to the file, if the file does not have an exports; however, empty export blocks are only needed to coerce a file to a module if has neither imports nor exports. 

To fix this, we augment the `FileModule` to record whether the module has any `goog.require`s, and then only add the empty block to the files without either.   When we do add the empty export block, also add a comment explaining its purpose.

### DIFF
--- a/src/test/java/com/google/javascript/gents/multiTests/module_imports/export_sideeffect.ts
+++ b/src/test/java/com/google/javascript/gents/multiTests/module_imports/export_sideeffect.ts
@@ -1,2 +1,5 @@
-export {};
+export {
+  //  Empty export added to coerce the file to be a module, since it does not
+  //  have any imports or exports.
+};
 console.log('hello');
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/javascript/gents/multiTests/module_imports/export_sideeffect.ts
```diff
@@ -1,2 +1,5 @@
-export {};
+export {
+  //  Empty export added to coerce the file to be a module, since it does not
```
mprobst: uber-nit: maybe make it a line comment, and make it a bit terser?

`export {/* gents: force this file to be an ES6 module (no imports or exports) */};'`

gmoothart: or a line-comment above:

```
// gents: force this file to be an ES6 module (no imports or exports) 
export {};
```

dpurp: Unfortunately, a newline is added after the comment (either by clang or gents), so this can't be on one line; explicitly making it multi-line, which is prettier than:
```javascript
export {/* gents: force this file to be an ES6 module (no imports or exports) */
};
```

@gmoothart: Unfortunately the comment can't be added before the node; otherwise, the comment gents separated from the node by clang-format, which thinks it's a file overview comment.

PR #405, opened 2016-10-24
feat: Add a newline between class properties and constructor

dpurp:


src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -123,6 +124,14 @@ boolean maybeOverrideCodeGen(Node n) {
           return true;
         }
         return false;
+      case MEMBER_FUNCTION_DEF:
+        // Add special newline insertion handling for constructors, since adding a newline after
+        // a node doesn't work between properties and the constructor due to semi-colon insertion.
+        if ("constructor".equals(n.getString())
```
gmoothart: Do you want to check for fields too, so you don't add a newline if the class doesn't have any fields?

PR #407, opened 2016-10-25
fix: Add more comprehensive tests for goog.scope.

dpurp:
Summary:
 - Switch to a multi-file approach to test imports.
 - Fix NPE when trying to renamed a bracket-accessed alias (Token: GET_ELEM).
 - Add support for de-aliasing goog.required values.
 - Add additional tests for imports and aliases.


src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -46,8 +48,9 @@ public void visit(NodeTraversal t, Node n, Node parent) {
     }
 
     switch (callName) {
-      case "goog.provide":
-        // Register the goog.provided namespaces, so that we can remove any aliases.
+      case "goog.provide":  // FALL-THROUGH
```
mprobst: nit: we only do FALL-THROUGH comments on case branches that contain code, multiple cases in a row are just regular code.

dpurp: Fixed.

PR #409, opened 2016-10-25
Fix ClassCastException.

mckamey:
Encountered an UnionType being cast as an ObjectType.


mckamey:
Just signed the CLA. Thanks!

PR #410, opened 2016-10-25
fix: Don't drop @fileoverview comments

### LINKED ISSUE #403: gents drops @fileoverview comments
@fileoverview jsdoc comments are dropped in gents output. I suspect it is because the comment is being tied to the top goog.module and then dropped when that node is dropped.

This may or may not be an issue when using goog.provide, we should investigate that.

input:
```
/** @fileoverview This is a file */
goog.module('A.B');
```

output:
```
<nothing>
```

desired output:
```
/** @fileoverview This is a file */
```

dpurp:
Fixes #403

### DIFF
--- a/src/test/java/com/google/javascript/gents/singleTests/module_class2.js
+++ b/src/test/java/com/google/javascript/gents/singleTests/module_class2.js
@@ -1,3 +1,6 @@
+/**
+ * @fileoverview Rivoting description of the file.
+ */
 goog.module("A.B.Klass");
 
 /** Possibly outdated information about Klass. */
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/javascript/gents/singleTests/module_class2.js
```diff
@@ -1,3 +1,6 @@
+/**
```
gmoothart: can you add a fileoverview to a file with goog.provide?

dpurp: done

PR #414, opened 2016-10-27
fix: Don't add empty exports when assigning goog.require to a var.

dpurp:
Summary:
The current ‘no empty export’ detection mechanism doesn’t take into account cases where the goog.require is assigned to a value. (e.g. `const A = goog.require(‘A’);`)

### DIFF
--- a/src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
+++ b/src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
@@ -62,12 +62,24 @@ public final class CollectModuleMetadata extends AbstractTopLevelCallback implem
 
   @Override
   public void visit(NodeTraversal t, Node n, Node parent) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
```diff
@@ -62,12 +62,24 @@ public void process(Node externs, Node root) {
 
   @Override
   public void visit(NodeTraversal t, Node n, Node parent) {
+    String filename = n.getSourceFileName();
+    @Nullable FileModule module = fileToModule.get(filename);
+
+    // const A = goog.require('path.to.A');
+    if (module != null && (n.isConst() || n.isLet() || n.isVar())) {
```
gmoothart: Maybe make a helper method for `n.isConst() || n.isLet() || n.isVar()` ?

PR #415, opened 2016-10-29
fix: Make newline insertion more consistent.

dpurp:
Summary:
Currently newline insertion breaks when the previous node ends with a semi-colon, due to Closure Compiler's automatic semi-colon insertion. This PR changes the insertion logic to insert a newline before the node is written, so that it does not fall victim to semi-colon insertion.

That is, currently Gents outputs:

``` typescript
const foo = 4
;
export class Bar {}
```

which gets formatted by clang-format to:

``` typescript
const foo = 4;
export class Bar {}
```

With this PR, Gents will output:

``` typescript
const foo = 4;

export class Bar {}
```

### DIFF
--- a/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
+++ b/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
@@ -64,20 +64,23 @@ public class GentsCodeGenerator extends CodeGenerator {
 
   /** Add newlines to the generated source. */
   private void maybeAddNewline(Node n) {
-    Node previousNode = n.getPrevious();
     boolean hasComment =
         nodeComments.hasComment(n)
             || nodeComments.hasComment(n.getParent())
-            || isEmptyAndHasComment(previousNode)
-            || (n.getParent() != null && isEmptyAndHasComment(n.getParent().getPrevious()));
+            || isPreviousEmptyAndHasComment(n)
+            || (n.getParent() != null && isPreviousEmptyAndHasComment(n.getParent()));
 
     if (!hasComment && TOKENS_TO_ADD_NEWLINES_BEFORE.contains(n.getToken())) {
       add("\n");
     }
   }
 
-  private boolean isEmptyAndHasComment(Node n) {
-    return n != null && n.getToken() == Token.EMPTY && nodeComments.hasComment(n);
+  private boolean isPreviousEmptyAndHasComment(Node n) {
+    if (n == null || n.getParent() == null) {
+      return false;
+    }
+    Node prev = n.getPrevious();
+    return prev != null && prev.getToken() == Token.EMPTY && nodeComments.hasComment(prev);
   }
 
   /**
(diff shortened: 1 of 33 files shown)


src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -55,25 +56,30 @@ protected void add(Node n, Context ctx) {
...
-    if (nextNode != null) {
-      if (nodeComments.getComment(nextNode) == null // Comments already prepend a newline.
-          && TOKENS_TO_ADD_NEWLINES_BEFORE.contains(nextNode.getToken())) {
-        add("\n");
-      }
+  /** Add newlines to the generated source. */
+  private void maybeAddNewline(Node n) {
+    Node previousNode = n.getPrevious();
+    boolean hasComment = nodeComments.hasComment(n) || nodeComments.hasComment(n.getParent());
+    if (isEmptyAndHasComment(previousNode)
```
gmoothart: can you add a helper function so you can move the initialization for hasComment onto one line? something like 

```
boolean hasComment = nodeComments.hasComment(n) || nodeComments.hasComment(n.getParent()) || previousLineHasComment(previousNode)
```

src/test/java/com/google/javascript/gents/multiTests/goog_provide_rewrite/export.ts
```diff
@@ -1,3 +1,4 @@
+
```
gmoothart: Can you special-case the newline insertion so that it doesn't create one as the first line in a file?

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -55,25 +56,30 @@ protected void add(Node n, Context ctx) {
...
-    Node nextNode = n.getNext();
-    if (nextNode != null) {
-      if (nodeComments.getComment(nextNode) == null // Comments already prepend a newline.
-          && TOKENS_TO_ADD_NEWLINES_BEFORE.contains(nextNode.getToken())) {
-        add("\n");
-      }
+  /** Add newlines to the generated source. */
+  private void maybeAddNewline(Node n) {
+    Node previousNode = n.getPrevious();
+    boolean hasComment = nodeComments.hasComment(n) || nodeComments.hasComment(n.getParent());
```
dpurp: I don't know, but I don't have any test cases, where it's not.  (more levels or a traversal can always be added later)

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
dpurp: Switched format to be a series of ORs, in which case (due to google-java-format) inlining is more condensed.

src/test/java/com/google/javascript/gents/multiTests/goog_provide_rewrite/export.ts
dpurp: That's not easy to check, but in practice that won't actually happen because most files start with `import` which doesn't have a newline prepended.

PR #419, opened 2016-11-09
Fixes exception with anonymous functions and field extraction

### LINKED ISSUE #418: TypeConversionPass confused by arrow functions
When a constructor contains an arrow function that writes to methods on `this`, the TypeConversionPass throws an exception.

These members should be promoted to fields, but not with a field initializer.

input:
```javascript
class Klass {
  constructor() {
    (() => {
      this.field = null;
    })();
  }
}
```

output:
```
    CALL 22 [length: 35] [free_call: 1] [source_file: fields.js] is not a string node
      Node(EXPR_RESULT): fields.js:23:4
        this.m = 'baz';
      Parent(BLOCK): fields.js:22:9
      (() => {

        Caused by:
        java.lang.UnsupportedOperationException: CALL 22 [length: 35] [free_call: 1] [source_file: fields.js] is not a string node
```

expected output:
```typescript
class Klass {
  field: any;
  constructor() {
    (() => {
      this.field = null;
    })();
  }
}
```

gmoothart:
Anonymous functions have no name and were causing an exception when we
tried to get it.

Fixes #418

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -151,11 +151,7 @@ public final class TypeConversionPass implements CompilerPass {
         }
 
         Node fnNode = NodeUtil.getEnclosingFunction(n);
-        Node fnParent = fnNode.getParent();
-
-        // Use the QualifiedName if the function is on an object/namespace: `foo.moreFoo()`;
-        // otherwise, use the string on the node `function foo()` => `foo`.
-        String fnName = fnParent.isGetProp()? fnParent.getQualifiedName() : fnParent.getString();
+        String fnName = getFnName(fnNode);
 
         // TODO(gmoothart): in many cases we should be able to infer the type from the rhs if there
         // is no jsDoc
@@ -438,7 +434,7 @@ public final class TypeConversionPass implements CompilerPass {
 
     Node fnNode = NodeUtil.getEnclosingFunction(declaration.exprRoot);
     if (fnNode != null) {
-      String fnName = fnNode.getParent().getString();
+      String fnName = getFnName(fnNode);
       if (!"constructor".equals(fnName)) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -438,7 +434,7 @@ private boolean canPromoteFieldInitializer(ClassMemberDeclaration declaration) {
 
     Node fnNode = NodeUtil.getEnclosingFunction(declaration.exprRoot);
     if (fnNode != null) {
-      String fnName = fnNode.getParent().getString();
+      String fnName = getFnName(fnNode);
```
dpurp: NodeUtil.getName(fnNode)

per example cl

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -151,11 +151,7 @@ public void visit(NodeTraversal t, Node n, Node parent) {
         }
 
         Node fnNode = NodeUtil.getEnclosingFunction(n);
-        Node fnParent = fnNode.getParent();
-
-        // Use the QualifiedName if the function is on an object/namespace: `foo.moreFoo()`;
-        // otherwise, use the string on the node `function foo()` => `foo`.
-        String fnName = fnParent.isGetProp()? fnParent.getQualifiedName() : fnParent.getString();
+        String fnName = getFnName(fnNode);
```
dpurp: this could be a call `foo.getBar()`, so it's not quite a function name in closure-compiler's nomenclature

gmoothart: This is the enclosing function, not a function call, so I don't think that's true?

gmoothart:
@dpurp PTAL. I added and cleaned up some test cases, use NodeUtil.getName where I can, and renamed getFnName for clarity.

Issue #420, opened 2016-11-11
A lack of documentation 😕


lhecker:
Hey!

First of all: Thanks for the project!

The problem I'm now facing though is that due to a lack of documentation I can't for the life of me figure out how to properly use this tool. And it's not like I've tried: I have _successfully_ generated a d.ts file from a simple protobuf project, by using something like this:

```
./clutz \
    --externs es3.js \
    --externs es6.js \
    --externs ... \
    my_generated_protobuf_code.js \
    > out.d.ts
```

where `es{3,6}.js` are the externs files from the [closure compiler](https://github.com/google/closure-compiler/tree/master/externs). Here I already faced my first problem: It took me ages to figure out how to solve the e.g. `Array is undeclared` errors! I actually thought that including the `es6_min.js` file in this project would be enough to solve it, because that's how the current readme made it sound like. In hindsight it's of course very much obvious that you have to use the proper files from the closure compiler, but back then it wasn't.

What I still haven't solved is the fact that the generated .d.ts file is a whopping 540kB, which is funny because my .js file in comparison is only something like 30 kB. The reason for that is that clutz is including **everything** without doing a dependency analysis and thus also includes the entirety of all externs in the output (e.g. es3/es6 stuff which is huge).

I tried using `--depgraphs` because #102 made it sound like it could be used to trim out unneeded code, but then I noticed I can't call `./clutz --depgraphs file > out` because for some reason I still have to provide a input file and nowhere I can read up on how to use this feature or how it works, without reading through all the Java code. 😕

Due to that I want to kindly ask the maintainers of this project to please write up some documentation - even if it would only be a short one. I'm sure future users of clutz would really appreciate this. 🙂

**Edit:** I just found #297. That would already kinda solve the "540kB problem".

rkirov:
Hello :)

Sorry for the frustration you have experienced. Clutz fits within a niche of Google internal tools, and even though they are all open source, a lot of knowledge on how to use them and most importantly examples are kept inside google's internal repos. On top of it, we haven't had much external usage to justify expanding on the documentation. So thanks for speaking up, otherwise I wouldn't even know that there is in external interest :)

More practically can you review https://github.com/angular/clutz/pull/423?

[...]

PR #423, opened 2016-11-13
Improve docs

rkirov:


lhecker:
Hmm... I thought about it a bit more and I think this could be further extended with what you said [here](https://github.com/angular/clutz/issues/420#issuecomment-260165825). Specifically that clutz is a niche tool inside google, which means that some options (depgraphs) are basically not publicly accessible. Furthermore it still requires a lot of prior knowledge - especially what externs are and why they are any different to "regular input files". Unfortunately I don't know of any site describing what externs are to closure though.

BTW "regular input files": Are there even "input files"? The command line says:

```
Usage: clutz [options...] arguments...
```

What's "arguments"? I thought that's what you feed clutz as basically the "root" files for which a .d.ts is generated. I thought `--externs` are only for type resolution but not for actually turning them into type definitions. But while I'm writing this I actually realize that there are no input files but only externs, right? 😅

I guess that could still be clarified.

lhecker:
Just now I discovered another "trick": If you combine the `es{3,5,6}.js` files with the output of [calcdeps.py](https://developers.google.com/closure/library/docs/calcdeps) as the list for `--exports` it mostly just works! 😄

PR #426, opened 2016-11-17
Allow fields with inferred types to be parameter properties.

### LINKED ISSUE #425: infer field type from parameter when creating parameter properties
When annotating a class field with something like `/** @private @const */` (i.e., using type inference for the value of the field) parameter properties are not created for those fields. Since the names are the same and we know the type of the parameter, we should create the parameter property anyway.

input:
```javascript
/**
 * @constructor
 * @param {Foo} parameterPropWithInferredType
 */
function A(parameterPropWithInferredType) {
  /** @private @const */
  this.parameterPropWithInferredType = parameterPropWithInferredType;
}
```

output:
```typescript
class A {
  parameterPropWithInferredType: any;
  constructor(parameterPropWithInferredType: Foo) {
    this.parameterPropWithInferredType = parameterPropWithInferredType  
  }
}
```

desired output:
```typescript
class A {
  constructor(private readonly parameterPropWithInferredType: Foo) {}
}
```

gmoothart:
Fixes #425

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -160,7 +160,6 @@ public final class TypeConversionPass implements CompilerPass {
         // declaration of the form this.name = name;
         if ("constructor".equals(fnName) &&
             declaration.jsDoc != null &&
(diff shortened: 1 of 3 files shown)


mprobst:
Nice!

PR #427, opened 2016-11-18
Support enums nested in typedefs that are defined in externs

mprobst:
This surprising pattern is from an actual piece of code encountered in the wild.
It is an obscure corner case, and it is unclear whether the NoType check is
correct or might trigger in other programs. I will try this change out on a
larger code base and see if it breaks things.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -634,7 +634,7 @@ private boolean needsAlias(Set<String> shadowedSymbols, String provide, TypedVar
     }
     // Only var declarations have collisions, while class, interface, and functions can coexist with
     // namespaces.
```
evmar: Amend this comment.

PR #429, opened 2016-11-27
perf: Single quotes + free up completed compiler passes

dpurp:
- Don't store completed compiler passes in locals unless they are used
later.
- Add visibility & static modifiers where appropriate.
- Output single quotes instead of double quotes to reduce the work
needed by clang-format.


mprobst:
I wonder – did you measure a performance difference? There might be some earlier GC happening with this, but I wouldn't expect a substantial impact?

dpurp:
No, I didn't measure it.  We noticed some high memory usage when converting our project, which caused the OS to start paging out to disk, and this was intended to hit some potentially low hanging fruit to allow the GC to reclaim some memory earlier on (most of these passes have at least one HashMap that grows based on the number of files in the manifest).

PR #431, opened 2016-12-06
Validate the AST, print file name when codegen fails.

mprobst:
Gents already reports the failure location for the various code
visitors, but does not report a location when code generation fails. To
track down AST issues, also validate the AST before emit.

### DIFF
--- a/src/main/java/com/google/javascript/gents/NodeComments.java
+++ b/src/main/java/com/google/javascript/gents/NodeComments.java
@@ -34,6 +34,7 @@ public class NodeComments {
   }
 
   void replaceWithComment(Node oldNode, Node newNode) {
+    newNode.useSourceInfoFrom(newNode);
     oldNode.getParent().replaceChild(oldNode, newNode);
     moveComment(oldNode, newNode);
   }
--- a/src/main/java/com/google/javascript/gents/Options.java
+++ b/src/main/java/com/google/javascript/gents/Options.java
@@ -68,7 +68,7 @@ public class Options {
     // Report duplicate definitions, e.g. for accidentally duplicated externs.
     options.setWarningLevel(DiagnosticGroups.DUPLICATE_VARS, CheckLevel.ERROR);
 
-    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6);
+    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6_TYPED);
     options.setLanguageOut(CompilerOptions.LanguageMode.NO_TRANSPILE);
 
     // Do not transpile module declarations
--- a/src/main/java/com/google/javascript/gents/Options.java
+++ b/src/main/java/com/google/javascript/gents/Options.java
@@ -68,7 +68,7 @@ public class Options {
     // Report duplicate definitions, e.g. for accidentally duplicated externs.
     options.setWarningLevel(DiagnosticGroups.DUPLICATE_VARS, CheckLevel.ERROR);
 
-    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6_TYPED);
+    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6);
     options.setLanguageOut(CompilerOptions.LanguageMode.NO_TRANSPILE);
 
     // Do not transpile module declarations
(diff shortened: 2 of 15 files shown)


src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -176,26 +183,42 @@ private void generateTypeScript() {
 
     new StyleFixPass(compiler, comments).process(externRoot, srcRoot);
 
+    new AstValidator(
+            compiler,
+            new AstValidator.ViolationHandler() {
+              @Override
+              public void handleViolation(String message, Node n) {
+                compiler.report(JSError.make(n, CLUTZ_MALFORMED_AST, message));
```
calebegg: This isn't that helpful, actually. It just spits out errors with no context:

```
ERROR - Clutz produced a malformed AST: Expected 0 children, but was 1

ERROR - Clutz produced a malformed AST: Expected EMPTY but was NAME

ERROR - Clutz produced a malformed AST: Expected expression but was EXPORT
```

I'm not sure if that's more useful than nothing, unfortunately. Is there any way to get more detail here? At least a filename, and ideally a location?

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -176,26 +183,42 @@ private void generateTypeScript() {
...
+                .setTypeRegistry(compiler.getTypeRegistry())
+                .setCodeGeneratorFactory(factory)
+                .setPrettyPrint(true)
+                .setLineBreak(true)
+                .setOutputTypes(true)
+                .build();
+
+        sourceFileMap.put(filepath, tryClangFormat(tsCode));
+      } catch (Throwable t) {
+        System.err.println("Failed while converting " + file.getSourceFileName());
```
calebegg: This, otoh, is very helpful :-)

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
mprobst: Thanks for trying this! I've tried to add source location to places where we create new nodes, but I'm sure there are lots of places missing – the API doesn't make this easy retroactively.

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
mprobst: Yeah, but it's even better when I rethrow the exception so we don't just silently fail (`o_O`)

src/main/java/com/google/javascript/gents/NodeComments.java
```diff
@@ -28,12 +28,13 @@ void clearComment(Node n) {
 
   void moveComment(Node from, Node to) {
     if (getComment(from) != null) {
-      putComment(to, getComment(from));
+      putComment(to, getComment(from));
```
gmoothart: (nit) whitespace

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -34,6 +37,9 @@
  * TypeScript.
  */
 public class TypeScriptGenerator {
+  /** Diagnostic that indicates Clutz somehow produced an incorrect AST structure. */
+  private static final DiagnosticType CLUTZ_MALFORMED_AST = DiagnosticType.error(
```
gmoothart: s/clutz/gents ?

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
calebegg: Your rethrow below is actually stopping these errors from being printed at all -- if I comment it out, then I see the errors again. I don't know enough about the control flow here to make a suggestion though.

Your additional `.useSourceInfoFrom` calls seem to have added context to one of the three errors, but the other two are still showing up context-less. But I don't think that should block this PR.

mprobst:
PTAL. Turns out we cannot really use the AST validation (grr). I'm now just adding another error.

Also, revolutionary change, I'm now validating that we don't get errors from compilation! :-)

This is a bit nasty because I don't really have a test of us doing the wrong thing, so I cannot validate what I'm stumbling through properly. I'll try and binary search through Caleb's file to see what's going on.

src/main/java/com/google/javascript/gents/Options.java
```diff
@@ -68,7 +68,7 @@ public CompilerOptions getCompilerOptions() {
     // Report duplicate definitions, e.g. for accidentally duplicated externs.
     options.setWarningLevel(DiagnosticGroups.DUPLICATE_VARS, CheckLevel.ERROR);
 
-    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6);
+    options.setLanguage(CompilerOptions.LanguageMode.ECMASCRIPT6_TYPED);
```
calebegg: This doesn't seem to work internally, FYI. :-)

PR #432, opened 2016-12-14
Upgrade external dependencies to more recent versions.

mprobst:
These reflect the latest versions available both externally and internally.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,13 +41,13 @@ repositories {
 
 dependencies {
   runtime 'args4j:args4j:2.0.26'
-  runtime 'com.google.code.gson:gson:2.3.1'
+  runtime 'com.google.code.gson:gson:2.4.0'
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -41,13 +41,13 @@ repositories {
 
 dependencies {
   runtime 'args4j:args4j:2.0.26'
-  runtime 'com.google.code.gson:gson:2.3.1'
+  runtime 'com.google.code.gson:gson:2.4.0'
 
-  compile 'com.google.javascript:closure-compiler:v20161024'
-  compile 'com.google.guava:guava:18.0'
+  compile 'com.google.javascript:closure-compiler:v20161201'
```
evmar: Do we use the same updated version inside g3?  I imagine this .gradle file is ignored for the internal build?

mprobst: We use the internal released version. Best we can do here is to track it closely. Same for Guava.

PR #436, opened 2017-01-24
fix: Fix issue wherein all comments appear at the top of the file

dpurp:
- Fix issue wherein all comments appear at the top of the file (before all code)
 - Refactors CommentLinkingPass to improve readability.

More details:
When there is a newline between the @fileoverview and the goog.provide, then the comments strangely go out of sync.

== Works ====
```java
/** 
 * @fileoverview the quick brown fox jumps over the lazy dog 
 */
goog.provide('foo.bar')
```

== Broken ====
```java
/** 
 * @fileoverview the quick brown fox jumps over the lazy dog 
 * notice that there is a blank line after this
 */

goog.provide('foo.bar')
```

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -1,5 +1,6 @@
 package com.google.javascript.gents;
 
+import com.google.common.collect.ImmutableList;
 import com.google.javascript.jscomp.Compiler;
 import com.google.javascript.jscomp.CompilerPass;
 import com.google.javascript.jscomp.NodeTraversal;
@@ -65,7 +66,8 @@ public final class CommentLinkingPass implements CompilerPass {
   public void process(Node externs, Node root) {
     for (Node script : root.children()) {
       if (script.isScript()) {
-        List<Comment> comments = compiler.getComments(script.getSourceFileName());
+        ImmutableList<Comment> comments =
+            ImmutableList.copyOf(compiler.getComments(script.getSourceFileName()));
         NodeTraversal.traverseEs6(compiler, script, new LinkCommentsForOneFile(comments));
       }
     }
@@ -80,15 +82,15 @@ public final class CommentLinkingPass implements CompilerPass {
    */
   private class LinkCommentsForOneFile implements Callback {
     /** List of all comments in the file */
-    private final List<Comment> comments;
+    private final ImmutableList<Comment> comments;
     /** Buffer of all comments that are currently grouped together */
-    private List<Comment> group = new ArrayList<>();
+    private List<Comment> currentCommentGroup = new ArrayList<>();
     private int lastCommentIndex = 0;
 
-    private LinkCommentsForOneFile(List<Comment> comments) {
+    private LinkCommentsForOneFile(ImmutableList<Comment> comments) {
       this.comments = comments;
-      if (comments.size() > 0) {
-        group.add(comments.get(0));
+      if (!comments.isEmpty()) {
+        currentCommentGroup.add(comments.get(0));
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -65,7 +66,8 @@ public NodeComments getComments() {
   public void process(Node externs, Node root) {
     for (Node script : root.children()) {
       if (script.isScript()) {
-        List<Comment> comments = compiler.getComments(script.getSourceFileName());
+        ImmutableList<Comment> comments =
```
mprobst: I just followed the code path for my own curiosity. It turns out this will not need to superfluously copy the list of comments, because the underlying type is actually already an `ImmutableList` and `copyOf` returns (through many indirections) the `ImmutableList` itself in that case.

You could add a comment, in case somebody with the same curiosity finds this :-)

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -203,22 +199,29 @@ public final boolean shouldTraverse(NodeTraversal nodeTraversal, Node n, Node pa
...
+          parent.addChildBefore(newFloatingComment(currentCommentGroup), n);
+        }
+        addNextCommentTo(currentCommentGroup);
       }
 
       if (getLastLineOfCurrentComment() == line) {
         // Comment on same line as code
-        linkCommentToNode(n);
+        linkCommentGroupToNode(currentCommentGroup, n);
+
```
mprobst: nit: empty lines?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -102,54 +104,48 @@ private boolean hasRemainingComments() {
...
 
     /** Returns the starting line number of the next comment */
     private int getFirstLineOfNextComment() {
-      return comments.get(lastCommentIndex + 1).location.start.line + 1;
+      return getNextComment().location.start.line + 1;
     }
 
     /** Shifts a new comment into the group */
-    private void forceAddCommentToGroup() {
+    private void addNextCommentTo(List<Comment> commentGroup) {
```
mprobst: you only ever call this with `commentGroup == currentCommentGroup` - is that intentional? Would it be more readable to use the field then?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -102,54 +104,48 @@ private boolean hasRemainingComments() {
...
       }
       lastCommentIndex++;
     }
 
-    /**
-     * Shifts a new comment into the group.
-     * If the new comment is separated from the current one by at least a line, outputs the
-     * current group of comments.
-     */
-    private void addCommentToGroup(Node n) {
```
mprobst: why did you inline this method? It seems like it just duplicates the code?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -102,54 +104,48 @@ private boolean hasRemainingComments() {
...
-      if (getFirstLineOfNextComment() - getLastLineOfCurrentComment() > 1) {
-        n.getParent().addChildBefore(newFloatingComment(), n);
-      } else {
```
mprobst: the only functional change here is to move the `forceAdd...` below out of this `else` branch and unconditionally execute this, right? What does that accomplish?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: Yes, sorry, I might be a bit too familiar with the guava collections; I wanted to signal that this isn't intended to change.  Adding a comment about the no-copy, copy.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: I had a hard time understanding what this method was trying to do (the javadoc didn't quite match the name, and it's not obvious from the code that it "adds a comment to the group".   I ultimately decided since it was so hard to come up with an accurate name, that maybe this isn't the right type of abstraction (even though it is used twice).

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: > you only ever call this with `commentGroup == currentCommentGroup`
yes, though I was hoping to remove `currentCommentGroup`.   My issue here (and with these group methods in general) is that the API felt confusing...  "Add to which group?!" 

If you'd rather not pass an argument (I'm ok with reverting the method argument), I would like to at least suggest changing the name of the method to something along the lines of `addNextCommentToCurrentGroup` or something along those lines.  That being said, I have slight preference for the parameter, since I think it has a slightly lower cognitive load, since all of the code is in focus.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: more readability.. it turns out the "forceAdd" is used as part of the advance-to-the-beginning of the loop driving loop, and makes more sense IMO to be there ("forceAdd" isn't really forcing...)

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: what did they do to you ;) 

(removed)

dpurp:
Sorry, yes, I didn't add much background here.

I was trying to investigate the issue that @roseperrone was having with comments becoming unattached to their nodes, and was having some issues trying to trace my way through this code.  I was hoping the answer might become obvious, if I made the code easier to follow and tried to reduce the cognitive overhead of the class.  Unfortunately, it didn't fix the issue :( though IMO it's a bit easier to follow now.

dpurp:
... and I found the bug! Fix added to this PR.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
mprobst: I think it's less confusing to drop the parameter. I agree this code is a bit ugly to understand and naming isn't ideal, but the base concept of buffering up comments, then flushing, makes sense to me.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -203,22 +201,32 @@ public final boolean shouldTraverse(NodeTraversal nodeTraversal, Node n, Node pa
         return true;
```
gmoothart: Can you remove this block since you've moved it into the `while` ?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
gmoothart: `addCommentOrFlush` ?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
dpurp: Nope. (it may not patch the entry condition)

PR #438, opened 2017-01-27
Use isNormalBlock instead of isBlock, which will be deleted soon

MatrixFrog:


dpurp:
Will merge this once the external release with `isNormalBlock` is out.

PR #440, opened 2017-01-31
Convert to Java 8.

dpurp:
Change clutz/gents to use Java8 and run Refaster over the codebase.

Note: closure-compiler is pinned at Guava v20, so not all Guava Java 8
goodness is available yet (v21) adds new convenience methods (namely,
FluentIterable -> Streams; ImmutableList.toImmutableList (collector))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,4 +7,3 @@ cache:
   - $HOME/.gradle
 jdk:
   - oraclejdk8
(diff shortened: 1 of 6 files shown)


dpurp:
This seems to have confused Travis, so I'm going to close this PR and re-export it as #442

PR #443, opened 2017-02-15
Properly convert imports from already converted to TS files.

### LINKED ISSUE #333: Incremental file conversion in gents
Allow gents to be run incrementally where you only convert a few files at a time.
In each progressive pass, it should update the imports in the previous files.

rkirov:
Identify already converted files based on the prefix of their
goog.modules. Adds a new option --alreadyConvertedPrefix.

If the file is already in TS, the goog.require call needs to be
changed to an ES6 import.

Closes #333

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
@@ -54,13 +54,20 @@ public final class ModuleConversionPass implements CompilerPass {
   /** fileName, namespace -> local name */
   private final Table<String, String, String> typeRewrite = HashBasedTable.create();
 
+  private final String alreadyConvertedPrefix;
+
   public Table<String, String, String> getTypeRewrite() {
     return typeRewrite;
   }
 
-  public ModuleConversionPass(AbstractCompiler compiler, PathUtil pathUtil, NameUtil nameUtil,
-      Map<String, FileModule> fileToModule, Map<String, FileModule> namespaceToModule,
-      NodeComments nodeComments) {
+  public ModuleConversionPass(
+      AbstractCompiler compiler,
+      PathUtil pathUtil,
+      NameUtil nameUtil,
+      Map<String, FileModule> fileToModule,
+      Map<String, FileModule> namespaceToModule,
+      NodeComments nodeComments,
+      String alreadyConvertedPrefix) {
     this.compiler = compiler;
     this.pathUtil = pathUtil;
     this.nameUtil = nameUtil;
@@ -68,6 +75,7 @@ public final class ModuleConversionPass implements CompilerPass {
 
     this.fileToModule = fileToModule;
     this.namespaceToModule = namespaceToModule;
+    this.alreadyConvertedPrefix = alreadyConvertedPrefix;
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


rkirov:
@dpurp, @gmoothart It appears Intellij auto-formatting reformatted some comments in this file. I think I am using the google standard formater settings, but let me know if that is what you guys use. I want to avoid churn, where the next person reformats differently.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -443,9 +480,14 @@ void registerLocalSymbol(String sourceFile, String fullLocalName, String require
     typeRewrite.put(sourceFile, requiredNamespace, localName);
   }
 
-  /** Metadata about an exported symbol. */
+  /**
```
dpurp: are you switching away from using google-java-format?

src/main/java/com/google/javascript/gents/Options.java
```diff
@@ -54,6 +54,11 @@
       metaVar = "EXTERNSMAP")
   String externsMapFile = null;
 
+  @Option(name = "--alreadyConvertedPrefix",
+      usage = "Goog.modules starting with this prefix are assumed to be already in TypeScript",
+      metaVar = "ALREADY_CONVERTED_PREFIX")
+  String alreadyConvertedPrefix = "google3";
```
dpurp: should this be a list?

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
rkirov: aha, that explains it. Switched to google-java-format

src/test/java/com/google/javascript/gents/multiTests/converts_ts_module_require/converts_ts_module_require.js
```diff
@@ -0,0 +1,3 @@
+const {foo} = goog.require("google3.already.ts.module");
+const wholeModule = goog.require("google3.already.ts.module");
```
dpurp: should there be a test for 

// this one shouldn't match
const wholeModule = goog.require("google3.already.tsomething.else");

src/main/java/com/google/javascript/gents/Options.java
rkirov: A single prefix should work well enough for our usecase. Tsickle converts all .ts modules to goog.modules starting with 'google3', and I don't think any closure native modules start with that. Am I missing something?

src/test/java/com/google/javascript/gents/multiTests/converts_ts_module_require/converts_ts_module_require.js
rkirov: do you mean `const wholeModule = goog.require("still.closure");` because it cannot start with `google3`?

src/test/java/com/google/javascript/gents/multiTests/converts_ts_module_require/converts_ts_module_require.ts
```diff
@@ -0,0 +1,4 @@
+import './already_converted_to_ts_keep';
```
gmoothart: Does this automatically calculate the correct relative path? I was expecting you to just turn the namespace into a path, i.e. `google3/already/ts/module`

src/test/java/com/google/javascript/gents/multiTests/converts_ts_module_require/converts_ts_module_require.js
rkirov: done

src/test/java/com/google/javascript/gents/multiTests/converts_ts_module_require/converts_ts_module_require.ts
rkirov: yes, it is using the same utility function when both files are migrated.

rkirov:
merged

PR #445, opened 2017-02-24
Locking the node package dependencies.

rkirov:
Removes on demand installation of `typescript@next` as
that can create version mismatch between different developers.

Fixes some formatting that changed with clang-format 1.0.48.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ auto-complete) as if the imported code was written in TypeScript.
 
 We don't offer a binary distribution, so first you need to build:
 ```shell
-$ npm install clang-format
+$ npm install 
 $ gradle build installDist
 ...
 BUILD SUCCESSFUL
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -50,5 +50,6 @@ bug in Closure Compiler).
 ## Supported Version of TypeScript
 Clutz produces declaration files that are guaranteed to be accepted by a
 version of TypeScript
-[2.0.6](https://github.com/Microsoft/TypeScript/tree/v2.0.6). The current test
-suite runs against `typescript@next`, so that is always a good choice.
+[2.1.6](https://github.com/Microsoft/TypeScript/tree/v2.1.6). The current test
+suite runs against the version of `typescript` in `npm-shrinkwrap.json`, so that is
```
alexeagle: you decided against yarn? any particular reason?

the other option is just to pin the dependencies in the package.json file instead of shrinkwrapping

src/test/java/com/google/javascript/gents/singleTests/classes.ts
```diff
@@ -63,25 +63,25 @@ class F {
 
 
   /**
-     * block comment
-     */
+   * block comment
```
dpurp: yay!

README.md
rkirov: Travis supports only a single language per project. We have 'java'. In java environments npm seems to be present but not yarn. I tried to install yarn in the before_script, but it failed so I gave up on it for now.

https://travis-ci.org/angular/clutz/builds/204840916

PR #446, opened 2017-02-24
Adds a gents module rewrite log.

rkirov:
When configured with the option --log, gents will write the module
conversion into a log. This log will be used by tools downsteam to
modify files outside the scope of gents.

Each line of the log contains three comma separated strings - (OriginalName,
TSFile, DefaultExportName).

OriginalName is the original goog.module/goog.provide name.
TSFile - the name of the TypeScript file that contains the converted Closure module/namespace.
DefaultExportName - Gents never uses TS default exports. This means if the
  original goog.module had a default export we generated a named export with a
  specific name. This field contains that name. If there was no default export,
  the field contains the empty string.

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleRenameLogger.java
+++ b/src/main/java/com/google/javascript/gents/ModuleRenameLogger.java
@@ -1,0 +1,51 @@
+package com.google.javascript.gents;
+
+import com.google.gson.Gson;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Generates a log that maps how goog.modules/goog.provides where mapped to TS modules.
+ *
+ * <p>The log contains: - the original goog.module/goog.provide name. - the name of the TypeScript
+ * file that contains the converted Closure module/namespace. - Gents never uses TS default exports.
+ * This means if the original goog.module had a default export we generated a named export with a
+ * specific name. The field 'defaultRename' contains that name. If there was no default export, the
+ * field contains the empty string.
+ *
+ * <p>Ex: (in file buz.js) goog.module(foo.bar) class A { ... } exports = A;
+ *
+ * <p>is translated to buz.ts: export class A { ... }
+ *
+ * <p>Which generates the following log line: foo.bar,buz.ts,A
+ */
+class ModuleRenameLogger {
+  private class LogItem {
+    String originalName;
+    String tsFile;
+    String defaultRename;
+
+    LogItem(String originalName, String tsFile, String defaultRename) {
+      this.originalName = originalName;
+      this.tsFile = tsFile;
+      this.defaultRename = defaultRename;
+    }
+  }
+
+  private Gson gson = new Gson();
+
+  String generateModuleRewriteLog(
+      Set<String> filesToConvert, Map<String, CollectModuleMetadata.FileModule> namespaceMap) {
+    StringBuilder builder = new StringBuilder();
+    for (Map.Entry<String, CollectModuleMetadata.FileModule> entry : namespaceMap.entrySet()) {
+      String file = entry.getValue().file;
+      String defaultRename =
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -225,8 +227,41 @@ public CodeGenerator getCodeGenerator(Format outputFormat, CodeConsumer cc) {
...
     }
 
+    result.moduleRewriteLog = generateModuleRewriteLog(filesToConvert, modulePrePass);
     ((GentsErrorManager) compiler.getErrorManager()).doGenerateReport();
-    return sourceFileMap;
+    return result;
+  }
+
+  /**
+   * Generates a log file that maps how goog.modules/goog.provides where mapped to TS modules.
```
mprobst: Have you considered making the log structured data, e.g. a JSON structure? That'd make it easier to parse for many downstream tools, easier to understand (self describing), and easier to detect errors. Plus you save a on a bunch of escaping nightmares (e.g. files containing commas). Also makes for easier extensibility if we need more metadata down the line.

```
[
  {originalNamespace: "...", tsPath: "...", defaultExport: "..."}.
  ...
]
```

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -225,8 +227,41 @@ public CodeGenerator getCodeGenerator(Format outputFormat, CodeConsumer cc) {
...
+   */
+  private String generateModuleRewriteLog(
+      Set<String> filesToConvert, CollectModuleMetadata modulePrePass) {
+    String output = "";
+    for (String s : modulePrePass.getNamespaceMap().keySet()) {
+      FileModule tsModule = modulePrePass.getNamespaceMap().get(s);
+      String file = tsModule.file;
+      String defaultRename = tsModule.exportedNamespacesToSymbols.getOrDefault("exports", "");
+      if (filesToConvert.contains(file)) {
+        output += (s + "," + file + "," + defaultRename + "\n");
```
mprobst: Appending to a `String` in Java creates a new object, so this is O(n^2). Consider using a `StringBuilder`.

Here's much more than you need to know about it:
http://www.pellegrino.link/2015/08/22/string-concatenation-with-java-8.html

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
gmoothart: +1 for using json (the gson lib is already a dependency). I think ease of future extensibility is a significant argument in favor.

rkirov: done

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
rkirov: done

src/main/java/com/google/javascript/gents/ModuleRenameLogger.java
```diff
@@ -0,0 +1,51 @@
...
+      this.defaultRename = defaultRename;
+    }
+  }
+
+  private Gson gson = new Gson();
+
+  public String generateModuleRewriteLog(
+      Set<String> filesToConvert, Map<String, CollectModuleMetadata.FileModule> namespaceMap) {
+    StringBuilder builder = new StringBuilder();
+    for (String s : namespaceMap.keySet()) {
```
dpurp: prefer to use the entry set instead (peformance)

for (Entry<String, FileModule> entry : namespaceMap.entries())

rkirov: good call

PR #448, opened 2017-02-25
Emit proper json for rewrite log.

rkirov:
Rename log files to log.json.


rkirov:
of course joining with '\n' is not valid json :)

PR #456, opened 2017-03-06
Revert bazel's output file mapping when going over already converted TS.

rkirov:
When gents encounters 'require('google3.foo.bar')' it can map it back
to a tscikle emitted blaze-bin/foo/bar.es5.js. An extra back mapping
is required to map to foo/bar.ts, which is what the final import should
emit.


rkirov:
I am not happy with how this is being tested. Right now I cover the .es5 suffix but not the blaze-bin directory. On the other hand, a more faithful test would require a lot more messing around with the file loaded (to read blaze-bin and blaze-genfiles), which I decided against, but can reconsider.

rkirov:
@andresdominguez this should fix one of the issues we observed last Monday during the migration

rkirov:
Interesting, I did it and it is a bit better (but not much). It is still an ad-hoc reversing of a transformation another tool does before hand - this time tsickle instead of blaze. I like it better only because tsickle is a tool fully in our control.

Issue #390, opened 2016-10-13
Closure Object map type not converted to TS properly


gmoothart:
input:

``` javascript
/** @type {Object<string, number>} */
var map = {};
```

output:

``` typescript
var map: Object<string, number> = {};
```

desired output:

``` typescript
let map: {[s: string]: number} = {};
```

PR #459, opened 2017-03-08
Convert Closure Object map type to TS properly

bowenni:
Fix https://github.com/angular/clutz/issues/390
@rkirov PTAL

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
@@ -313,6 +313,16 @@ public final class TypeAnnotationPass implements CompilerPass {
               if ("Array".equals(typeName)) {
                 return arrayType(convertTypeNodeAST(block.getFirstChild()));
               }
+
+              // Convert index signature types
+              if ("Object".equals(typeName)) {
+                TypeDeclarationNode indexSignatureNode =
+                    indexSignatureType(
+                        convertTypeNodeAST(block.getFirstChild()),
+                        convertTypeNodeAST(block.getSecondChild()));
+                return indexSignatureNode;
+              }
+
               // Convert generic types
               return parameterizedType(
                   root,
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -405,6 +415,29 @@ public TypeDeclarationNode convertTypeNodeAST(Node n, boolean isReturnType) {
...
   }
 
+  /** Returns a new node representing an index signature type. */
+  TypeDeclarationNode indexSignatureType(
+      TypeDeclarationNode keyType, TypeDeclarationNode valueType) {
+    TypeDeclarationNode node = new TypeDeclarationNode(Token.INDEX_SIGNATURE);
+    TypeDeclarationNode first = null;
+    switch (keyType.getToken()) {
+      case STRING_TYPE:
+        first = new TypeDeclarationNode(Token.STRING_KEY, "s");
```
rkirov: use 'key' instead of 's'

src/test/java/com/google/javascript/gents/singleTests/object_map.js
```diff
@@ -0,0 +1,8 @@
+/** @type {Object<string, number>} */
+var map1 = {};
+
+/** @type {Object<number, string>} */
+var map2 = {1: "foo"};
+
+/** @type {Object<number, string|number>} */
```
rkirov: also try a nested one /** @type {Object<string, Object<number, number>>} */

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
bowenni: Done.

src/test/java/com/google/javascript/gents/singleTests/object_map.js
bowenni: Added.

Issue #458, opened 2017-03-08
Add errorprone check to gradle


dpurp:
http://errorprone.info/docs/installation#gradle

Example: https://github.com/google/error-prone/blob/master/examples/gradle/build.gradle

PR #460, opened 2017-03-08
add error prone to the build.

rkirov:
Fixes existing error prone issues.

Could not figure out how to make error prone warnings errors. So for now
just print the warnings during building.


rkirov:
It works for me locally but fails on travis - https://travis-ci.org/angular/clutz/builds/208821370
No idea why :(

dpurp:
Fixes #458

build.gradle
```diff
@@ -72,4 +75,4 @@ test {
```
dpurp: do you need a classpath/runtime dependency on error-prone as well?

https://github.com/google/error-prone/blob/master/examples/gradle/build.gradle#L9

rkirov:
according to https://plugins.gradle.org/plugin/net.ltgt.errorprone this should be enough for gradle 2.1+

dpurp:
I wonder if this is the issue:
http://stackoverflow.com/questions/37140114/gradle-check-fails-in-travis-ci

rkirov:
Gradle wrapper fixed the issue and it appear that Travis is caching the extra downloads (gradle 3.2), so ready to PTAL.

Issue #453, opened 2017-03-03
local alias for a goog.provide base breaks gents


rkirov:
The following pattern:

```javascript
goog.provide('a.b.c.Base');

goog.scope(function() {
var c = a.b.c;

c.Base = function() {...}
});
```

makes gents emit

```javascript
let c = a.b.c;

c.Base = class { ... }
```

which is not correct. Emitting simply:

```javascript
export class Base { ... }
```

should work.

PR #461, opened 2017-03-10
Fix local alias in goog.scope for goog.provide

bowenni:
@rkirov @dpurp Please review my attempt to fix https://github.com/angular/clutz/issues/453
Two fixes:
(1) The original assumption is that the node returned by `maybeRecordAndRemoveAlias()` is ought to be moved. This is not true. In some cases (for example `var foo = a.b.c;` where `a.b.c` comes from `goog.provide`), we detach this node and return the next node which has not been checked at all. We shouldn't simply move the next node in this case.
(2) Instead of checking if the alias matches goog.provide precisely, we could check if the alias matches goog.provide partially so that local partial alias works.

### DIFF
--- a/src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
+++ b/src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
@@ -73,11 +73,18 @@ public final class RemoveGoogScopePass extends AbstractTopLevelCallback implemen
 
     @Nullable Node nodeToMove = blockOfScopeContents.getFirstChild();
     while (nodeToMove != null) {
-      nodeToMove = maybeRewriteAlias(nodeToMove);
-      if (nodeToMove == null) {
-        break;
+      // After maybeRewriteAlias, nodeToMove is either (1)already detached or (2)needs to be moved.
+      @Nullable Node nodeToCheck = maybeRewriteAlias(nodeToMove);
+      // (1) nodeToMove is detached, alias to provided namespace is recorded, the next node is
+      // returned. The next node should be 'maybeRewriteAlias()' checked before it can be moved out
+      // of goog.scope
+      if (nodeToCheck != null) {
+        nodeToMove = nodeToCheck;
+        continue;
       }
 
+      // (2) Alias is re-assigned with the provided namespace. In this case, node is not detached
+      // and null is returned. nodeToMove needs to be moved out of goog.scope
       // Store the next node in a temp variable since detaching the node breaks the chain.
       Node nextNodeToMove = nodeToMove.getNext();
       nodeToMove.detachFromParent();
@@ -92,18 +99,23 @@ public final class RemoveGoogScopePass extends AbstractTopLevelCallback implemen
     compiler.reportCodeChange();
   }
 
+  /**
+   * If the node is a local alias declaration for a provided namespace then store mapping and detach
+   * the node. If the node is an assignment for a local alias's property then rewrite the local
+   * alias.
+   *
+   * @return the next node if the current node is detached, return null otherwise.
+   */
   private Node maybeRewriteAlias(Node node) {
     switch (node.getFirstChild().getToken()) {
       case NAME:
-        node = maybeRecordAndRemoveAlias(node.getFirstChild());
-        break;
+        return maybeRecordAndRemoveAlias(node.getFirstChild());
       case ASSIGN:
-        maybeReasignAlias(node.getFirstChild());
-        break;
+        maybeReassignAlias(node.getFirstChild());
+        return null;
       default:
-        break;
+        return null;
     }
-    return node;
   }
 
   private Node maybeRecordAndRemoveAlias(Node assign) {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/gents/singleTests/goog_scope_local_alias.js
```diff
@@ -0,0 +1,29 @@
...
+});
+
+goog.scope(function() {
+/** @constructor */
+a.b.c.Base3 = function() {}
+});
+
+goog.scope(function() {
+var c = a.b.c;
+var foo = 1;
```
gmoothart: what is this case testing?

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -93,40 +100,50 @@ private void rewriteGoogScope(Node n) {
   }
 
   private Node maybeRewriteAlias(Node node) {
+    // Return the next node if the current node is detached.
```
gmoothart: Can you flesh this out into documentation for the method and put it in a jsdoc ?

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -93,40 +100,50 @@ private void rewriteGoogScope(Node n) {
...
+    for (String providedNamespace : providedNamespaces) {
+      @Nullable String rhsQualifiedName = node.getQualifiedName();
+      if ((null != rhsQualifiedName) && providedNamespace.contains(node.getQualifiedName())) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private Node maybeReassignAlias(Node assign) {
```
gmoothart: This function always returns null? That doesn't seem right.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -93,40 +100,50 @@ private void rewriteGoogScope(Node n) {
...
     }
-    return next;
+    return null;
   }
 
-  private void maybeReasignAlias(Node assign) {
+  private boolean isInProvidedNameSpace(Node node) {
+    for (String providedNamespace : providedNamespaces) {
+      @Nullable String rhsQualifiedName = node.getQualifiedName();
+      if ((null != rhsQualifiedName) && providedNamespace.contains(node.getQualifiedName())) {
```
gmoothart: should this be `.startsWith` instead of `.contains` ?

src/test/java/com/google/javascript/gents/singleTests/goog_scope_local_alias.js
dpurp: I'm guessing this is intended to verify that `foo` is not exported.   Instead of naming the var `foo`, can you use a name that describes what its intended goal is?

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -93,40 +100,50 @@ private void rewriteGoogScope(Node n) {
...
   private Node maybeRecordAndRemoveAlias(Node assign) {
     Node next = assign.getParent();
     Node lhs = assign;
     Node rhs = assign.getLastChild();
     if (rhs == null) { // var foo;
-      return next;
+      return null;
     }
-    if (providedNamespaces.contains(rhs.getQualifiedName())) {
+    if (isInProvidedNameSpace(rhs)) {
```
dpurp: s/NameSpace/Namespace

src/test/java/com/google/javascript/gents/singleTests/goog_scope_local_alias.js
bowenni: Yes this is testing non-`goog.provide`d local variables are not exported.
Fixed the name.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -92,37 +99,53 @@ private void rewriteGoogScope(Node n) {
...
       compiler.reportCodeChange();
+      return next;
+    }
+    return null;
+  }
+
+  private boolean isInProvidedNamespace(Node node) {
+    for (String providedNamespace : providedNamespaces) {
+      @Nullable String rhsQualifiedName = node.getQualifiedName();
+      if ((null != rhsQualifiedName) && providedNamespace.startsWith(node.getQualifiedName())) {
```
dpurp: nit: can you reverse the null check? `rhsQualifedName != null`

src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -73,11 +73,18 @@ private void rewriteGoogScope(Node n) {
...
     while (nodeToMove != null) {
-      nodeToMove = maybeRewriteAlias(nodeToMove);
-      if (nodeToMove == null) {
-        break;
+      // After maybeRewriteAlias, nodeToMove is either (1)already detached or (2)needs to be moved.
+      @Nullable Node nodeToCheck = maybeRewriteAlias(nodeToMove);
+      // (1) nodeToMove is detached, alias to provided namespace is recorded, the next node is
+      // returned. The next node should be 'maybeRewriteAlias()' checked before it can be moved out
+      // of goog.scope
+      if (null != nodeToCheck) {
```
dpurp: nit: can you reverse the check `nodeToCheck != null`

bowenni: Fixed.

I was under the impression that constants on the left are [better](http://www.moserware.com/2008/01/constants-on-left-are-better-but-this.html) but Java isn't vulnerable to this problem.

bowenni:
travis doesn't seem to run a new build. "Restart build" restarts a build without the new commits.

PR #462, opened 2017-03-10
Add AnyDuringMigration type

alexeagle:


rkirov:
I think this is the wrong place for this, nothing to do with clutz or closure.

alexeagle:
We'll put it in the TS stdlib instead.

Issue #404, opened 2016-10-22
test: compile .ts code to check validity


mprobst:
We should have a test that compiles the generated `.ts` code to check that we actual output valid TypeScript.

bowenni:
Currently a lot of generated TS in our tests are broken. Even some inputs are invalid Closure code.
As a first step, we will fix the tests before compiling .ts code.

mprobst:
We used to have an integration test that would pipe all .ts files into
TypeScript:
https://github.com/angular/clutz/blob/master/src/test/java/com/google/javascript/clutz/DeclarationSyntaxTest.java

Does that no longer run, or does it not cover enough?

bowenni:
Still runs but only for `.d.ts`?

mprobst:
Ah I see, I didn't get that this was specifically about `.ts` as opposed to
`.d.ts`. Makes sense.

bowenni:
Close via #488

PR #476, opened 2017-03-28
fix(gents): PR1/2 Fix some gents tests so inputs and outputs can actually compile.

bowenni:
Prepare for https://github.com/angular/clutz/issues/404

### DIFF
--- a/src/test/java/com/google/javascript/gents/singleTests/comments.js
+++ b/src/test/java/com/google/javascript/gents/singleTests/comments.js
@@ -28,19 +28,27 @@ A.prototype.foo = function() {};
 // This is just some extra stuff
 var foo = function(deleted, notdeleted) { return deleted + notdeleted; };
 
-// The following comment should be mostly deleted
+// The following comments should be mostly deleted
 /**
- * @constructor
- * @extends {A}
  * @type {number}
  * @private {number}
  * @protected {number}
  * @param {number} foo description of foo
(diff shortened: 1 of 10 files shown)


src/test/java/com/google/javascript/gents/singleTests/comments.js
```diff
@@ -40,7 +40,7 @@ var foo = function(deleted, notdeleted) { return deleted + notdeleted; };
  * @package {number}
  * @const {number}
  */
-var x;
+var x = 1;
```
dpurp: huh... this is special.  

i'm not sure this can be both a `@constructor` and `@const {number}`

bowenni: It will need to be split into more variables because most of these annotations are not compatible with each other. It defeats the purpose of this test to have many invalid annotations as the comment says `// The following comment should be mostly deleted`.
I wonder if we should just remove this test case since this is unlikely to compile.

gmoothart: As long as there is a test for each of these being removed, I don't think it matters if they are on different variables.

bowenni: Done.
PTAL.

bowenni:
Friendly ping.

PR #479, opened 2017-03-28
Use `.destroyForcibly`

bowenni:
AFAIK we're on JDK 8?

btw I'm trying to generalize this file to gents so we can compile `.ts`


src/test/java/com/google/javascript/clutz/DeclarationSyntaxTest.java
```diff
@@ -66,6 +67,8 @@ public boolean accept(File dir, String name) {
           "--noImplicitAny",
           "--strictNullChecks");
 
+  private static final int MAXIMUM_TIME_TO_RUN_TSC_IN_SECONDS = 2;
```
mprobst: I'd inline this, it's not like you'll reuse the constant, and it's perfectly readable with the TimeUnit.SECONDS below.

PR #481, opened 2017-03-29
fixes rewriteGoogProvide when alias is last

rkirov:
Introduces a RewriteStatus inner class to track 'stillAttached'
state, separate from a null Node, which could mean nothing more to
process.


src/main/java/com/google/javascript/gents/RemoveGoogScopePass.java
```diff
@@ -100,39 +100,56 @@ private void rewriteGoogScope(Node n) {
...
       default:
-        return null;
+        return stillAttached;
     }
   }
 
-  private Node maybeRecordAndRemoveAlias(Node assign) {
-    Node next = assign.getParent();
+  /** Sentinel value used to express that the node is still attached. */
+  static RewriteStatus stillAttached = new RewriteStatus(IR.number(0));
```
bowenni: Should this be a static final constant?

rkirov: definitely, done

bowenni:
Much cleaner! Thanks!

PR #489, opened 2017-04-05
Support non-nullable/nullable annotated inner typedefs

bowenni:


src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -199,14 +199,19 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-            //            STRING_KEY a
-            //            STRING number
-            //        COLON
-            //            STRING_KEY b
-            //            STRING number
+            if ((typedefTypeRoot.getToken() == Token.BANG)
+                || (typedefTypeRoot.getToken() == Token.QMARK)) {
+              typedefTypeRoot = typedefTypeRoot.getFirstChild();
+            }
+            //(BANG|QMASK)
```
mprobst: typo QMARK

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -199,14 +199,19 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+            //    LC
+            //        LB
```
mprobst: what if I have a typedef that's a simple alias?

```js
/** @typedef {string} */
var MyStringType;
```

In general, a typedef can contain any arbitrary type JSDoc tree – we must be careful that we actually handle those trees. If it is an invariant guaranteed by other code that there can only be member defs here, we should at least add a comment pointing to the code, or better try to move the code together in one location.

bowenni: Yeah this invariant doesn't hold if the input is a class inner typedef with a simple type alias.

To fix this, I think it doesn't make sense to convert a simple alias to a TypeScript interface -  it should be a type alias instead. We should only create an interface if it's an inner typedef and it contains an object. WDYT?

mprobst: I agree.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -199,14 +199,19 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-            //            STRING_KEY a
-            //            STRING number
-            //        COLON
-            //            STRING_KEY b
-            //            STRING number
+            if ((typedefTypeRoot.getToken() == Token.BANG)
+                || (typedefTypeRoot.getToken() == Token.QMARK)) {
+              typedefTypeRoot = typedefTypeRoot.getFirstChild();
+            }
+            //(BANG|QMARK)
```
mprobst: nit: always follow "`//`" with whitespace: "`// `"

src/test/java/com/google/javascript/gents/singleTests/inner_types.js
```diff
@@ -45,11 +61,17 @@ MyClass.InnerTypedef;
...
+MyClass.InnerMyStringType;
+
 /**
  * @typedef {{
  *     a: {b: {c: number}},
  * }}
  */
 var Typedef;
 
+/** @typedef {string} */
```
mprobst: I think there's one more thing missing: 

```js
/** @typedef {?string} */
var MyOptionalString;
```

Same for `!`.

bowenni: Done.

Sorry! Keep missing cases!

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -87,50 +87,78 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-            types.put(interfaceName, interfaceMember);
-            interfaceMember.setJSDocInfo(bestJSDocInfo);
-            Node interfaceNode = new Node(Token.INTERFACE, IR.empty(), IR.empty(), interfaceMember);
-            Node nameNode = Node.newString(Token.NAME, interfaceName);
-            nameNode.addChildToBack(interfaceNode);
-            Node exportNode = new Node(Token.EXPORT, new Node(Token.CONST, nameNode));
-            replaceExpressionOrAssignment(n, parent, exportNode);
+            Node typedefNode = bestJSDocInfo.getTypedefType().getRoot();
+            if (containsSimpleTypes(typedefNode)) {
+              createTypeAlias(n, parent);
```
mprobst: can you break or return after this, instead of an indented else?

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -87,50 +87,78 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-          // members converted in TypeAnnotationPass.
+          // Converts a class inner typedef into either (1) a top level interface, which then later
+          // has its members converted in TypeAnnotationPass; or (2) a type alias.
           // Most class inner @typedef meant @record in closure but they were added before @record
           // was supported. Also in TypeScript interfaces are preferred to type alias because of
-          // better error reporting and ability to extend.
+          // better error reporting and ability to extend. However simple types such as string are
+          // still type aliases.
           bestJSDocInfo = NodeUtil.getBestJSDocInfo(n);
           if (bestJSDocInfo != null && bestJSDocInfo.hasTypedefType()) {
```
mprobst: could you invert this condition and early exit with a `break` or so, instead of adding more and more indent?

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -87,50 +87,78 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-            replaceExpressionOrAssignment(n, parent, typeDef);
-          }
+          createTypeAlias(n, parent);
           break;
         default:
           break;
       }
     }
 
+    private boolean containsSimpleTypes(Node typedefNode) {
```
mprobst: This should be the other way around – we should whitelist specifically the pattern we support, and reject everything else.

Our code depending on this only supports a specific pattern, so we should detect the specific pattern and handle it. I.e. this should be called `containsRecordType` and you should invert the condition. I think the logic is fine though.

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -87,50 +87,78 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+      }
+
+      return true;
+    }
+
+    private void createTypeAlias(Node n, Node parent) {
+      JSDocInfo bestJSDocInfo = NodeUtil.getBestJSDocInfo(n);
+      if (bestJSDocInfo != null && bestJSDocInfo.hasTypedefType()) {
+        String name;
+        if (n.getToken() == Token.NAME) {
```
mprobst: maybe should be a switch?

bowenni:
Addressed review comments.

src/test/java/com/google/javascript/gents/singleTests/inner_types.ts
```diff
@@ -24,13 +24,34 @@ export interface InnerTypedefWithAssignment {
   b: number;
 }
 
+export interface InnerTypedefNonNullable {
+  a: number;
+  b: number;
+}
+
+export interface InnerTypedefNullable {
```
rkirov: add a //!! comment that intentionally we loose the nullability here

PR #492, opened 2017-04-11
fix(gents): fix an edge case when a GETPROP node contains inline comments that are more than one line before the property name node.

bowenni:
Problem:
```javascript
a(
// comment
)
.b();
```
adds an EMPTY node before `b` node and attaches the comment to that EMPTY node. This breaks [GEPROP's children count assertion in Closure CodeGenerator](https://github.com/google/closure-compiler/blob/120d1c7a8d96d9c81a92e6600e1c0e2c7cbdcb98/src/com/google/javascript/jscomp/CodeGenerator.java#L737-L738)

We want to map the comment to a node, either `a` or `b` in this example. 
I choose to map the comment to `a` because mapping the comment to `b` requires overriding [code generator for GETPROP nodes](https://github.com/google/closure-compiler/blob/120d1c7a8d96d9c81a92e6600e1c0e2c7cbdcb98/src/com/google/javascript/jscomp/CodeGenerator.java#L722-L757).

In general, this (adding a new EMPTY node) may break not only GETPROP nodes but any nodes that have children count assertions in code generator. ~~But the counter examples are probably too contrived to worry about?~~ *Too early to call. Here's another counter example:*
```javascript
if (a) {
  // comment
}
else {
}
```
produces 4 children with one EMPTY node:
```
IF 1 [length: 23]
    NAME a 1 [length: 1]
    BLOCK 1 [length: 7]
    EMPTY
    BLOCK 4 [length: 3]
```
and this [breaks closure code generator](https://github.com/google/closure-compiler/blob/120d1c7a8d96d9c81a92e6600e1c0e2c7cbdcb98/src/com/google/javascript/jscomp/CodeGenerator.java#L817)

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -180,6 +180,22 @@ public final class CommentLinkingPass implements CompilerPass {
       return c;
     }
 
+    /**
+     * If the parent is NOT a GETPROP node, then outputs floating comments by attaching to a new
+     * EMPTY node before the current node. If the parent is a GETPROP node, then attach the comments
+     * to the parent node to avoid messing up GETPROP node's children count.
+     *
+     * @param n current node
+     * @param parent parent node
+     */
+    private void outputFloatingCommentFromBufferToParentOrCurrentNode(Node n, Node parent) {
+      if (parent.getToken() == Token.GETPROP) {
+        linkCommentBufferToNode(parent);
+      } else {
+        parent.addChildBefore(newFloatingCommentFromBuffer(), n);
+      }
+    }
+
     /** Returns if the current comment is directly adjacent to a line. */
     private boolean isCommentAdjacentToLine(int line) {
       int commentLine = getLastLineOfCurrentComment();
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -223,12 +241,14 @@ public final boolean shouldTraverse(NodeTraversal nodeTraversal, Node n, Node pa
       } else if (getLastLineOfCurrentComment() == line - 1) {
         // Comment ends just before code
         linkCommentBufferToNode(n);
-      } else if (!hasRemainingComments()) {
-        // Exhausted all comments, output floating comment before current node
-        parent.addChildBefore(newFloatingCommentFromBuffer(), n);
+      } else if (!hasRemainingComments() && !outputCommentAfterLine) {
+        // Exhausted all comments, output floating comment before current node or at parent node.
```
bowenni: We haven't necessarily exhausted all comments yet. It's possible that we just outputted the current group of comments in the while loop above and shifted the next comment in buffer (which is the real last comment, therefore `!hasRemaingComments()`).

src/test/java/com/google/javascript/gents/singleTests/comments.ts
```diff
@@ -36,3 +36,17 @@ const x;
...
 /** @export */
 let m: number = 4;
+let a = function() {
+  return {b: function() {}};
+};
+let c = function() {};
+
+// comment before GETPROP
+
+// comment in GETPROP
```
bowenni: This comment is above `a()` after migration which is not ideal.
This is due to the limitation that comments are mapped from nodes.

bowenni:
Not ready for code review yet.
There are more edge cases not considered...

Issue #493, opened 2017-04-12
comment pass creates new EMPTY nodes that confuses code generator


bowenni:
Background:
Our comments conversion has a limitation that [a comment must be attached to a node](https://github.com/angular/clutz/blob/fa3c63542df96b217b269afa49a01d453a208e11/src/main/java/com/google/javascript/gents/NodeComments.java#L9). This way, when code generator emits TS code, it knows where to [append the corresponding comment](https://github.com/angular/clutz/blob/fa3c63542df96b217b269afa49a01d453a208e11/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java#L31-L35) to the node.

Problem0:
Given a comment that is more than one line above the next node(not exactly the same as next line, see example below), we [create a new EMPTY node to hold that comment](https://github.com/angular/clutz/blob/fa3c63542df96b217b269afa49a01d453a208e11/src/main/java/com/google/javascript/gents/CommentLinkingPass.java#L215). This breaks the children count assertion in code generator for some nodes (such as [`GETPROP` expects 2 children](https://github.com/google/closure-compiler/blob/120d1c7a8d96d9c81a92e6600e1c0e2c7cbdcb98/src/com/google/javascript/jscomp/CodeGenerator.java#L737-L738), [`IF` expects 2 or 3 children](https://github.com/google/closure-compiler/blob/120d1c7a8d96d9c81a92e6600e1c0e2c7cbdcb98/src/com/google/javascript/jscomp/CodeGenerator.java#L817)).
Example:
```javascript
a(
// comment
)
.b();  <- next node is here.
```
```javascript
if (a) {
  // comment
}
else {  <- next node is 2 lines away from the comment
}
```
produces 4 children:
```
IF 1 [length: 23]
    NAME a 1 [length: 1]
    BLOCK 1 [length: 7]
    EMPTY
    BLOCK 4 [length: 3]
```

Problem1:
Only comments attached to nodes that requires code generator to call [`add()` function](https://github.com/angular/clutz/blob/fa3c63542df96b217b269afa49a01d453a208e11/src/main/java/com/google/javascript/gents/GentsCodeGenerator.java#L28) will be kept. Other comments are stored in the `nodesToComment` map but are never used.
[...]

PR #494, opened 2017-04-12
fix(gents) - Attempt to fix some comments conversion issues

bowenni:
Fixes https://github.com/angular/clutz/issues/493.
Current error pattern:
```javascript
a(
// comment
)
.b();
```
```javascript
if (a) {
  // comment
}
else {
}
```

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -1,6 +1,7 @@
 package com.google.javascript.gents;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import com.google.javascript.jscomp.Compiler;
 import com.google.javascript.jscomp.CompilerPass;
 import com.google.javascript.jscomp.NodeTraversal;
@@ -50,6 +51,20 @@ public final class CommentLinkingPass implements CompilerPass {
   private static final Pattern[] COMMENT_REPLACEMENTS = {
     Pattern.compile("//\\s*goog.scope\\s*(?<keep>)")
   };
+
+  /**
+   * These nodes can expand their children with new EMPTY nodes. We will use that as a placeholder
+   * for floating comments.
+   */
+  private static final ImmutableSet<Token> TOKENS_CAN_HAVE_EMPTY_CHILD =
+      ImmutableSet.of(Token.SCRIPT, Token.MODULE_BODY, Token.BLOCK);
+  /**
+   * Comments assigned to these nodes will not be emitted during code generation because these nodes
+   * don't use add(Node n, Context ctx). Therefore we avoid assigning comments to them.
+   */
+  // TODO(bowenni): More nodes missing here?
+  private static final ImmutableSet<Token> TOKENS_IGNORE_COMMENTS =
+      ImmutableSet.of(Token.LABEL_NAME, Token.NAME, Token.GENERIC_TYPE, Token.BLOCK);
 
   private final Compiler compiler;
   private final NodeComments nodeComments;
@@ -136,6 +151,10 @@ public final class CommentLinkingPass implements CompilerPass {
 
     /** Flushes the comment buffer, linking it to the provided node. */
     private void linkCommentBufferToNode(Node n) {
+      if (!canHaveComment(n)) {
+        linkCommentBufferToNode(n.getParent());
+        return;
+      }
       StringBuilder sb = new StringBuilder();
       String sep = "\n";
       for (Comment c : commentBuffer) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -223,12 +291,14 @@ public final boolean shouldTraverse(NodeTraversal nodeTraversal, Node n, Node pa
       } else if (getLastLineOfCurrentComment() == line - 1) {
         // Comment ends just before code
         linkCommentBufferToNode(n);
-      } else if (!hasRemainingComments()) {
-        // Exhausted all comments, output floating comment before current node
```
bowenni: We haven't necessarily exhausted all comments yet. It's possible that we just outputted the current group of comments in the while loop above and shifted the next comment in buffer (which is the real last comment, therefore `!hasRemaingComments()`).

src/test/java/com/google/javascript/gents/singleTests/comments.js
```diff
@@ -44,3 +44,28 @@ var x;
...
+var a = function() {
+  return {
+    b: function() {}
+  }
+};
+var c = function() {};
+
+// comment before GETPROP
+a(
+// comment in GETPROP
```
rkirov: I am still confused why this is in GETPROP and not in FUNC_CALL (or something of sorts).

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -180,6 +210,42 @@ private Node newFloatingCommentFromBuffer() {
...
+        outputFloatingCommentFromBuffer(parent);
+      }
+    }
+
+    private boolean canHaveComment(Node n) {
+      if (TOKENS_SET_IGNORE_COMMENTS.contains(n.getToken())) {
+        return false;
+      }
+
+      // Special case: GETPROP node's second child(the property name) loses comments.
```
rkirov: second or last? Do we even have guarantees GETPROP has only two children? This AST is atrocious :)

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -180,6 +210,42 @@ private Node newFloatingCommentFromBuffer() {
...
+     * sensitive about its children count/order, then try to put the comment in current node. If the
+     * current node doesn't accept comments (for example a BLOCK node will ignore any comments
+     * mapped to it), then attach the comments to the parent node. The parent node might not be the
+     * best place to hold the comment but at least we don't lose the comment and we don't break code
+     * generation.
+     *
+     * @param n
+     */
+    private void outputFloatingCommentFromBuffer(Node n) {
+      Node parent = n.getParent();
```
rkirov: it the worst case scenario we hit ROOT and  ROOT.getParent() is null, so maybe add if (n == null) return;
I am ok with dropping some comments in extreme situations instead of crashing with NPE.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
rkirov: Instead of being smart in interleaving collecting comments and writing them, maybe down the road we can refactor this as a two step process:
- collect all comments (moving them up and down as we see fit)
- emit all comment from NodeComments mapping.
I think this will avoid such problems (if I understand them correctly)

bowenni: Discussed offline. This is OK for a bug fix.
In general this is confusing here. A lot of invariants are hidden in code. We should consider refactoring if we encounter more issues with comments.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -51,6 +54,29 @@
...
+  /**
+   * These nodes can expand their children with new EMPTY nodes. We will use that as a placeholder
+   * for floating comments.
+   */
+  private static final Token TOKENS_ACCEPT_EMPTY_CHILD[] = {
+    Token.SCRIPT, Token.MODULE_BODY, Token.BLOCK
+  };
+
+  private static final Set<Token> TOKENS_SET_ACCEPT_EMPTY_CHILD =
+      new HashSet<Token>(Arrays.asList(TOKENS_ACCEPT_EMPTY_CHILD));
```
bowenni: Done.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -180,6 +210,42 @@ private Node newFloatingCommentFromBuffer() {
       return c;
     }
 
+    /**
+     * If the parent node can survive code generation with an extra EMPTY node child, then outputs
```
bowenni: Done.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -180,6 +210,42 @@ private Node newFloatingCommentFromBuffer() {
...
+     * floating comments by attaching to a new EMPTY node before the current node. If the parent is
+     * sensitive about its children count/order, then try to put the comment in current node. If the
```
bowenni: Done.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
bowenni: Done.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
bowenni: GETPROP has only two children so second is last.
Changed to getSecondChild() so it's consistent.

src/main/java/com/google/javascript/gents/NodeComments.java
```diff
@@ -12,6 +12,13 @@ void putComment(Node n, String comment) {
     nodeToComment.put(n, comment);
   }
 
+  void addComment(Node n, String comment) {
```
bowenni: Consolidated.

src/test/java/com/google/javascript/gents/singleTests/comments.js
bowenni: The AST looks like this:
```
CALL 1 [length: 18]
    GETPROP 1 [length: 16]
        CALL 1 [length: 14] [free_call: 1]
            NAME a 1 [length: 1]
        STRING b 3 [length: 1]
```
So the next node (`STRING b 3 [length: 1]`) is in `GETPROP`.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -180,6 +199,43 @@ private Node newFloatingCommentFromBuffer() {
...
 
+    /**
+     * If the parent node can survive code generation with an extra EMPTY node child, then output
+     * floating comments by attaching to a new EMPTY node before the current node. Otherwise try to
+     * put the comment in current node. If the current node doesn't accept comments (for example a
+     * BLOCK node will ignore any comments mapped to it), then attach the comments to the parent
+     * node. The parent node might not be the best place to hold the comment but at least we don't
+     * lose the comment and we don't break code generation.
+     *
+     * @param n
```
dpurp: either document `n` or drop the `@param`

PR #495, opened 2017-04-13
Fixes NPEs for constructors with destructuring.

rkirov:
Constructors with destructuring cannot use the 'public' shorthand:

class C {
  a: string;
  constructor({a}: {a: string}) {
    this.a = a;
  }
}

Also fixes a bug where gents half-moved ctors with default args.  Since
CodeGenerator seems to avoid emitting 'public'/'private' keywords on
arguments with defaults. Back off and leave this transformation to
tslint.


dpurp:
I'm fine with removing this feature if it's covered by tslint, but please verify that tslint takes care of the inferred const case below; otherwise, we'll produce many more `AnyDuringMigration`.

Input:
```javascript
const Baz = goog.defineClass(null, {
  constructor(foo, bar) {
    /** @private @const */
    this.foo = foo;

    /** @private @const */
    this.bar = bar;
  }
}
```

Expected Output:
```typescript
class Baz {
  constructor(private readonly foo, private readonly bar) {}
}
```

PR #496, opened 2017-04-13
fix(gents): ignore @constructor annotations if inside goog.defineClass

bowenni:
Previously we get a nested class:
```javascript
class A {
  constructor: any = class {  };
}
```

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -77,7 +77,8 @@ public final class TypeConversionPass implements CompilerPass {
         case FUNCTION:
           bestJSDocInfo = NodeUtil.getBestJSDocInfo(n);
           if (bestJSDocInfo != null
-              && (bestJSDocInfo.isConstructor() || bestJSDocInfo.isInterface())) {
+              && bestJSDocInfo.isConstructorOrInterface()
+              && !isConstructorInGoogDefineClass(n)) {
             convertConstructorToClass(n, bestJSDocInfo);
           }
           break;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -418,6 +419,17 @@ void convertDefineClassToClass(Node n) {
...
+  boolean inGoogDefineClass(Node n) {
+    if (n == null) {
+      return false;
+    }
+
+    if (n.isCall() && "goog.defineClass".equals(n.getFirstChild().getQualifiedName())) {
+      return true;
+    }
+
+    return inGoogDefineClass(n.getParent());
```
rkirov: I would skip traversing up here. What if you have something deeply nested what kinda made sense (I am making up stuff, but you never know):

```
goog.defineClass(null, {
  foo: function() {
    callMyApiThatExpectsAnonymousClass(/** @constructor */ function () {});
});
```

I think this should be a bit more restrictive and it should also check that the prop name is 'constructor'.

src/test/java/com/google/javascript/gents/singleTests/classes.js
```diff
@@ -75,6 +75,17 @@ const F = goog.defineClass(null, {
...
 });
 
+/**
+ * goog.defineClass with @constructor annotation
+ */
+var GoogDefinedClassWithConstructorAnnotation = goog.defineClass(null, {
+  /**
+   * @constructor
+   */
+  constructor: function() {
```
rkirov: add a test with params, so that the ctor function is kept.

bowenni: Done.

src/main/java/com/google/javascript/gents/TypeConversionPass.java
bowenni: Done.

PR #500, opened 2017-04-24
Added support for inline JsDoc types.

### LINKED ISSUE #352: Support inline types
Currently gents supports the traditional `@param` and `@return` based Closure typing, but it should also support Closure's concise inline types.

Example:

``` javascript
function /** string */ f(/** number */ x) {return x + ' apples'}

var /** string */ x = 'fruit';
```

Reference:
https://github.com/google/closure-compiler/wiki/Annotating-Types#function-declarations

dzinxed:
Fixes #352.

Reference:
https://github.com/google/closure-compiler/wiki/Annotating-Types#function-declarations

One unrelated test (interface.ts) had to be fixed for this change to pass.
It put a wrong comment before this change, now it puts it (still wrongly)
elsewhere.

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
@@ -147,37 +147,42 @@ public final class TypeAnnotationPass implements CompilerPass {
             break;
           }
           if (NodeUtil.isNameDeclaration(parent)) { // Variable declaration
-            if (bestJSDocInfo != null) {
-              setTypeExpression(n, bestJSDocInfo.getType(), false);
-            }
+            maybeSetTypeExpression(n, bestJSDocInfo, false);
           } else if (parent.isParamList()) { // Function parameters
             JSDocInfo parentDocInfo = NodeUtil.getBestJSDocInfo(parent.getParent());
-            if (parentDocInfo == null) {
-              break;
-            }
-            JSTypeExpression parameterType = parentDocInfo.getParameterType(n.getString());
-            if (parameterType != null) {
-              Node attachTypeExpr = n;
-              // Modify the primary AST to represent a function parameter as a
-              // REST node, if the type indicates it is a rest parameter.
-              if (parameterType.getRoot().getToken() == Token.ELLIPSIS) {
-                attachTypeExpr = IR.rest(n.getString());
-                nodeComments.replaceWithComment(n, attachTypeExpr);
-                compiler.reportCodeChange();
+            if (parentDocInfo != null) {
+              JSTypeExpression parameterType = parentDocInfo.getParameterType(n.getString());
+              if (parameterType != null) {
+                // Parameter is declared using verbose @param syntax before the function definition.
+                Node attachTypeExpr = n;
+                // Modify the primary AST to represent a function parameter as a
+                // REST node, if the type indicates it is a rest parameter.
+                if (parameterType.getRoot().getToken() == Token.ELLIPSIS) {
+                  attachTypeExpr = IR.rest(n.getString());
+                  nodeComments.replaceWithComment(n, attachTypeExpr);
+                  compiler.reportCodeChange();
+                }
+                // Modify the AST to represent an optional parameter
+                if (parameterType.getRoot().getToken() == Token.EQUALS) {
+                  attachTypeExpr = IR.name(n.getString());
+                  attachTypeExpr.putBooleanProp(Node.OPT_ES6_TYPED, true);
+                  nodeComments.replaceWithComment(n, attachTypeExpr);
+                  compiler.reportCodeChange();
+                }
+                setTypeExpression(attachTypeExpr, parameterType, false);
+                break;
               }
-              // Modify the AST to represent an optional parameter
-              if (parameterType.getRoot().getToken() == Token.EQUALS) {
-                attachTypeExpr = IR.name(n.getString());
-                attachTypeExpr.putBooleanProp(Node.OPT_ES6_TYPED, true);
-                nodeComments.replaceWithComment(n, attachTypeExpr);
-                compiler.reportCodeChange();
-              }
-              setTypeExpression(attachTypeExpr, parameterType, false);
             }
+            // If we didn't break, then maybe the type is inlined just before the parameter?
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -147,37 +147,42 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-                nodeComments.replaceWithComment(n, attachTypeExpr);
-                compiler.reportCodeChange();
-              }
-              // Modify the AST to represent an optional parameter
-              if (parameterType.getRoot().getToken() == Token.EQUALS) {
-                attachTypeExpr = IR.name(n.getString());
-                attachTypeExpr.putBooleanProp(Node.OPT_ES6_TYPED, true);
-                nodeComments.replaceWithComment(n, attachTypeExpr);
-                compiler.reportCodeChange();
+            if (parentDocInfo != null) {
```
mprobst: prefer to always early return, i.e. `if (parentDocInfo == null) return;`, so we have less indentation and state to worry about below. AFAICT that won't make a difference to the code, will it?

src/test/java/com/google/javascript/gents/singleTests/interface.ts
```diff
@@ -23,9 +23,8 @@ class Y implements Interface2, Interface3 {
 }
 
 export interface StructuralInterface { bar(a: string): number; }
-
-/** !ns.StructuralInterface */
 const structInterfaceImpl: StructuralInterface = {
+  /** !ns.StructuralInterface */
```
mprobst: Is this an intentional change? It'd be nice if we could keep the comments on the variables. Maybe we can rather change the way how we find the comment instead of changing the way how we link the comments?

src/test/java/com/google/javascript/gents/singleTests/inline_types.js
```diff
@@ -5,6 +5,6 @@ var /** !Foo */ foo = {};
 
 const /** string */ x = 'fruit';
 
-// TODO(#352): Support inline typing for functions.
-// input: function /** string */ f(/** number */ x) {return x + ' apples'}
-// output: function f(x: number): string {return x + ' apples'}
+function /** string */ f(/** number */ x, z, /** string */ y) {
```
mprobst: What'd happen if we mix function JSDoc (with or without `@param`s) with inline comments? I think your code handles that nicely, but could you add a test for the situation?

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -147,37 +147,42 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+                  nodeComments.replaceWithComment(n, attachTypeExpr);
+                  compiler.reportCodeChange();
+                }
+                setTypeExpression(attachTypeExpr, parameterType, false);
+                break;
               }
-              setTypeExpression(attachTypeExpr, parameterType, false);
             }
+            // If we didn't break, then maybe the type is inlined just before the parameter?
+            maybeSetTypeExpression(n, bestJSDocInfo, false);
```
mprobst: It took me a bit to figure out what this does. Maybe it'd be more readable if we explicitly called `getType()` here, and check for `null`?

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -225,6 +230,13 @@ public void visit(NodeTraversal t, Node n, Node parent) {
           break;
       }
     }
+
+    private void maybeSetTypeExpression(
+        Node nodeToType, JSDocInfo jsDocInfo, boolean isReturnType) {
+      if (jsDocInfo != null) {
```
mprobst: It seems like most callsites do not need the nullability guard in this function. I think I'd prefer to repeat the `!= null` check if it's only two or three sites, it's a bit more explicit in what's going on.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
dzinxed: It actually made a difference here (see old line 177). I refactored the code to make this more explicit.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
dzinxed: I inlined this in places where removing comments (see my next commit) was unnecessary.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
dzinxed: Same as above.

src/test/java/com/google/javascript/gents/singleTests/interface.ts
dzinxed: This has actually been resolved in my next commit.

src/test/java/com/google/javascript/gents/singleTests/inline_types.js
dzinxed: Thank you for a good idea. Adding a second version uncovered a flaw in my solution. I pushed a fix.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -225,6 +211,57 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+      if (nodeComment == null) {
+        return;
+      }
+      nodeComment = nodeComment.replaceFirst("\\n?" + Pattern.quote(toRemove), "");
+      nodeComments.clearComment(commentNode);
+      nodeComments.putComment(commentNode, nodeComment);
+      compiler.reportCodeChange();
+    }
+
+    private boolean setFunctionParameterTypeFromParentDoc(Node n, Node parent) {
```
mprobst: nit: maybe document the return here? it's not obvious from the method name.

dzinxed: Done.

mprobst:
Thanks for the contribution!

PR #508, opened 2017-05-09
Emit inferred props on a goog.module object.

### LINKED ISSUE #506: const inferred from function calls are not properly emitted
For example:

```javascript
goog.module('a.b.c');

/**
 * @param {string} str 
 * @return {string}
 */
f = function(str) {
  return str;
};

const c = f('foo');

exports.c = c;
```

Does not emit the type of `c` properly.

rkirov:
Clutz walks all `compiler.getTopScope().getAllSymbols()`. It appears
that for module objects

```javascript
goog.module('a.b');
const c = 0;
exports.c = c;
```

Closure does not register a corresponding 'a.b.c' symbol, while

```javascript
goog.module('a.b');
/** @const */
const c = 0;
exports.c = c;
```

does produce the symbol in question. To fix the first situation,
we also traverse the unemitted props on the 'a.b' object.

Fixes #506


rkirov:
Yes, it doesn't break anyone (which is the only definition of 'works' :))

PR #510, opened 2017-05-11
Inline deprecated options.setIdeMode call.

rkirov:
For now replace with exact contents of setIdeMode:
https://github.com/google/closure-compiler/blob/7580f657e1d513efb06eb1ff9821e872a8b743c6/src/com/google/javascript/jscomp/CompilerOptions.java#L2044
except setAllowHotswapReplaceScript since both clutz and gents do not
support hotswapping code.

Surprisingly removing setAllowHotswapReplaceScript changes type
inference (what looks like for the better).


rkirov:
cc @MatrixFrog

PR #511, opened 2017-05-11
Remove continueAfterError flag.

rkirov:
Clutz and Gents are already pretty lenient about errors.

However, when a true error is encountered by closure, continuing after
it can result in an even more catastrophic internal compiler error.

For example, the following snippet uses ES7 proposed syntax
```
function f() {
    {...res}
    switch (0) {
        case undefined:
            return 0
        }
    }
```
without setContinueAfterErrors - a normal syntax error.
with setContinueAfterErrors - an internal compiler error.

/cc @MatrixFrog

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -391,9 +391,13 @@ class DeclarationGenerator {
       List<SourceFile> sourceFiles, List<SourceFile> externs, Depgraph depgraph)
       throws AssertionError {
     compiler.compile(externs, sourceFiles, opts.getCompilerOptions());
-    precomputeChildLists();
-    collectTypedefs();
-    String dts = produceDts(depgraph);
+    String dts = "";
+    // If there is an error top scope is null.
+    if (compiler.getTopScope() != null) {
+      precomputeChildLists();
+      collectTypedefs();
+      dts = produceDts(depgraph);
+    }
     errorManager.doGenerateReport();
     return dts;
   }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -391,9 +391,13 @@ String generateDeclarations(
       List<SourceFile> sourceFiles, List<SourceFile> externs, Depgraph depgraph)
       throws AssertionError {
     compiler.compile(externs, sourceFiles, opts.getCompilerOptions());
-    precomputeChildLists();
-    collectTypedefs();
-    String dts = produceDts(depgraph);
+    String dts = "";
```
mprobst: initialize to null? That's more idiomatic, I guess? Do we want to return an empty string or null from this for errors?

rkirov: Currently, it seems like we always write the file - errors or no errors. Does blaze rely on that behavior. I am fearful of breaking something if we don't write anything in presence of errors.

mprobst: It's fine to keep it as is, I think. blaze/bazel will error if you did not produce your output *and* didn't return with an error code. If you return with an error code, it doesn't matter whether you wrote or not.

rkirov:
merged

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: added a comment to try null and no emit on error.

PR #513, opened 2017-05-15
Add a test for inner typedef inside a goog.module.

rkirov:
Includes a usage file.


rkirov:
I was trying to debug an user issue but I couldn't find a good test for this simple case.

mprobst:
LGTM.

PR #517, opened 2017-05-23
Fix TTE generator code to work with instance methods on $q.

### LINKED ISSUE #505: clutz loses types for ES6 classes with goog.provides
The following code:

```javascript
goog.provide('a.u.C');
goog.scope(function() {
  class C {
    foo() { return 0;}
  }
  /**
   * @export
   */
  a.u.C = C;
});
```

Produces `var C: any` instead of keeping track of the `foo` method.

rkirov:
Upgrade to patterns in 1.6 angular typings, including a new TTE for
Promise.when.

Fixes #505

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2195,27 +2195,38 @@ class DeclarationGenerator {
             + classTemplatizedType
             + " ;";
       }
-      return null;
+      if (propName.equals("when")) {
+        return "when < RESULT, T > (value: T, successCallback: (promiseValue: T) => "
+            + classTemplatizedType
+            + "|RESULT, errorCallback: null | undefined | "
+            + " ((reason: any) => any), notifyCallback?: (state: any) => any): "
+            + classTemplatizedType
+            + ";";
+      }
+      String promiseName = className + ".Promise";
+      return getPromiseMethod(propName, promiseName);
     }
 
     private String getSignatureForStaticTTEFn(String propName, FunctionType ftype) {
       // If ftype is foo.bar.Promise.all, extract className as Promise.
       String className = getUnqualifiedName(getNamespace(ftype.getDisplayName()));
+      String maybePromiseMethod = getPromiseMethod(propName, className);
+      if (maybePromiseMethod != null) {
+        return "static " + maybePromiseMethod;
+      }
+      return null;
+    }
+
+    private String getPromiseMethod(String propName, String className) {
       switch (propName) {
         case "resolve":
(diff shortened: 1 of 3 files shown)


rkirov:
cc @mgr01

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2195,27 +2195,36 @@ private String getSignatureForInstanceTTEFn(
...
         case "race":
-          return "static race < T > (values : T [] ) : " + className + " < T > ;";
+          return "race < T > (values : T [] ) : " + className + " < T > ;";
         case "all":
-          return "static all(promises : "
-              + className
-              + " < any > [] ) : "
-              + className
-              + " < any [] > ;";
+          return "all(promises : " + className + " < any > [] ) : " + className + " < any [] > ;";
```
mprobst: Shouldn't this be `all(promises: Promise<T>[]): Promise<T[]>`, or can't TS figure that signature out? In any case, it should probably match the signature of the promises in the standard library and in the angular typings, if possible.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2195,27 +2195,36 @@ private String getSignatureForInstanceTTEFn(
             + classTemplatizedType
             + " ;";
       }
-      return null;
+      if (propName.equals("when")) {
+        return "when < RESULT, T > (value: T, successCallback: (promiseValue: T) => "
+            + classTemplatizedType
+            + "|RESULT, errorCallback: null | undefined | "
+            + " ((reason: any) => any), notifyCallback?: (state: any) => any): "
```
bowenni: `notifyCallback` -> `progressCallback`?

rkirov: I used the name in the angular.d.ts, which describes the same api but in TS.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: That was the old state of the angular.d.ts where it was copied from. Might as well update to the new version, hopefully doesn't make g3 syncing a pain.

rkirov: Ended up not so simple, because .all has to support also tuple types. If we are to do this correctly we have to emit *all* overrides for all from here:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/angular/index.d.ts#L1014

I am leaving this for another PR.

rkirov:
merged

PR #520, opened 2017-06-03
Remove @type annotation for multiple lines annotations.

bowenni:
`.` matches any character except a newline therefore multi-lines annotations wouldn't be removed.

### DIFF
--- a/src/test/java/com/google/javascript/gents/singleTests/comments.ts
+++ b/src/test/java/com/google/javascript/gents/singleTests/comments.ts
@@ -37,6 +37,13 @@ const xConst: number = 1;
 
 class X extends A {}
 
+class B {
+  twoLines: {param1: string | undefined} = {param1: 'param1'};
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/gents/singleTests/comments.ts
```diff
@@ -29,14 +29,16 @@ let foo = function(deleted: number, notdeleted: number): number {
 };
 
 // The following comments should be mostly deleted
-/**
```
rkirov: why remove this one?

bowenni: Multi-line `@type` annotation matched 
`
/**
  * @type {number}
  * @private {number}
  * @protected {number}
  * @param {number} foo description of foo
  * @public {number}
  * @package {number}
  */
`
as a whole. I'm excluding `@` in the body. I think it should resolve the problem?

PR #522, opened 2017-06-15
Ignore all intellij files

DrJKL:


DrJKL:
I signed it!

DrJKL:
@rkirov

rkirov:
finally read through - https://git-scm.com/docs/gitignore (wow its trickier) and I think this is correct. My mental model was a lot simpler :/

PR #526, opened 2017-06-23
fix: add Object to the list of aliased global types.

mprobst:


rkirov:
You have to also update - https://github.com/angular/clutz/blob/master/src/test/java/com/google/javascript/clutz/general_with_platform.d.ts

mprobst:
Yeah, I actually had to update quite a few things, we use Object in places. Done though.

PR #528, opened 2017-06-30
Move access to .getTypeRegistry after .compile.

rkirov:
In preparation for using options.setAssumeForwardDeclations(true),
one has to make sure that compiler.getTypeRegistry is never called
before compiler.compile(...,...,options). Otherwise, an internal
state is incorrectly built that uses the wrong value for
forwardDeclaredTypes.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -236,8 +236,8 @@ class DeclarationGenerator {
 
   private static final String MODULE_PREFIX = "module$exports$";
 
-  private final JSType UNKNOWN_TYPE;
-  private final JSType NUMBER_TYPE;
+  private JSType unknownType;
+  private JSType numberType;
 
   public static void main(String[] args) {
     Options options = null;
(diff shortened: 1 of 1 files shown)


rkirov:
Alternatively there is `compile.init`, but at this point all the APIs look equally unsafe to me :/

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -236,8 +236,8 @@
 
   private static final String MODULE_PREFIX = "module$exports$";
 
-  private final JSType UNKNOWN_TYPE;
-  private final JSType NUMBER_TYPE;
+  private JSType UNKNOWN_TYPE;
```
mprobst: while you're touching this, these should be `lowerCamelCase`

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -302,8 +302,9 @@ public static void main(String[] args) {
         new ClutzErrorManager(
             System.err, ErrorFormat.MULTILINE.toFormatter(compiler, true), opts.debug);
     compiler.setErrorManager(errorManager);
-    UNKNOWN_TYPE = compiler.getTypeRegistry().getNativeType(JSTypeNative.UNKNOWN_TYPE);
-    NUMBER_TYPE = compiler.getTypeRegistry().getNativeType(JSTypeNative.NUMBER_TYPE);
+    // Calling compiler.getTypeRegistry() is not safe to call here,
```
mprobst: nit: Calling ... to call

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -302,8 +302,9 @@ public static void main(String[] args) {
...
+    // because it initialized some internal compiler structures before
```
mprobst: initialize_s_

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: dibe

Issue #483, opened 2017-03-29
gents does not support generics


rkirov:
The following code 

```javascript
/**
 * @constructor
 * @final
 * @template T
 */
goog.structs.AvlTree = function () {...}
```

gets transformed to

```
class AvlTree {...}
```

Most likely this is because closure's AST does not support this, but maybe some post processing hack can be applied (at least in this simple case).

rkirov:
This was fixed with https://github.com/angular/clutz/commit/5b4ec54f298993bfd43e59daf05ee5ab655cd297

PR #529, opened 2017-06-30
Keep generic @extends comments to make post processing easier.

stringham:
This may help with implementing https://github.com/angular/clutz/issues/483, and in the meantime it helps you identify places that need to be manually fixed.

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -33,8 +33,11 @@ public final class CommentLinkingPass implements CompilerPass {
    */
   private static final Pattern[] JSDOC_REPLACEMENTS = {
     Pattern.compile(
-        BEGIN_JSDOC_LINE + "@(extends|implements|type)[ \t]*(\\{[^@]*\\})[ \t]*(?<keep>)" + EOL,
+        BEGIN_JSDOC_LINE + "@(implements|type)[ \t]*(\\{[^@]*\\})[ \t]*(?<keep>)" + EOL,
         Pattern.DOTALL),
+    //keep generic extends for post processing
+    Pattern.compile(
+        BEGIN_JSDOC_LINE + "@extends[ \t]*(\\{[^@<>]*\\})[ \t]*(?<keep>)" + EOL, Pattern.DOTALL),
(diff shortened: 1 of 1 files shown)


stringham:
I signed it!

rkirov:
Design-wise, this will make tsickle reject the output due to:

https://github.com/angular/tsickle/blob/master/src/jsdoc.ts#L100

But maybe it is desirable because then the user will go and manually fix the type. @evmar WDYT?

Process-wise, please add a test for this in comments.js / .ts.

evmar:
(I don't have much context, but:)  If you can translate @extends to TS extends then it's better to translate, but if you can't, then probably better to preserve the comment.

stringham:
The idea here is based on #483 where @rkirov said:

>Most likely this is because closure's AST does not support this, but maybe some post processing hack can be applied (at least in this simple case).

It's already creating code that won't compile properly unless it's fixed, and this will help us identify the places that need `<T>` generic braces added.

I agree that it would be ideal to translate the @extends, and clutz seems to be able to gather that information. However clutz does this by getting all of the symbols, and I'm not sure we have access to the symbols when generating typescript in gents.

For some reason, the type information is available when declaring a member type:

```typescript
class MyClass {
    private avlTree: AvlTree<number>;
}
```

comes through okay.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -33,8 +33,11 @@
    */
   private static final Pattern[] JSDOC_REPLACEMENTS = {
     Pattern.compile(
-        BEGIN_JSDOC_LINE + "@(extends|implements|type)[ \t]*(\\{[^@]*\\})[ \t]*(?<keep>)" + EOL,
+        BEGIN_JSDOC_LINE + "@(implements|type)[ \t]*(\\{[^@]*\\})[ \t]*(?<keep>)" + EOL,
         Pattern.DOTALL),
+    //keep generic extends for post processing
```
rkirov: add a space after '//' and '// TODO: investigate whether generics params can be emitted correctly'.

rkirov:
Yep, I am convinced since we already do the wrong thing, it would be better to keep enough information in the output .ts so that once can at least manually get to the correct state without having to go back to the original .js.

rkirov:
This PR is too stale at this point. Thanks for contributing back, we are still doing many migrations in google and will reopen if this appears again.

Issue #485, opened 2017-03-29
gents doesn't convert static fields


bowenni:
```javascript
/**
 * @constructor
 */
function A() {};
A.b = {};
```
is converted to 
```javascript
class A {}
A.b = {};
```

It should be converted to
```javascript
class A {
  static b = {};
}
```

PR #530, opened 2017-06-30
Always move field declarations into classes as static methods, even if there is no type information

stringham:
This fixes https://github.com/angular/clutz/issues/485

This makes the check for adding static members more specific. Any field that is not annotated with `@enum` and that is not a class will be added as a static member.

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -208,8 +208,10 @@ public final class TypeConversionPass implements CompilerPass {
           if (declaration.rhs != null && declaration.rhs.isFunction()) {
             moveMethodsIntoClasses(declaration);
           } else {
-            // Ignore field declarations without a type annotation
-            if (declaration.jsDoc != null && declaration.jsDoc.getType() != null) {
+            // Ignore field declarations that are enums or classes
+            String comment = nodeComments.getComment(n);
+            if (!(declaration.rhs != null && declaration.rhs.isClass())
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -208,8 +208,10 @@ public void visit(NodeTraversal t, Node n, Node parent) {
           if (declaration.rhs != null && declaration.rhs.isFunction()) {
             moveMethodsIntoClasses(declaration);
           } else {
-            // Ignore field declarations without a type annotation
-            if (declaration.jsDoc != null && declaration.jsDoc.getType() != null) {
+            // Ignore field declarations that are enums or classes
+            String comment = nodeComments.getComment(n);
+            if (!(declaration.rhs != null && declaration.rhs.isClass())
+                && (comment == null || comment.indexOf("@enum") < 0)) {
```
rkirov: Rather avoid ad-hoc searching for '@enum' here. I think `NodeUtil.getBestJSDocInfo(n).hasEnumParameterType()` should work here? Which reminds me, do we have a test for enums statics?

rkirov:
Looks good, just a few nits.

Also PR description should read 'move static field declarations into classes'. I am a bit stickler on nomenclature, but in my minds `C.foo` and `class C { static foo }` a just two ways of declaring static fields (closure promoted the first one and TS uses the second one).

rkirov:
This PR is too stale at this point. Thanks for contributing back, we are still doing many migrations in google and will reopen if this appears again.

PR #531, opened 2017-06-30
Switch to accepting ECMASCRIPT_NEXT

rkirov:
Adds a test for async/await code.


src/test/java/com/google/javascript/clutz/async_await_with_platform.d.ts
```diff
@@ -0,0 +1,16 @@
+declare namespace ಠ_ಠ.clutz {
+  class module$exports$asyncawait extends module$exports$asyncawait_Instance {
+  }
+  class module$exports$asyncawait_Instance {
+    private noStructuralTyping_: any;
+    bar ( ) : any ;
```
rkirov: Obviously not ideal that these are 'any's, but not sure how to change this.

PR #532, opened 2017-06-30
Reduce the list of reserved words.

rkirov:
It appears we were too restrictive, for some of these words,
TypeScript is ok with 'namespace foo.<Word> {}'.

For example, 'namespace foo.interface {}' is ok.

Moreover, the fix for encountering reserved words is lossy, i.e.
misses more complex types, so we rather use it only when forced.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -165,11 +165,18 @@ class DeclarationGenerator {
           "whatwg_encoding.js",
           "window.js");
 
-  /** Reserved words in ES6.1. Thanks to https://mathiasbynens.be/notes/reserved-keywords. */
+  /**
+   * Words for which we cannot generate 'namespace foo.Word {}'; Instead clutz tries to roll them up
+   * into properties of the parent namespace, but that can be lossy.
+   *
+   * <p>To regenerate this list use the following script with a list of candidates.
+   *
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -182,28 +189,19 @@
...
           "finally",
           "for",
           "function",
           "if",
-          "implements",
           "import",
           "in",
           "instanceof",
-          "interface",
-          "let",
```
rkirov: let is ok, but not const?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -165,11 +165,17 @@
...
           "window.js");
 
-  /** Reserved words in ES6.1. Thanks to https://mathiasbynens.be/notes/reserved-keywords. */
+  /**
+   * Words for which we cannot generate 'namespace foo.Word {}'; Instead clutz tries to roll them up
+   * into properties of the parent namespace, but that can be lossy.
+   *
+   * <p>To regenerate this list use the following script with a list of candidates.
+   *
+   * <p>> for k in $(cat /tmp/keywords); do echo "namespace foo.$k {};" > /tmp/test.ts; \ > tsc
```
evmar: If this comment is HTML i think your ">" prompts will be eaten, maybe switch to $?  (I also was confused by the random ">" in the middle of the command)

rkirov: gah, sorry, I ran the auto-formatter and didn't look at the output.

PR #545, opened 2017-08-16
Add option for supporting partial program input to clutz.

rkirov:
When the flag is set, clutz will use setAssumeForwardDeclaration for all
unknown types, and not error on incomplete inputs.

Adds /partial/ directory for tests that run with that option on.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -71,6 +71,14 @@ public class Options {
   )
   List<String> entryPoints = new ArrayList<>();
 
+  @Option(
+    name = "--partialInput",
+    usage =
+        "allow input of incomplete programs. All unknown types will be treated as forward"
+            + "declared."
+  )
+  boolean partialInput;
+
   @Argument List<String> arguments = new ArrayList<>();
 
   Depgraph depgraph;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -105,7 +109,7 @@ StringSubject diagnosticStream() {
       roots.add("main.js");
     }
 
-    List<SourceFile> externFiles = new ArrayList<>();
+    List<SourceFile> externFiles;
```
evmar: This seems unrelated (?)

Harmless if so, but weird if unintentional

src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -71,6 +71,14 @@
   )
   List<String> entryPoints = new ArrayList<>();
 
+  @Option(
+    name = "--partialInput",
+    usage =
+        "allow input of incomplete programs. All unknown types will be treated as forward"
+            + "declared."
```
mprobst: nit: you're missing a space here in the string.

src/test/java/com/google/javascript/clutz/partial/missing_type.d.ts
```diff
@@ -0,0 +1,10 @@
+declare namespace ಠ_ಠ.clutz.module$exports$missing$type {
+  var x : any ;
```
mprobst: This should end up being `var x : Missing ;` eventually, but needs further changes, right?

src/test/java/com/google/javascript/clutz/ProgramSubject.java
rkirov: yep, unrelated clean-up. Intellij made is different color, so hard to resists :)

src/test/java/com/google/javascript/clutz/partial/missing_type.d.ts
rkirov: yep, one PR at a time. I am unrolling a bunch of local changes I have internally. That would be the next PR.

Issue #540, opened 2017-08-03
Static methods on @enum {number} get dropped


takamori:
If you have the following code, the static method `my.Fruit.getDebugColor` gets dropped in the typings file output by clutz, which results in an error when trying to use it from TypeScript code.
```
/** @enum {number} */
my.Fruit = {
  APPLE: 0,
  PEAR: 1
};

/**
 * @param {my.Fruit} fruit
 * @return {string} A color for debugging.
 */
my.Fruit.getDebugColor = function(fruit) {
  switch (fruit) {
    case my.Fruit.APPLE: return 'red';
    case my.Fruit.PEAR: return 'green';
    default: return 'invisible';
  }
};
```

It would be helpful to address this case somehow for users who are converting to TypeScript.

mprobst:
Given that we're moving closer to use proper TypeScript enums, for the type checking benefits, I think this is infeasible as long as TS doesn't support methods on enums. See #546.

rkirov:
Once we move to proper TS enums, we can emit an enum + namespace pair like this:

```typescript
declare enum X {
  A,
  B
}
declare namespace X {
  function foo(): void;
}
```
But it is not of high priority at this point. Will reopen the bug if we end up working on it.

PR #547, opened 2017-08-18
Special case `@enum {number}` enums.

### LINKED ISSUE #546: Special case `number` enums to use TS enums
Clutz currently emits enums as a combination of type alias and object listing the values:

```ts
type Fruit = number /* or whatever */;
var Fruit = {APPLE: Fruit, BANANA: Fruit};
```

That's because Closure enums in the general case can also be arbitrary values, not just numbers (+ strings as in TS2.4). This works reasonably well, but removes some of the advantages of enums, e.g. assignment incompatibility between the underlying type and the enum, and exhaustiveness checks in the type system.

We could fix this for a large set of use cases by special casing number. We should also emit `&{brand: never}` to achieve assignment incompatibility.

mprobst:
Emitting proper TypeScript enums allows the TypeScript compiler to make
additional inferences about the code, e.g. to support tagged unions,
type narrowing, exhaustive switches etc.

This also adds an `&{brand: never}` token to all other enums, to prevent
accidental assignment compatibility of values.

Because enums with the same members are not assignment compatible in TS
(enums are essentially nominally typed), this also has to update the
aliased enum emit.

Fixes #546.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1205,7 +1205,7 @@ class DeclarationGenerator {
       } else {
         maybeEmitJsDoc(symbol.getJSDocInfo(), /* ignoreParams */ false);
         if (type.isEnumType()) {
-          visitEnumType(emitName, (EnumType) type);
+          visitEnumType(emitName, emitName, (EnumType) type);
           return;
         }
         if (isTypedef(type)) {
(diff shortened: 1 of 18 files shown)


mprobst:
Running this through global presubmit, as the assignability and enum aliasing changes might yield some surprises...

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1475,46 +1475,87 @@ private String normalizeIObjectTemplateName(ObjectType type, String displayName)
...
-      for (String elem : sorted(type.getElements())) {
-        emit(elem);
-        emit(":");
-        // No need to use type.getMembersType(), this must match the type alias we just declared.
+      String unqualifiedName = getUnqualifiedName(symbolName);
+
+      // @enums can be aliased by assignment. Emit a type alias + value alias for the situation.
+      String elementsTypeName = type.getElementsType().getReferenceName();
+      if (!qualifiedName.equals(elementsTypeName)) {
+        emitComment(symbolName + " aliases enum " + elementsTypeName);
```
rkirov: I don't see 'aliases enum' in any of the tests below. We are either missing a test coverage here or there is a bug.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1475,46 +1475,87 @@ private String normalizeIObjectTemplateName(ObjectType type, String displayName)
...
+        for (String elem : sorted(type.getElements())) {
+          emit(elem);
+          emit(",");
+          emitBreak();
+        }
+        unindent();
+        emit("}");
+        emitBreak();
+      } else {
+        visitTypeAlias(type.getElementsType().getPrimitiveType(), unqualifiedName, true);
```
rkirov: feels like a bunch of duplicated code, but I am sure you considered abstracting some of it into helpers.

rkirov:
Also https://github.com/angular/clutz/issues/540 reminds me that we need a test for enums with methods, even if we are to keep the current behavior of not emitting them, so that we can be sure your change does not inadvertently produce something non-sensical.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: We filter the `// !!` comments before doing test comparisons, to keep test files less noisy. We should probably not do that for comments we emit for the user to see (as opposed to comments we put into the test files). Maybe that's all silly, but it's a change for a different CL :-)

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: I tried, but it turns out there's enough subtle variance between the emits that it's not worth it. E.g. we emit `enum Foo {` vs `var Foo = {`, then each element gets a type annotation or not, need a trailing semi or not, etc.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1475,46 +1475,87 @@ private String normalizeIObjectTemplateName(ObjectType type, String displayName)
...
 
-    private void visitTypeAlias(JSType registryType, String unqualifiedName) {
+    private void visitTypeAlias(
+        JSType registryType, String unqualifiedName, boolean emitNeverBrand) {
       emit("type");
       emit(unqualifiedName);
       emit("=");
       visitType(registryType, true, false);
+      // emit a brand to prevent accidental compatibility of values with an enum.
+      if (emitNeverBrand) emit("&{brand: never}");
```

mprobst:
PTAL.

mprobst:
Now at one or two CLs left to fix internal usage...

PR #548, opened 2017-08-24
Support custom [Symbol.iterator] implementations.

mprobst:
While Closure Compiler accepts code iterating types that implement
`[Symbol.iterator]`, it does not have a mechanism to represent computed
properties such as `Symbol.iterator`, so Clutz cannot really emit the
property itself.

As a workaround, simply emit a `[Symbol.iterator]` property for all
types that extend the well-known `Iterable` class, and special case its
return value for the well-known `IteratorIterable` class (called
`IteratorIterable` in Closure).

This also requires not emitting Iterable and IteratorIterable types, and
renaming the involved interfaces, such as `IteratorResult`.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -47,6 +47,7 @@ import com.google.javascript.rhino.jstype.ObjectType;
 import com.google.javascript.rhino.jstype.ProxyObjectType;
 import com.google.javascript.rhino.jstype.RecordType;
 import com.google.javascript.rhino.jstype.TemplateType;
+import com.google.javascript.rhino.jstype.TemplateTypeMap;
 import com.google.javascript.rhino.jstype.TemplatizedType;
 import com.google.javascript.rhino.jstype.UnionType;
 import com.google.javascript.rhino.jstype.Visitor;
@@ -124,8 +125,6 @@ class DeclarationGenerator {
           "RefererrerPolicy",
           "ResponseInit",
           "ReferrerPolicy",
-          "Iterable",
-          "IteratorIterable",
           "WritableStreamSink",
           "PipeOptions",
           "WritableStreamDefaultWriter",
@@ -235,6 +234,8 @@ class DeclarationGenerator {
 
   private JSType unknownType;
   private JSType numberType;
+  @Nullable private JSType iterableType;
+  @Nullable private JSType iteratorIterableType;
 
   public static void main(String[] args) {
     Options options = null;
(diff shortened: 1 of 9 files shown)


mprobst:
g3 is green before the special case CL, running another presubmit with it.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2038,6 +2051,47 @@ private void visitProperties(
...
+      if (iteratorIterableType != null && instanceType.isSubtype(iteratorIterableType)) {
+        implemented = iteratorIterableType;
+        returnType = "IterableIterator";
+      } else if (iterableType != null && instanceType.isSubtype(iterableType)) {
+        implemented = iterableType;
+        returnType = "Iterator";
+      } else {
+        return;
+      }
+      emit("// Symbol.iterator inserted by Clutz for Iterable subtype");
```
rkirov: emitComment?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2136,6 +2194,46 @@ private void emitProperty(
     }
 
     /**
+     * Closure Compiler does not support tuple types, and thus cannot express the proper type for
+     * "Map<K, V>.entries()", which is an IterableIterator of a [K, V] tuple.
```
rkirov: Not clear from the comment why the wider type `(K|V)[]` is not used. It is simply that [K, V] is 'more correct' or  destructuring is not accepted by TS.

src/test/java/com/google/javascript/clutz/map_entries_with_platform.js
```diff
@@ -0,0 +1,10 @@
+goog.provide('map_entries.Map');
```
rkirov: I try to make all new tests use goog.module as a more realistic to how new JS code looks like. Just a suggestion, not a hard rule.

rkirov:
The commit message says ' the well-known IteratorIterable class (called
IteratorIterable in Closure).' one of those should be `IterableIterator`.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Done.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Expanded the comment.

src/test/java/com/google/javascript/clutz/map_entries_with_platform.js
mprobst: Done.

mprobst:
Done re commit message.

mprobst:
g3 green, merged.

PR #550, opened 2017-08-31
Emit implicitly forward declared types for partialInputs.

rkirov:
Adds supports for generic types. For now exclude 'partial' tests from
usage and syntax tests, so that we do not need to provide all
missing types.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -42,6 +42,7 @@ import com.google.javascript.rhino.jstype.JSType;
 import com.google.javascript.rhino.jstype.JSTypeNative;
 import com.google.javascript.rhino.jstype.JSTypeRegistry;
 import com.google.javascript.rhino.jstype.NamedType;
+import com.google.javascript.rhino.jstype.NoResolvedType;
 import com.google.javascript.rhino.jstype.NoType;
 import com.google.javascript.rhino.jstype.ObjectType;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1594,6 +1595,18 @@ public Void caseTemplateType(TemplateType templateType) {
 
             @Override
             public Void caseNoType(NoType type) {
+              if (type.isNoResolvedType() && opts.partialInput) {
```
mprobst: Maybe add an implementation comment pointing out what's going on in here?

mprobst: Invert the condition and early exit? `if (!noResolved || !opts..) { emit(any); return; } ...`

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
```diff
@@ -102,6 +102,11 @@ static String getExternFileNameOrNull(String testFileName) {
     return filesList;
   }
 
+  public static List<File> getTestInputFilesNoPartial(FilenameFilter filter) {
+    File[] testFiles = getPackagePath().toFile().listFiles(filter);
+    return Lists.newArrayList(testFiles);
```
mprobst: `Arrays.asList(testFiles)` saves copying the array

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #553, opened 2017-09-06
Add support for emitting extending/implementing an unresolved base.

rkirov:
Adds an empty ClutzMissingBase class to closure.lib.d.ts for the case
that closure fails to resolve the base type. I could not reproduce this
code path in test, but have seen it in the wild. Closure errs on simple
attempts to extend a non-resolved base symbol and in partial compilation
we can use the forward declared name.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1542,6 +1542,27 @@ private void visitTypeDeclaration(JSType type, boolean isVarArgs, boolean isOpti
       }
     }
 
+    private boolean emitNoResolvedType(NoType type, String defaultEmit) {
```
mprobst: `maybeEmitNoResolvedType`?

mprobst: add some docs on what this does? It's not obvious from the signature.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1542,6 +1542,27 @@ private void visitTypeDeclaration(JSType type, boolean isVarArgs, boolean isOpti
...
+      // for which we just emit the literal type written along with any type parameters.
+      NoResolvedType nType = (NoResolvedType) type;
+      if (nType.getReferenceName() != null) {
+        String displayName = type.getDisplayName();
+        emit(Constants.INTERNAL_NAMESPACE + "." + displayName);
+        List<JSType> templateTypes = nType.getTemplateTypes();
+        if (templateTypes != null && templateTypes.size() > 0) {
+          emitGenericTypeArguments(type.getTemplateTypes().iterator());
+        }
+      } else {
```
mprobst: I think I'd invert the if and do another early return here.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1542,6 +1542,27 @@ private void visitTypeDeclaration(JSType type, boolean isVarArgs, boolean isOpti
...
+        emit(defaultEmit);
```
mprobst: can you document when/why this happens, or is that part of the mystery?

src/test/java/com/google/javascript/clutz/partial/missing_base.js
```diff
@@ -0,0 +1,33 @@
+goog.module('mising.extend');
+
+//!! Using class syntax requires default externs but the test runs faster
+//!! without them. Just use @constructor instead.
+
+/**
+ * @constructor
+ * @extends {direct.ref.A}
```
mprobst: is the "direct.ref" bit significant?

src/test/java/com/google/javascript/clutz/partial/missing_base.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+function B() {
+
+}
+
+/**
+ * @constructor
+ * @extends {direct.ref.AT<string, number>}
+ */
+function BT() {
```
mprobst: The naming here is not super obvious for me – maybe `MissingParentClassTemplated` would be clearer than `BT`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: This code is shared between the type emitter and the extends/implements emitter. It can easily happen for types , when one has type inference involved. We have tests for that. However, for extends/implements I cannot reproduce it with test (i.e. `ClutzBaseClass` doesn't appear anywhere).

Added a todo

src/test/java/com/google/javascript/clutz/partial/missing_base.js
rkirov: There was a way to just get a handle of `A` instead of the qualified namespace `direct.ref.A`, at least once I had that issue, and eversince got the habit of making all type qualified. Doesn't hurt.

src/test/java/com/google/javascript/clutz/partial/missing_base.js
rkirov: renamed to BTemplated

PR #554, opened 2017-09-07
Emit missing types as template parameters.

rkirov:
For unknown to me reason those end up as NamedType instead of NoType,
thus they need special handling logic.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1647,7 +1647,22 @@ class DeclarationGenerator {
 
             @Override
             public Void caseNamedType(NamedType type) {
-              visitType(type.getReferencedType());
+              JSType refType = type.getReferencedType();
+              // When we dealing with partial inputs, we often end up with named type, without a
+              // corresponding referencedType. Instead of emitting 'any', we emit literally the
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1647,7 +1647,22 @@ public Void caseUnionType(UnionType type) {
...
             @Override
             public Void caseNamedType(NamedType type) {
-              visitType(type.getReferencedType());
+              JSType refType = type.getReferencedType();
+              // When we dealing with partial inputs, we often end up with named type, without a
+              // corresponding referencedType. Instead of emitting 'any', we emit literally the
+              // name of the type as originally written.
+              // It appears that when one writes '@type {A<B>}' and both are missing from the compilation
+              // unit - A ends up as NoType, while B ends up as NamedType.
+              if (opts.partialInput && refType == unknownType) {
```
mprobst: should this be `refType.isUnknownType()`? I think the `isXXX` methods are preferable over comparing with a singleton instance, in case at some point they refactor the instance to no longer be singleton (as I did with `NoResolvedType`)

rkirov: Done. Works like a charm, I should just remove the field `unknownType` in another cl.

PR #555, opened 2017-09-07
Emit a default type 'any' for all type parameters when partial input.

rkirov:
Because closure allows generic types to be used without passing any
type arguments, incremental clutz cannot differentiate between
Foo a plain type or Foo a generic type that has implicilty '?' as
a type argument.

However, it is a syntactic mistake in TypeScript to use the generic type
without any type arguments if they do not have default values. To work
around this issue, in incremental mode, clutz will now emit '= any' for
all type parameters.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1262,7 +1262,7 @@ class DeclarationGenerator {
       visitTemplateTypes(ftype);
       emit("=");
       emit(typeName);
-      visitTemplateTypes(ftype);
+      visitTemplateTypes(ftype, Collections.emptySet(), false);
       emit(";");
       emitBreak();
       if (!ftype.isInterface()) {
@@ -1325,7 +1325,7 @@ class DeclarationGenerator {
         visitTemplateTypes(ftype);
         emit("extends");
         emit(name + INSTANCE_CLASS_SUFFIX);
-        visitTemplateTypes(ftype);
+        visitTemplateTypes(ftype, Collections.emptySet(), false);
         emit("{");
         indent();
         emitBreak();
@@ -1398,10 +1398,22 @@ class DeclarationGenerator {
     }
 
     private void visitTemplateTypes(ObjectType type) {
-      visitTemplateTypes(type, Collections.<String>emptySet());
+      visitTemplateTypes(type, Collections.<String>emptySet(), true);
     }
 
-    private void visitTemplateTypes(ObjectType type, Set<String> alreadyEmittedTemplateType) {
+    /**
+     * Emits template types for a given type. For example for <code>T<A,B></code>, this method will
+     * emit <code><A,B></code>.
+     *
+     * @param type the type in question
+     * @param alreadyEmittedTemplateType when visiting methods the class template types will be
+     *     reported by closure, but should not be emitted.
+     * @param isDeclaration isDeclaration {@pre true} if this type declares the template types,
+     *     {@pre false} if it instantiates a generic type. In the former case, Clutz emits defaults
+     *     for the template parameters.
+     */
+    private void visitTemplateTypes(
+        ObjectType type, Set<String> alreadyEmittedTemplateType, boolean isDeclaration) {
       if (type.hasAnyTemplateTypes() && !type.getTemplateTypeMap().isEmpty()) {
         List<String> realTemplateType = new ArrayList<>();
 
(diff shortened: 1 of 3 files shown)


rkirov:
Tried to keep argument/parameter distinction in mind here (https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter), but you will be the final judge whether I succeeded.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1398,10 +1398,31 @@ private void visitVarDeclaration(String name, JSType type) {
...
+      visitTemplateTypes(type, Collections.<String>emptySet(), true);
     }
 
-    private void visitTemplateTypes(ObjectType type, Set<String> alreadyEmittedTemplateType) {
+    /**
+     * Emits template types for a given type. For example for
+     *
+     * <pre>T<A,B></pre>
+     *
+     * , this method will emit
```
mprobst: wow, that's a very ugly formatting google-java-format chose there. But maybe it's wrapping `<pre>` because that's a block element, and this would look better if you used `<code>`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1411,14 +1432,21 @@ private void visitTemplateTypes(ObjectType type, Set<String> alreadyEmittedTempl
...
-          // To perserve the sementics from closure we skip emitting redeclared variables.
+          // To preserve the semantics from closure we skip emitting redeclared variables.
           if (alreadyEmittedTemplateType.contains(displayName)) {
             continue;
           }
 
           if (displayName.contains("IObject#")) {
             displayName = normalizeIObjectTemplateName(type, displayName);
           }
+          // When we emit partial programs, we cannot differentiate whether Foo
```
mprobst: This is a good fix, but I think it might be even more generic. AFAIU generally the semantics of Closure are that you can leave out generic type arguments, and it'll simply assume an unknown type (~= any) in its place. So this is correct under all circumstances, not just for partialInput, isn't it?

That'd allow us to drop the logic that inserts `<any>` for missing template parameters in the other place.

src/test/java/com/google/javascript/clutz/partial/missing_generic.js
```diff
@@ -0,0 +1,18 @@
+goog.module('missing.gen');
+
+//!! When emitting with missing sources, we have to emit T = any, to
+//!! support downsteam consumers that use missing.gen.C without passing
+//!! any type parameters.
+/**
+ * @constructor
+ * @template T
+ */
+function C(){}
```
mprobst: I'd really, really prefer to name these after the case they are testing. `C` says nothing at all, `GenericClassDefaultsArgumentsToAny` or even just `GenericClass` would make this so much more readable. You're not calling all the variables in your source code `a`, `b`, `c`, etc either.

I've been guilty of this (I think I introduced this pattern in the first place), but after debugging a bunch of test cases, I've really come around on it.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1398,10 +1398,31 @@ private void visitVarDeclaration(String name, JSType type) {
...
+     * <pre>T<A,B></pre>
+     *
+     * , this method will emit
+     *
+     * <pre><A,B></pre>
+     *
+     * .
+     *
+     * @param type the type in question
+     * @param alreadyEmittedTemplateType in certain scopes some template types will be already
```
mprobst: maybe you could mention that this is for `@template` on methods that repeat an `@template` of the class (is it?)

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1398,10 +1398,31 @@ private void visitVarDeclaration(String name, JSType type) {
...
+     *     emitted and have to be skipped.
+     * @param isDeclaration this method will be called when defining new template types, like
```
mprobst: `this method ...` is a bit misleading as the doc string start, isn't it? maybe `isDeclaration {@pre true} if this type declares the template types, {@pre false} if it instantiates a generic type. In the former case, Clutz emits defaults for the template parameters.`

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Actually, thinking about it a bit more, I'm no longer that sure. Using a class but forgetting its generic argument and then getting any is a bit sad - you lose a lot of type safety, and it's very easy to do so unintentionally.

I wonder if we can default to `{}` (empty type) instead of `any`?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: much better.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Ack, I think this is fine.

PR #556, opened 2017-09-08
Generate the right destructuring call for imports that are kept in js

iteriani:


iteriani:
Oops I added myself as a googler

iteriani:
Confirmed that it's ok since it's me.

Issue #537, opened 2017-07-21
gents does not generate the right destructuring call for imports that are kept in js


rkirov:
Simple repro:

```javascript
// a.js
const {X} = goog.require('b');
const a = X.A;
```

and 

```javascript
// b_keep.js
goog.module('b');

/** @enum {number} */
exports.X = {A: 0};
```

produces ```import X from 'goog:b.X';``` in `a.ts` instead of ```import {X} from 'goog:b';```

bowenni:
A variant of this:
```javascript
// a.js
goog.require('foo.bar');
foo.bar.baz();
```
and
```javascript
// b_keep.js
goog.provide('foo.bar');
foo.bar.baz = function() {
  ...
}
```
produces
```typescript
// a.ts
import bar from 'goog:foo.bar';
```

A better output will be
```typescript
// a.ts
import {baz} from 'goog:foo.bar';
baz();
```
or maybe
```typescript
import * as bar from 'goog:foo.bar';
bar.baz();
```

bowenni:
This is partially solved in https://github.com/angular/clutz/pull/569
The destructuring imports case is not handled yet.

PR #557, opened 2017-09-09
Add test cases to show incorrect behavior for destructuring closure code.

iteriani:
The correct behavior in the translated TS code should be:

import {R} from 'goog:module';
import {foobar} from 'goog:provides';

instead of
import R from 'goog:module.R';
import foobar from 'goog:provides';

### DIFF
--- a/src/test/java/com/google/javascript/gents/multiTests/module_imports/import_binding.ts
+++ b/src/test/java/com/google/javascript/gents/multiTests/module_imports/import_binding.ts
@@ -6,12 +6,12 @@ import * as namespace from './export_namespace';
 import {D} from './export_both';
 import * as DExports from './export_both';
 import {bar} from './export_both';
-import foobar from 'goog:provide';
-import {R} from 'goog:provide.R';
-
+import R from 'goog:module.R';
+import foobar from 'goog:provides';
 
 // Use code in JS.
-const baz = foobar.A;
+R();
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/gents/multiTests/module_imports/import_binding.ts
```diff
@@ -6,6 +6,10 @@ import * as namespace from './export_namespace';
 import {D} from './export_both';
 import * as DExports from './export_both';
 import {bar} from './export_both';
+import R from 'goog:module.R';
```
mprobst: this is a bit misleading - if you don't `declareLegacyNamespace`, `"module".R` is completely different from `"module.R"`. That is, this code should fail, there shouldn't be anything providing `goog:module.R`.

Is that what you're after here? Also, the test doesn't seem to fail..

iteriani: This illustrates the incorrect behavior, which should be 
import {R} from 'goog:module';

when fixed.

rkirov:
This was fixed with https://github.com/angular/clutz/issues/537. Please reopen if it is not quite there. Bowen improved the tests in MultiTest/jslib_imports/* so you can use those in the future to create repro's for importing from code that stays behind in javascript.

PR #560, opened 2017-09-12
fix NPE that occurs when a missing type is shadowed by a known namesp…

rkirov:
…ace.

In general, gents is designed to be permissive for missing types and
tries to emit them as syntactically seen. However, that requires care
further along to make sure that appearing 'null's are handled.

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
@@ -522,6 +522,9 @@ public final class TypeAnnotationPass implements CompilerPass {
     allTypes.addAll(symbolToModule.keySet());
 
     String importedNamespace = nameUtil.findLongestNamePrefix(typeName, allTypes);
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -522,6 +522,10 @@ String convertTypeName(String sourceFile, String typeName) {
     allTypes.addAll(symbolToModule.keySet());
 
     String importedNamespace = nameUtil.findLongestNamePrefix(typeName, allTypes);
+    // Closure compilation can be permissive to the point where there are missing types.
+    // Gents supports these usecases by syntactically emitting the type as written. The file might
+    // not TS compiler, but at least the type is not lost and the user does not have to switch back
```
mprobst: "not TS compiler"?

rkirov: ops. done.

PR #569, opened 2017-09-25
Fix import syntax for an edge case where the module to import stays in closure and has no default exports.

bowenni:


src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -324,14 +324,19 @@ void convertRequireToImportStatements(Node n, String fullLocalName, String requi
...
+    // This is a javascript namespace
     if (module.shouldUseOldSyntax()) {
+      Node nodeToImport = null;
+      // If it has a default export then use `import foo from "goog:bar";`
+      if (module.hasDefaultExport) {
+        nodeToImport = Node.newString(Token.NAME, localName);
+      } else {
+        // If it doesn't have a default export then use `import * as foo from "goog:bar";`
+        nodeToImport = Node.newString(Token.IMPORT_STAR, localName);
+      }
```
rkirov: what about destructuring imports. At least add a todo for me to look into that here.

src/test/java/com/google/javascript/gents/multiTests/provide_imports/default_exports1_keep.js
```diff
@@ -0,0 +1,2 @@
+goog.provide('d.export1');
```
rkirov: provide_imports test is already very messy and hard to understand what is being tested. 

Can you either:
- extend jslib_imports, I think it's lib.B and lib.C are basically your export1 and export2.
- create a new test.

src/test/java/com/google/javascript/gents/multiTests/provide_imports/default_exports2_keep.js
```diff
@@ -0,0 +1,2 @@
+goog.provide('d.export2');
```
rkirov: I think you wanted this to be 'goog.module'. I am even surprised this works?!? It is not valid closure, AFAIK.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -324,14 +324,19 @@ void convertRequireToImportStatements(Node n, String fullLocalName, String requi
 
     String referencedFile = pathUtil.getImportPath(n.getSourceFileName(), module.file);
 
-    // Uses default import syntax as this is a javascript namespace
+    // This is a javascript namespace
```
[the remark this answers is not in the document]

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
bowenni: I tried to handle destructing imports but got lost and broke many unrelated tests.
I added a TODO and would like solve it in a different PR with a clean branch.

src/test/java/com/google/javascript/gents/multiTests/provide_imports/default_exports1_keep.js
bowenni: Moved to jslib_imports.

src/test/java/com/google/javascript/gents/multiTests/provide_imports/default_exports2_keep.js
bowenni: Done.

It's 6 test cases now:
{goog.module, goog.provide} × {no default exports, only default exports, both default and non-default exports}

src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
```diff
@@ -180,6 +191,10 @@ void registerProvidesModule(Node n, String file, String namespace) {
     /** {@code True}, if the module has any imports (e.g.{@code goog.require}). */
     private boolean hasImports = false;
 
+    private String namespace = "";
```
rkirov: Add a jsdoc for this new field. The word 'namespace' is used through out the code base to mean slightly different things, that create a lot of confusion. If you can quantify ('originalNamespace'?) it somehow it would be even better.

src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
```diff
@@ -180,6 +191,10 @@ void registerProvidesModule(Node n, String file, String namespace) {
...
+    /** Module has a default export. */
```
rkirov: Technically, prior to conversion there is no notion of 'default export' in goog.module system (its a purely ES6 concept). It would be more correct to say 'the module's clutz generated .d.ts will have a default export'

src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
bowenni: Done. Renamed to `namespaceInJS`.

src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
bowenni: Done.

PR #571, opened 2017-10-04
Fix clutz to emit properly modules that default export a single typedef.

rkirov:
A type-only goog.module appears significantly different from one with
values, so we need a new code path for that.


rkirov:
FYI, this was a pair programmed with Bowen.

PR #572, opened 2017-10-04
Add an option to skip emit on certain suffixes.

rkirov:
This is going to be used for tsickle
(https://github.com/angular/tsickle) emitted files, where we plan to
glue up the .d.ts instead of reemitting in clutz.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/tsickle_emit.skip.tsickle.js
+++ b/src/test/java/com/google/javascript/clutz/tsickle_emit.skip.tsickle.js
@@ -1,0 +1,7 @@
+goog.module('a.path.to.the.module');
+
+class KlosureKlass {
+
+}
+
+exports.KlosureKlass = KlosureKlass;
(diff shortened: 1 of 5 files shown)


rkirov:
FYI, this was a pair programmed with Lucas.

src/test/java/com/google/javascript/clutz/tsickle_emit.skip.tsickle.js
```diff
@@ -0,0 +1,7 @@
+goog.module('a.path.to.the.module');
+
+class KlosureKlass {
```
mprobst: If I remember correctly, we had some code that iterates through types used by other code, and still emits those.

Can you add a test for a situation where you have another file `emitted.js` that references a type from `tsickle_emit.skip.tsickle.js`?

rkirov: Ah, right. Fixed with https://github.com/angular/clutz/pull/574.

Issue #573, opened 2017-10-06
gents doesn't generate the right destructuring imports


CyberjujuM:
Gents transforms:

const {a, b, c, d, e} = goog.require('my.project.core.constants');

into:

import {a} from './core/constants';

Anything after "a" is dropped.

bowenni:
I think this is https://github.com/angular/clutz/issues/392
There's a [TODO](https://github.com/angular/clutz/blob/8eb366ce96efb95ed60a8cd8a0e57ff4ab7147bc/src/main/java/com/google/javascript/gents/ModuleConversionPass.java#L232)

Issue #392, opened 2016-10-14
Support multiple destructured imports in gents


dpurp:
Input:

``` javascript
var A = goog.require("default.A");
var X = goog.require("default.A");
const {namedExport1, namedExport2} = goog.require("named.A.B"); // THIS IS
var B = goog.require("namespace.A.B");
let C = goog.require("sideeffect.A.B.C");
const D = goog.require("both.A.B.C.D");
```

Expected Output:

``` typescript
import './export_sideeffect';

import * as DExports from './export_both';
import {D} from './export_both';
import {A} from './export_default';
import {A as X} from './export_default';
import * as B from './export_namespace';
import {namedExport1, namedExport2} from './export_named';
```

Actual:

``` typescript
import {A} from './export_default';
import {A as X} from './export_default';

// THIS LINE IS BROKEN
const {namedExport1, namedExport2} = goog.require('named.A.B');
import * as B from './export_namespace';
import './export_sideeffect';
import {D} from './export_both';
import * as DExports from './export_both';
```

dpurp:
#393 Adds support for a single destructured import; further work needs to be done to support multiple destructured imports.

PR #576, opened 2017-10-09
supports multiple destructured imports

bowenni:
Fixes https://github.com/angular/clutz/issues/392
Fixes https://github.com/angular/clutz/issues/573

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
@@ -17,7 +17,10 @@ import com.google.javascript.rhino.IR;
 import com.google.javascript.rhino.JSDocInfo;
 import com.google.javascript.rhino.Node;
 import com.google.javascript.rhino.Token;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 
@@ -221,20 +224,24 @@ public final class ModuleConversionPass implements CompilerPass {
           Node callNode = node;
           String requiredNamespace = callNode.getLastChild().getString();
           String localName = n.getFirstChild().getQualifiedName();
-          convertRequireToImportStatements(n, localName, requiredNamespace, false);
+          convertRequireToImportStatements(
+              n, Collections.singletonList(localName), requiredNamespace, false);
           return;
         }
 
-        // var {foo} = goog.require(...);
+        // var {foo, bar} = goog.require(...);
         if (node.isObjectPattern()
             && node.getNext().getFirstChild() != null
             && node.getNext().getFirstChild().matchesQualifiedName("goog.require")) {
-          // TODO(#392): Support multiple destructured import values here.
-          //     Currently, this only allows for a single destructured import value.
-          String namedExport = node.getFirstChild().getString();
+          Node importedNode = node.getFirstChild();
+          List<String> namedExports = new ArrayList();
+          while (importedNode != null) {
+            namedExports.add(importedNode.getString());
+            importedNode = importedNode.getNext();
+          }
           String requiredNamespace = node.getNext().getFirstChild().getNext().getString();
-          requiredNamespace += "." + namedExport;
-          convertRequireToImportStatements(n, namedExport, requiredNamespace, true);
+          requiredNamespace += "." + namedExports.get(0);
+          convertRequireToImportStatements(n, namedExports, requiredNamespace, true);
           return;
         }
       } else if (n.isExprResult()) {
@@ -245,7 +252,8 @@ public final class ModuleConversionPass implements CompilerPass {
         }
         if (callNode.getFirstChild().matchesQualifiedName("goog.require")) {
           String requiredNamespace = callNode.getLastChild().getString();
-          convertRequireToImportStatements(n, requiredNamespace, requiredNamespace, false);
+          convertRequireToImportStatements(
+              n, Collections.singletonList(requiredNamespace), requiredNamespace, false);
           return;
         }
       }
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -280,13 +288,28 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
...
    * </pre>
    */
   void convertRequireToImportStatements(
-      Node n, String fullLocalName, String requiredNamespace, boolean isDestructuringImports) {
+      Node n,
+      List<String> fullLocalNames,
+      String requiredNamespace,
+      boolean isDestructuringImports) {
+    // The rest of the functions assume that fullLocalNames contains one and only one element if
+    // this is a destructuring import.
```
rkirov: do you mean 'if this is not a destructuring import'?

If so, that is ok because it is invalid ES6 to have `import * as foo, * as bar` or to have anything but `*` on the LHS.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -297,14 +320,20 @@ void convertRequireToImportStatements(
...
     }
 
-    // Local name is the shortened namespace symbol
-    String localName = nameUtil.lastStepOfName(fullLocalName);
-
     FileModule module = namespaceToModule.get(requiredNamespace);
 
     String moduleSuffix = nameUtil.lastStepOfName(requiredNamespace);
+
+    // Local name is the shortened namespace symbol
```
rkirov: I don't understand this comment. It is a good time to expand on the comments, maybe and example would help here.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
rkirov: Otherwise, also ok, but file a bug because it is possible to have 'import {foo as bar, buz as buzz} from ' ...

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -372,9 +414,12 @@ void convertRequireToImportStatements(
...
       imported = true;
 
-      registerLocalSymbol(n.getSourceFileName(), fullLocalName, requiredNamespace, localName);
-      // Switch to back up name if necessary
-      localName = backupName;
+      for (int i = 0; i < fullLocalNames.size(); i++) {
+        registerLocalSymbol(
+            n.getSourceFileName(), fullLocalNames.get(i), requiredNamespace, localNames.get(i));
+        // Switch to back up name if necessary
+        localNames.set(i, backupNames.get(0));
```
rkirov: get(i) ?

mprobst:
I poked a bit at this code but didn't get to anything reviewable. My impression was that the logic that first combined the import path and the local name to a full name, and then later on splits it out again, should be refactored. It'd probably help to introduce a datastructure and/or augment the existing one to represent the different conditions for the import cases Rado lists.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
bowenni: Expanded the comment and added a few examples.

`localNames` is not always local names, it's the last part of the namespace if the goog.require() is a side effect only import. Somehow that's consistent with the rest of the code.
Surprise :)

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
bowenni: Yeah that's what I meant (I missed "not").

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -221,20 +224,24 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
+          Node importedNode = node.getFirstChild();
+          List<String> namedExports = new ArrayList();
+          while (importedNode != null) {
+            namedExports.add(importedNode.getString());
+            importedNode = importedNode.getNext();
+          }
           String requiredNamespace = node.getNext().getFirstChild().getNext().getString();
-          requiredNamespace += "." + namedExport;
-          convertRequireToImportStatements(n, namedExport, requiredNamespace, true);
+          requiredNamespace += "." + namedExports.get(0);
```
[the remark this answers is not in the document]
bowenni: TL;DR: pick any element from the namedExports it will work.

Reasons:
In `convertRequireToImportStatements()` function, we use `requiredNamespace` as a key to a few HashMaps for different purposes. For example there's a `namespaceToModule` HashMap that we use it to get the file module.
For non destructuring imports, this key is just the `requiredNamespace`. For multiple destructuring imports, the key is  \`${requiredNamespace}.${namedExport}\`. For example, given ` const {foo, bar} = goog.require('a');`, I would expect the `namespaceToModule` HashMap to contain ('a', fileModule), but instead it contains ('a.foo', fileModule) and ('a.bar', fileModule). They map to the same fileModule.

This variable should be renamed to `requiredNamespaceOrRequiredNamespacePlusNamedExportForDestructuringImports` to reflect what it really is.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java

bowenni:
@mprobst Yes I agree. This definitely needs some refactoring. There're some hidden surprises that don't really make sense.
The proper fix/refacotring is WIP but complex. Rado and I agreed to ship this inline fix to unblock the YouTube team first. I'm working on the refactoring in the mean time.

mprobst:
SGTM.

bowenni:
@rkirov  PTAL

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
rkirov: wow, surprise indeed :)

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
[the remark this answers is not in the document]
rkirov: I see. Thanks for explaining.

PR #577, opened 2017-10-13
refactors import conversion pass

bowenni:
The original `convertRequireToImportStatements()` function contains a big Cartesian product:
{`goog.require()`, `var x = goog.require()`, `var {x} = goog.require()`} × { `import x from 'goog:';`, `import {x} from './foo';`, `import * as x from './foo';`, `import './sideEffects';`} × {imported file kept in JS, imported file already in TS, imported file has side effect}
There are some hacks so all these conditions are handled in one function. For example `requiredNamespace` is appended with the local name in some cases and the suffix is stripped later.

This refactoring tries to split the destructuring case out. It also makes the variables really reflect what they contain. For example, `requiredNamespace` will be always the required namespace or module with no local names appended, making future code reading and bug fixes a little easier.

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
@@ -142,34 +142,34 @@ public final class ModuleConversionPass implements CompilerPass {
           }
           break;
         case GETPROP:
-        {
-          JSDocInfo jsdoc = NodeUtil.getBestJSDocInfo(child);
-          if (jsdoc == null || !jsdoc.containsTypeDefinition()) {
-            // GETPROPs on the root level are only exports for @typedefs
+          {
+            JSDocInfo jsdoc = NodeUtil.getBestJSDocInfo(child);
+            if (jsdoc == null || !jsdoc.containsTypeDefinition()) {
+              // GETPROPs on the root level are only exports for @typedefs
+              break;
+            }
+            if (!fileToModule.containsKey(fileName)) {
+              break;
+            }
+            FileModule module = fileToModule.get(fileName);
+            Map<String, String> symbols = module.exportedNamespacesToSymbols;
+            String exportedNamespace = nameUtil.findLongestNamePrefix(child, symbols.keySet());
+            if (exportedNamespace != null) {
+              String localName = symbols.get(exportedNamespace);
+              Node export =
+                  new Node(
+                      Token.EXPORT,
+                      new Node(
+                          Token.EXPORT_SPECS,
+                          new Node(Token.EXPORT_SPEC, Node.newString(Token.NAME, localName))));
+              export.useSourceInfoFromForTree(child);
+              parent.addChildAfter(export, n);
+              // Registers symbol for rewriting local uses.
+              registerLocalSymbol(
+                  child.getSourceFileName(), exportedNamespace, exportedNamespace, localName);
+            }
             break;
           }
-          if (!fileToModule.containsKey(fileName)) {
-            break;
-          }
-          FileModule module = fileToModule.get(fileName);
-          Map<String, String> symbols = module.exportedNamespacesToSymbols;
-          String exportedNamespace = nameUtil.findLongestNamePrefix(child, symbols.keySet());
-          if (exportedNamespace != null) {
-            String localName = symbols.get(exportedNamespace);
-            Node export =
-                new Node(
-                    Token.EXPORT,
-                    new Node(
-                        Token.EXPORT_SPECS,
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -280,8 +294,109 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
     }
   }
 
+  /** Encapsulates a {@code goog.require(...)} statement. */
```
rkirov: 'A single statement containing goog.require.' otherwise it is not clear that you cover 'const {a,b} = goog.require(...)'.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -224,8 +224,11 @@ public void visit(NodeTraversal t, Node n, Node parent) {
           Node callNode = node;
           String requiredNamespace = callNode.getLastChild().getString();
           String localName = n.getFirstChild().getQualifiedName();
-          convertRequireToImportStatements(
-              n, Collections.singletonList(localName), requiredNamespace, false);
+          ModuleImport moduleImport =
```
rkirov: I feel most of the body of this 'visit' method should be in the ctor for ModuleImport, or at least a static method (you might want to return null if the node is misidentified).

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -292,192 +407,206 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
...
-    if (n.getFirstChild() != null && n.getFirstChild().isDestructuringLhs()) {
+    Node requireLHS = moduleImport.originalImportNode.getFirstChild();
+    if (moduleImport.isDestructuringImport) {
       importSpec = new Node(Token.IMPORT_SPECS);
-      for (String fullLocalName : fullLocalNames) {
+      for (String fullLocalName : moduleImport.fullLocalNames) {
         importSpec.addChildToBack(IR.name(fullLocalName));
       }
-    } else if (n.getFirstChild() != null && n.getFirstChild().isName()) {
+    } else if (requireLHS != null && requireLHS.isName()) {
```
rkirov: moduleImport can have a method for this too - maybe 'isFullModuleImport'?

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -280,8 +294,109 @@ public boolean shouldTraverse(NodeTraversal t, Node n, Node parent) {
...
+        boolean isDestructuringImport) {
+      this.originalImportNode = originalImportNode;
+      this.fullLocalNames = fullLocalNames;
+      this.requiredNamespace = requiredNamespace;
+      this.isDestructuringImport = isDestructuringImport;
+      this.module = namespaceToModule.get(requiredNamespace);
+      this.referencedFile =
+          pathUtil.getImportPath(originalImportNode.getSourceFileName(), module.file);
+      this.moduleSuffix = nameUtil.lastStepOfName(requiredNamespace);
+      this.backupName = this.moduleSuffix;
```
rkirov: Not sure if possible, but might be a good idea to encapsulate the backup logic within ModuleImport class.

Is it possible for the decision whether to use localName or backupName to be behind a getNameFor...() method?

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
bowenni: I could move the entire body of `visit()` to a `maybeConvertRequireToImport(Node n)` function but not sure what we get from that. The current implementation seems quite readable to me.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
bowenni: That's a good idea, but I tried it out and it seems to complicate the logic.
backup name is only used in a very specific case: A non destructuring require that converts to be two import statements, both `import {value} from './file';` and `import * as valueExports from './file';`. In this specific case, the latter import will use the backup name. So we need to add some extra flags in ModuleImport class for getNameFor...() method to work properly.
When it's inlined in the `convertNonDestructuringRequireToImportStatements()`, it's quite clear when we switch to backup name. So I'd keep it as is.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java

PR #578, opened 2017-10-13
Adds enum support in gents.

rkirov:
For string and number enums convert them to TypeScript, all other
types of enums are left as object literals.

### DIFF
--- a/src/test/java/com/google/javascript/gents/singleTests/enum.ts
+++ b/src/test/java/com/google/javascript/gents/singleTests/enum.ts
@@ -1,0 +1,40 @@
+
+/**
+ * A non-trivial comment.
+ * @enum {number}
+ */
+export enum NumEnum {
+  A,
+  B
+}
+
+/** @enum {number} */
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/gents/singleTests/enum.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A non-trivial comment.
+ * @enum {number}
+ */
+export enum NumEnum {
+  A,
+  B
+}
+
+/** @enum {number} */
```
bowenni: It'd nice if we can remove the Closure annotations for converted enums.

rkirov: Yes, see my commit comment. It appears to be all or nothing decision and I went for nothing, because of the cases not coverred. Maybe we can refactor gents to keep track of what was converted. In anycase that would require some heavier changes, maybe we can pair on that on Monday.

PR #580, opened 2017-10-28
Remove calls to reportCodeChange, which is going away soon

MatrixFrog:


MatrixFrog:
It's difficult to debug this because the errors' stack traces aren't showing up in the Travis output, or when I run this locally. Maybe someone from the Clutz team can use this as a starting point though.

MatrixFrog:
Nevermind, found a flag to show the stacktrace, got everything fixed. Failures look the same as in https://github.com/angular/clutz/pull/581 so unrelated to this PR, I think.

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -485,7 +485,6 @@ void convertConstructorToClass(Node n, JSDocInfo jsDoc) {
 
     typeNode.setJSDocInfo(n.getJSDocInfo());
     nodeComments.replaceWithComment(n, typeNode);
-    compiler.reportCodeChange();
```
rkirov: why is it ok to sometimes 'reportCodeChange' and sometimes just remove it? I might be lacking some basic understanding of how closure's AST mutation works.

MatrixFrog: To be honest I'm not sure. It might be that there's another change report in the same change scope making this one unnecessary. It also might be that there really *should* be a report here, but since this is a standalone tool rather than an optimization pass in the compiler, it's not that critical anyway.

PR #582, opened 2017-11-02
Upgrade to closure v20171023

rkirov:
Files without goog.module or goog.provide, appear to report a new
synthetic provide 'module$src$...', which we now filter out in clutz.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -427,9 +427,20 @@ class DeclarationGenerator {
       String originalPath = compilerInput.getSourceFile().getOriginalPath();
       if (opts.skipEmitRegExp != null && Pattern.matches(opts.skipEmitRegExp, originalPath))
         continue;
-      transitiveProvides.addAll(compilerInput.getProvides());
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -427,9 +427,20 @@ String produceDts(Depgraph depgraph) {
       String originalPath = compilerInput.getSourceFile().getOriginalPath();
       if (opts.skipEmitRegExp != null && Pattern.matches(opts.skipEmitRegExp, originalPath))
         continue;
-      transitiveProvides.addAll(compilerInput.getProvides());
+      Collection<String> inputProvides = compilerInput.getProvides();
+      Collection<String> filteredProvides = new ArrayList<>();
```
evmar: This seems unrelated to the change?

rkirov: It is related. Its in the commit description

rkirov: The new version of closure started reporting new synthetic provides. If we dont filter them out we create a bunch of garbage goog: module exports.

rkirov:
Yes, simply upgrading closure, makes all goldens with empty files fail.

PR #583, opened 2017-11-16
 Do not use typedefs that are not emitted.

mprobst:
`skipEmitRegExp` would lead to typedefs that are not emitted being referenced by code that is emitted.

This also contains a number of minor fixups I made on the way.


mprobst:
G3 is green.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -337,9 +337,11 @@ void collectTypedefs() {
       if (type == null
```
evmar: Better first line for this commit: "Minor syntax fixes".  ("Minor fixes" is too vague, hard to know if it's fixing minor clutz bugs or some other thing.)

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -683,8 +700,7 @@ private void processUnprovidedTypes(Set<String> provides, Set<String> transitive
         if (symbolInput != null && symbolInput.isExtern()) continue;
 
         // skip types that come from files that we skip emit on.
```
evmar: This comment now reads like it's peering into the impl of shouldSkipVar.  I'd say maybe just drop it.

.gitignore
```diff
@@ -15,5 +15,8 @@ out/
 *.idea
```
evmar: Commit title: "*git* ignore [...]" (thought it was talking about making clutz ignore it somehow)

mprobst: Done.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Done.

PR #585, opened 2017-11-18
Remove StringArrayOptionHandler from depgraphs and externs options

whymarrh:
This change removes `StringArrayOptionHandler` from the depgraphs and externs options (`--depsgraphs` and `--externs`) as they cause file names with spaces to be incorrectly split on the spaces.

As per conventional shell quoting, passing `--externs 'foo bar.js'` should result in `["foo bar.js"]` but with the handler attached it would result in `["foo", "bar.js"]`. This behaviour is the expected behaviour according to [the source for `StringArrayOptionHandler` in args4j](https://git.io/vF92d) that contains a comment with the following:

>     java -jar aaa.jar -s banan hruska jablko
>     java -jar aaa.jar -s banan "hruska jablko"
>     java -jar aaa.jar -s "banan hruska jablko"
>     java -jar aaa.jar -s banan hruska jablko -l 4 -r
>     java -jar aaa.jar -t 222 -s banan hruska jablko -r
>
> All of them result in a single string array that contains three tokens:
> `banan`, `hruska`, and `jablko`.

I don't think this handler is appropriate for options representing file names.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -29,16 +29,14 @@ public class Options {
   @Option(
     name = "--externs",
     usage = "list of files to read externs definitions (as separate args)",
-    metaVar = "EXTERN...",
-    handler = StringArrayOptionHandler.class
+    metaVar = "EXTERN..."
   )
   List<String> externs = new ArrayList<>();
 
   @Option(
     name = "--depgraphs",
     usage = "only generate output for files listed as a root in the given depgraphs",
-    metaVar = "file.depgraph...",
-    handler = StringArrayOptionHandler.class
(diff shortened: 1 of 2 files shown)


whymarrh:
I now realize that the flag names don't really make sense as plural anymore—should I rename them?

mprobst:
Ack, splitting on space there is wrong. Surprising how long we made it without anybody running into files with whitespace in them :-)

If we want to stay interface compatible, we could implement our own `StringArrayOptionHandler` style class that'd accept a list of values following `--externs`, but doesn't split the individual strings on whitespace.

Would you care to give that a spin?

whymarrh:
I'll try my hand at that, sure. (I'll close this PR here as, yeah, it's a breaking change.)

PR #586, opened 2017-11-18
Replace args4j StringArrayOptionHandler to handle quoted params

whymarrh:
This PR fixes the handling of quoted command line arguments by replacing the args4j string array option handler with one that does not split params on their whitespace.

As an example, running the following should respect the conventional shell quoting:

```bash
clutz a.js b.js c.js \
    --externs foo.js bar.js 'baz qux.js' \
    -o out.d.ts
```

And should read `baz qux.js` as a file name with a space in it.

The `StringArrayOptionHandler` included in args4j incorrectly split `baz qux.js` into two tokens.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -16,9 +16,50 @@ import org.kohsuke.args4j.Argument;
 import org.kohsuke.args4j.CmdLineException;
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
-import org.kohsuke.args4j.spi.StringArrayOptionHandler;
+import org.kohsuke.args4j.OptionDef;
+import org.kohsuke.args4j.spi.OptionHandler;
+import org.kohsuke.args4j.spi.Parameters;
+import org.kohsuke.args4j.spi.Setter;
 
 public class Options {
+  /**
+   * An {@link OptionHandler} the parses an array of strings as an option.
+   *
+   * <p>Whereas the builtin {@link org.kohsuke.args4j.spi.StringArrayOptionHandler} will split
+   * parameters with spaces in them into their space-delimited components (e.g. {@code "foo bar"}
+   * into {@code foo} and {@code bar}), this option handler does not:
+   *
+   * <pre>
+   *  java Example --foo bar baz 'bar baz'
+   *  // => Results in 'bar', 'baz', and 'bar baz'
+   * </pre>
+   *
+   * This approach follows the conventions of shell quoting to allow option values with spaces.
+   */
+  public static class StringArrayOptionHandler extends OptionHandler<String> {
+    public StringArrayOptionHandler(CmdLineParser parser, OptionDef option, Setter<String> setter) {
+      super(parser, option, setter);
+    }
+
+    @Override
+    public String getDefaultMetaVariable() {
+      return "ARG ...";
+    }
+
+    @Override
+    public int parseArguments(Parameters params) throws CmdLineException {
+      final int paramsSize = params.size();
+      for (int i = 0; i < paramsSize; i++) {
+        String param = params.getParameter(i);
+        if (param.startsWith("-")) {
+          return i;
+        }
+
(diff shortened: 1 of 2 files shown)


whymarrh:
This PR comes as a replacement for #585

src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -16,9 +16,36 @@
...
 import org.kohsuke.args4j.CmdLineParser;
 import org.kohsuke.args4j.Option;
-import org.kohsuke.args4j.spi.StringArrayOptionHandler;
+import org.kohsuke.args4j.OptionDef;
+import org.kohsuke.args4j.spi.OptionHandler;
+import org.kohsuke.args4j.spi.Parameters;
+import org.kohsuke.args4j.spi.Setter;
 
 public class Options {
+  public static class StringArrayOptionHandler extends OptionHandler<String> {
```
mprobst: doesn't need to be public, does it?

mprobst: maybe this should have some minimal docs explaining why we have our own handler here (in particular given it has the same name as the standard one)?

whymarrh: Sure thing, I'll add a docblock

whymarrh: I think it does, though I'll double-check

mprobst: package visibility (no keyword) should be enough if you only need it from the test.

whymarrh: It also needs to be discoverable by args4j, which uses reflection to find the class. I had run into an issue with the visibility, but I'll see if args4j can find it with package visibility later today.

whymarrh: Looking at it now it appears, yes, this class does need to be `public`. When it is package private I get the following from args4j:

```text
Exception in thread "main" org.kohsuke.args4j.IllegalAnnotationError: java.lang.IllegalAccessException: Class org.kohsuke.args4j.OptionHandlerRegistry$DefaultConstructorHandlerFactory can not access a member of class com.google.javascript.clutz.Options$StringArrayOptionHandler with modifiers "public"
	at org.kohsuke.args4j.OptionHandlerRegistry$DefaultConstructorHandlerFactory.getHandler(OptionHandlerRegistry.java:226)
	at org.kohsuke.args4j.OptionHandlerRegistry.createOptionHandler(OptionHandlerRegistry.java:188)
	at org.kohsuke.args4j.CmdLineParser.createOptionHandler(CmdLineParser.java:178)
	at org.kohsuke.args4j.CmdLineParser.addOption(CmdLineParser.java:146)
	at org.kohsuke.args4j.ClassParser.parse(ClassParser.java:34)
	at org.kohsuke.args4j.CmdLineParser.<init>(CmdLineParser.java:96)
	at org.kohsuke.args4j.CmdLineParser.<init>(CmdLineParser.java:71)
	at com.google.javascript.clutz.Options.<init>(Options.java:164)
	at com.google.javascript.clutz.DeclarationGenerator.main(DeclarationGenerator.java:239)
Caused by: java.lang.IllegalAccessException: Class org.kohsuke.args4j.OptionHandlerRegistry$DefaultConstructorHandlerFactory can not access a member of class com.google.javascript.clutz.Options$StringArrayOptionHandler with modifiers "public"
	at sun.reflect.Reflection.ensureMemberAccess(Reflection.java:102)
	at java.lang.reflect.AccessibleObject.slowCheckMemberAccess(AccessibleObject.java:296)
	at java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:288)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:413)
	at org.kohsuke.args4j.OptionHandlerRegistry$DefaultConstructorHandlerFactory.getHandler(OptionHandlerRegistry.java:222)
	... 8 more
```

mprobst:
Makes sense. Thanks for the contribution!

PR #588, opened 2017-11-18
Remove line terminators from emitted comments

whymarrh:
This PR updates the comments emitted to strip [any line terminator code points][1] from the line.

  [1]:http://www.ecma-international.org/ecma-262/6.0/#table-33

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -10,6 +10,7 @@ import static com.google.javascript.rhino.jstype.JSTypeNative.STRING_TYPE;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.stream.Collectors.toCollection;
 
+import com.google.common.base.CharMatcher;
 import com.google.common.base.Joiner;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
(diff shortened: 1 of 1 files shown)


whymarrh:
I ran into this issue where newlines would break the outputted defns as it would contain single-line comments about properties being skipped whose names contained line terminators.

e.g.

```
//!! skipping property goog.string.specialEscapeChars_. because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_. because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_.	 because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_. because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_. because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_. because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_. because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_." because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_.< because it is not a valid symbol.
//!! skipping property goog.string.specialEscapeChars_.\ because it is not a valid symbol.
```

mprobst:
Should this maybe rather skip or escape unprintable characters? It's not only line terminators after all, it's also \0 in the example above, or potentially anything else below ASCII <32.

mprobst:
Also see `com.google.common.base.CharMatcher`, might make for an easier implementation (`replaceFrom`).

whymarrh:
> Should this maybe rather skip or escape unprintable characters? It's not only line terminators after all, it's also `\0` in the example above, or potentially anything else below ASCII <32. Not sure what character set TS allows in comments.

I think the only characters that are truly an issue are the line terminators as those will end the comment and start a new uncommented line. What issue(s) will `\0` cause?

> Also see `com.google.common.base.CharMatcher`, might make for an easier implementation (`replaceFrom`).

Neat, thanks. I'll take a look.

whymarrh:
Reading your comment again, I think yes, it would be ideal if we could somehow escape all unprintable characters, but I don't know if that possible. I don't know of a way to escape a newline in a single-line comment, for example.

PR #591, opened 2017-11-22
Resolve Module Bindings

### LINKED ISSUE #551: incremental clutz need to resolve module bindings
In the following scenario:

```typescript
goog.module('other.module');

const {C} = goog.require('a.b.c');

/** @type {C} */
const c = new C();
exports.c = c;
```

Incremental Clutz (i.e. run with --partialInputs on this one file) will emit verbatim the type of the binding `C` as `ಠ_ಠ.clutz.module$contents$other$module_C`.

The type to emit that would correspond to what is emitted on the other end is ` ಠ_ಠ.clutz.module$exports$a$b$c$C`

LucasSloan:
When parsing partial inputs, closure doesn't know the proper name of an imported symbol, so it emits it as part of the current module.

This PR walks the closure ast to grab the module id of the current file as well as the id of the import to construct a map of unresolved symbol names to resolved symbol names, and uses it to replace unresolved names when emitting NoResolvedType nodes.

Fixes #551

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -24,7 +24,6 @@ import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
 import com.google.common.io.Files;
 import com.google.javascript.jscomp.AbstractCommandLineRunner;
-import com.google.javascript.jscomp.Compiler;
 import com.google.javascript.jscomp.CompilerInput;
 import com.google.javascript.jscomp.DiagnosticType;
 import com.google.javascript.jscomp.ErrorFormat;
@@ -272,7 +271,7 @@ class DeclarationGenerator {
   @Nullable private JSType iteratorIterableType;
 
   private final Options opts;
-  private final Compiler compiler;
+  private final InitialParseRetainingCompiler compiler;
   private final ClutzErrorManager errorManager;
   private StringWriter out = new StringWriter();
 
@@ -294,9 +293,15 @@ class DeclarationGenerator {
    */
   private final Set<String> typesUsed = new LinkedHashSet<>();
 
+  /**
+   * In partial mode, closure doesn't know the correct name of imported symbols, and importRenameMap
+   * is used to store the mappings from the closure supplied names to the correct names.
+   */
+  private Map<String, String> importRenameMap = Collections.emptyMap();
+
   DeclarationGenerator(Options opts) {
     this.opts = opts;
-    this.compiler = new Compiler();
+    this.compiler = new InitialParseRetainingCompiler();
     compiler.disableThreads();
     this.errorManager =
         new ClutzErrorManager(
@@ -404,6 +409,9 @@ class DeclarationGenerator {
       throws AssertionError {
     // Compile should always be first here, because it sets internal state.
     compiler.compile(externs, sourceFiles, opts.getCompilerOptions());
+    if (opts.partialInput) {
+      importRenameMap = ImportRenameMapBuilder.build(compiler.getParsedInputs());
+    }
     unknownType = compiler.getTypeRegistry().getNativeType(JSTypeNative.UNKNOWN_TYPE);
     numberType = compiler.getTypeRegistry().getNativeType(JSTypeNative.NUMBER_TYPE);
     iterableType = compiler.getTypeRegistry().getType("Iterable");
@@ -1669,6 +1677,11 @@ class DeclarationGenerator {
         return;
       }
       String displayName = type.getDisplayName();
+      // In partial mode, closure doesn't know the correct name of imported symbols, if the name
+      // matches one in the precomputed map, replace it with the original declared name
+      if (importRenameMap.containsKey(displayName)) {
+        displayName = importRenameMap.get(displayName);
+      }
       emit(Constants.INTERNAL_NAMESPACE + "." + displayName);
       List<JSType> templateTypes = nType.getTemplateTypes();
       if (templateTypes != null && templateTypes.size() > 0) {
--- a/src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
+++ b/src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
@@ -1,0 +1,161 @@
+package com.google.javascript.clutz;
+
+import com.google.javascript.rhino.Node;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * If Clutz is running in incremental mode, closure can't resolve imported symbols, so it gives them
+ * names in the current namespace (ie `module$contents$current$file`), instead of names that
+ * reference the original namespace (ie `module$exports$imported$file`). ImportRenameMapBuilder
+ * builds a map from the local name to the exported name so the original names can be substituted.
+ */
+public class ImportRenameMapBuilder {
+
+  /**
+   * Build takes a collection of source files, parses them with the compiler, and walks the ast to
+   * find any `const variable = goog.require()` statements to build a map from the name closure
+   * generates in incremental mode to the exported name of the import.
+   */
+  public static Map<String, String> build(Collection<Node> parsedInputs) {
+    Map<String, String> importRenameMap = new HashMap<>();
+    for (Node ast : parsedInputs) {
+      String moduleId = getGoogModuleId(ast);
+      // Closure can only put the symbol name into the current module's namespace, if the source
+      // file is a module, so if it isn't a module, just bail
+      if (moduleId != null) {
+        importRenameMap.putAll(build(moduleId, ast));
+      }
+    }
+    return importRenameMap;
+  }
+
+  private static boolean isGoogModuleCall(Node statement) {
+    if (!statement.isExprResult()) {
+      return false;
+    }
+
+    Node expression = statement.getFirstChild();
+    return expression.isCall() && expression.getFirstChild().matchesQualifiedName("goog.module");
+  }
+
+  private static boolean isGoogRequireAssignment(Node statement) {
+    if (!(statement.isConst() || statement.isVar() || statement.isLet())) {
+      return false;
+    }
+
+    Node rightHandSide = statement.getFirstFirstChild();
+
+    return rightHandSide != null
+        && rightHandSide.isCall()
+        && rightHandSide.getFirstChild().matchesQualifiedName("goog.require");
+  }
+
+  private static boolean isGoogRequireDestructuringAssignment(Node statement) {
+    if (!(statement.isConst() || statement.isVar() || statement.isLet())) {
+      return false;
+    }
+
+    if (!statement.getFirstChild().isDestructuringLhs()) {
+      return false;
+    }
+
+    Node destructuringAssignment = statement.getFirstChild();
+
+    Node rightHandSide = destructuringAssignment.getChildAtIndex(1);
+
+    return rightHandSide.isCall()
+        && rightHandSide.getFirstChild().matchesQualifiedName("goog.require");
+  }
+
+  private static String getGoogModuleId(Node astRoot) {
+    if (astRoot.getFirstChild() == null || !astRoot.getFirstChild().isModuleBody()) {
+      return null;
+    }
+
+    Node moduleBody = astRoot.getFirstChild();
+    for (Node statement : moduleBody.children()) {
+      if (isGoogModuleCall(statement)) {
+        return statement.getFirstChild().getChildAtIndex(1).getString();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Build does the actual work of walking over the AST, finding any goog.require() assignments or
+   * destructuring assignments, parsing them, and generating the mappings from local symbol names to
+   * exported symbol names.
+   */
+  private static Map<String, String> build(String localModuleId, Node astRoot) {
+    Map<String, String> importRenameMap = new HashMap<>();
+    if (astRoot.getFirstChild() == null || !astRoot.getFirstChild().isModuleBody()) {
+      return importRenameMap;
+    }
+
+    Node moduleBody = astRoot.getFirstChild();
+    for (Node statement : moduleBody.children()) {
+      if (isGoogRequireAssignment(statement)) {
+        // `const C = goog.require()`
+        String importedModuleId = statement.getFirstFirstChild().getChildAtIndex(1).getString();
+        String variableName = statement.getFirstChild().getString();
+
+        String localSymbolName = buildLocalSymbolName(localModuleId, variableName);
+        String exportedSymbolName = buildWholeModuleExportSymbolName(importedModuleId);
+        importRenameMap.put(localSymbolName, exportedSymbolName);
+      } else if (isGoogRequireDestructuringAssignment(statement)) {
+        // `const {C, Clazz: RenamedClazz} = goog.require()`
+        String importedModuleId =
+            statement.getFirstChild().getChildAtIndex(1).getChildAtIndex(1).getString();
+        for (Node destructured : statement.getFirstFirstChild().children()) {
+          String originalName = destructured.getString();
+          // Destructuring can use the original name `const {A} = goog.require("foo.a")` or rename
+          // it `const {A: RenamedA} = ...`, and closure uses whichever in the symbol name it
+          // generates, so we have to extract it.
+          String variableName;
+          if (destructured.getFirstChild() != null) {
+            variableName = destructured.getFirstChild().getString();
+          } else {
+            variableName = originalName;
+          }
+
+          String localSymbolName = buildLocalSymbolName(localModuleId, variableName);
+          String exportedSymbolName = buildNamedExportSymbolName(importedModuleId, originalName);
+          importRenameMap.put(localSymbolName, exportedSymbolName);
+        }
+      }
+    }
+
+    return importRenameMap;
+  }
+
+  /**
+   * The exported symbol can take 2 forms - one where it refers to everything that the module
+   * exports and another where it refers to just one thing the module exports. If the original
+   * module used the `exports = ...` style, the symbol name is just the module name.
+   *
+   * <p>TODO(lucassloan): this only holds for importing from a goog.module see:
+   * https://github.com/angular/clutz/issues/596
+   */
+  private static String buildWholeModuleExportSymbolName(String importedModuleId) {
+    return "module$exports$" + importedModuleId.replace(".", "$");
+  }
+
+  /**
+   * The exported symbol can take 2 forms - one where it refers to everything that the module
+   * exports and another where it refers to just one thing the module exports. If the original
+   * module used the `exports.foo = ...` style, the symbol name is the module name plus the
+   * individual export's name.
+   *
+   * <p>TODO(lucassloan): this only holds for importing from a goog.module see:
+   * https://github.com/angular/clutz/issues/596
+   */
+  private static String buildNamedExportSymbolName(String importedModuleId, String originalName) {
+    return "module$exports$" + importedModuleId.replace(".", "$") + "_" + originalName;
+  }
+
+  private static String buildLocalSymbolName(String importingModuleId, String variableName) {
+    return "module$contents$" + importingModuleId.replace(".", "$") + "_" + variableName;
+  }
+}
--- a/src/test/java/com/google/javascript/clutz/partial/missing_base.d.ts
+++ b/src/test/java/com/google/javascript/clutz/partial/missing_base.d.ts
@@ -1,4 +1,4 @@
-declare namespace ಠ_ಠ.clutz.module$exports$mising$extend {
+declare namespace ಠ_ಠ.clutz.module$exports$missing$extend {
   class B extends B_Instance {
   }
   class B_Instance extends ಠ_ಠ.clutz.direct.ref.A {
@@ -7,14 +7,22 @@ declare namespace ಠ_ಠ.clutz.module$exports$mising$extend {
   }
   class BTemplated_Instance extends ಠ_ಠ.clutz.direct.ref.ATemplated < string , number > {
   }
-  class D extends D_Instance {
+  class ClassExtendingMissingDestructuredRequire extends ClassExtendingMissingDestructuredRequire_Instance {
   }
-  //!! This emit is wrong and will be fixed with
-  //!! https://github.com/angular/clutz/issues/551.
-  class D_Instance extends ಠ_ಠ.clutz.module$contents$mising$extend_C {
+  class ClassExtendingMissingDestructuredRequire_Instance extends ಠ_ಠ.clutz.module$exports$missing$base_MissingDestructuredRequire {
   }
+  class ClassExtendingMissingRequire extends ClassExtendingMissingRequire_Instance {
+  }
+  class ClassExtendingMissingRequire_Instance extends ಠ_ಠ.clutz.module$exports$missing$base {
+  }
+  class ClassExtendingRenamedDestructuredRequire extends ClassExtendingRenamedDestructuredRequire_Instance {
+  }
+  class ClassExtendingRenamedDestructuredRequire_Instance extends ಠ_ಠ.clutz.module$exports$missing$base_OriginalName {
+  }
+  var DeclarationOfMissingRequire : ಠ_ಠ.clutz.module$exports$missing$base | null ;
+  function FuncWithMissingRequireParam (c : ಠ_ಠ.clutz.module$exports$missing$base | null ) : void ;
 }
-declare module 'goog:mising.extend' {
-  import alias = ಠ_ಠ.clutz.module$exports$mising$extend;
+declare module 'goog:missing.extend' {
+  import alias = ಠ_ಠ.clutz.module$exports$missing$extend;
   export = alias;
 }
--- a/src/test/java/com/google/javascript/clutz/partial/missing_base.js
+++ b/src/test/java/com/google/javascript/clutz/partial/missing_base.js
@@ -1,4 +1,4 @@
-goog.module('mising.extend');
+goog.module('missing.extend');
 
 //!! Using class syntax requires default externs but the test runs faster
 //!! without them. Just use @constructor instead.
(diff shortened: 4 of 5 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
     }
   }
 
+  private boolean isGoogCall(Node node) {
+    if (!node.isCall()) {
```
mprobst: Closures uses `matchesQualifiedName` with a constant constructed as `IR.getprop(IR.name("goog"), IR.string("require"))`. Maybe that'd be a bit less wordy here?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+
+    Node getPropNode = node.getFirstChild();
+
+    if (getPropNode.getChildAtIndex(1) == null || !getPropNode.getChildAtIndex(1).isString()) {
+      return false;
+    }
+    return getPropNode.getChildAtIndex(1).getString().equals("require");
+  }
+
+  private boolean isGoogModule(Node node) {
```
mprobst: It's unclear from the name whether this checks if the Node is a source file that's a goog.module, or whether it checks if this is the goog.module call. Maybe "call" should appear in the name?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    for (Node child : node.children()) {
+      String moduleId = getGoogModuleId(child);
+      if (moduleId != null) {
+        return moduleId;
+      }
+    }
+    return null;
+  }
+
+  private Map<String, String> getImportRenameMap(String moduleId, Node node) {
```
mprobst: maybe `buildImportRenameMap`? This sounds like a simple getter, where it's actually a complex calculation.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    Map<String, String> importRenameMap = new HashMap<>();
+    if (isGoogRequireAssignment(node)) {
+      String variableName = node.getFirstChild().getString();
+      String importingModuleId = node.getFirstFirstChild().getChildAtIndex(1).getString();
+      importRenameMap.put(
```
mprobst: document where these strings come from, and that it means to establish this mapping.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    }
+
+    if (node.getFirstFirstChild() == null) {
+      return false;
+    }
+
+    return isGoogRequire(node.getFirstFirstChild());
+  }
+
+  private String getGoogModuleId(Node node) {
```
mprobst: document what `node` is here?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    if (isGoogModule(node)) {
+      return node.getChildAtIndex(1).getString();
+    }
+    for (Node child : node.children()) {
```
mprobst: document why this is the right thing to do?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+
+    return importRenameMap;
+  }
+
   String generateDeclarations(
       List<SourceFile> sourceFiles, List<SourceFile> externs, Depgraph depgraph)
       throws AssertionError {
     // Compile should always be first here, because it sets internal state.
+    compiler.initOptions(opts.getCompilerOptions());
+    for (SourceFile sf : sourceFiles) {
```
mprobst: could you move this entire calculation out of DeclarationGenerator? The file is very large, and this seems like a fairly easy thing to pull out.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    }
+
+    return importRenameMap;
+  }
+
   String generateDeclarations(
       List<SourceFile> sourceFiles, List<SourceFile> externs, Depgraph depgraph)
       throws AssertionError {
     // Compile should always be first here, because it sets internal state.
+    compiler.initOptions(opts.getCompilerOptions());
```
mprobst: why is this call necessary? Document?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    for (SourceFile sf : sourceFiles) {
+      Node ast = compiler.parse(sf);
+      String moduleId = getGoogModuleId(ast);
+      if (moduleId != null) {
```
mprobst: what if `moduleId == null`? document why this only applies to `goog.module`. There's also `goog.scope()` to consider, but that's for a later CL.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+    Map<String, String> importRenameMap = new HashMap<>();
+    if (isGoogRequireAssignment(node)) {
+      String variableName = node.getFirstChild().getString();
+      String importingModuleId = node.getFirstFirstChild().getChildAtIndex(1).getString();
+      importRenameMap.put(
+          "module$contents$" + moduleId.replace(".", "$") + "_" + variableName,
+          "module$exports$" + importingModuleId.replace(".", "$") + "_" + variableName);
+    }
+    for (Node child : node.children()) {
+      importRenameMap.putAll(getImportRenameMap(moduleId, child));
```
mprobst: why is this recursive? Docs...

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1669,6 +1769,9 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
         return;
       }
       String displayName = type.getDisplayName();
+      if (importRenameMap.containsKey(displayName)) {
```
mprobst: docs

src/test/java/com/google/javascript/clutz/partial/missing_base.js
```diff
@@ -28,6 +28,18 @@ function D() {
 
 }
 
+/**
+ * @param {C!} c
```
mprobst: please use speaking names for all symbols,  future reviewers will praise their ability to understand what this is testing :-)

mprobst: `@param {C!}` seems odd, shouldn't the `!` go before the type?

src/test/java/com/google/javascript/clutz/partial/missing_base.js
```diff
@@ -1,4 +1,4 @@
-goog.module('mising.extend');
+goog.module('missing.extend');
```
mprobst: this seems to be lacking a few cases:
- the recursive call in getImportRenameMap
- a file not containing a goog.module call, but top level assignments (maybe that's a Closure level error?)
- destructuring call `const {x} = goog.require(...);`
- top level `goog.require` where the symbol required is from a `goog.provide` style file, not from a `goog.module`

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -399,10 +401,108 @@ void generateDeclarations() {
...
+      }
+    }
+    return null;
+  }
+
+  private Map<String, String> getImportRenameMap(String moduleId, Node node) {
+    Map<String, String> importRenameMap = new HashMap<>();
+    if (isGoogRequireAssignment(node)) {
+      String variableName = node.getFirstChild().getString();
+      String importingModuleId = node.getFirstFirstChild().getChildAtIndex(1).getString();
```
mprobst: is this correct in the recursive case?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: whose `moduleId` is `moduleId`, the importer or importee? maybe rename the variables?

src/test/java/com/google/javascript/clutz/partial/missing_base.d.ts
```diff
@@ -11,10 +11,12 @@ declare namespace ಠ_ಠ.clutz.module$exports$mising$extend {
   }
   //!! This emit is wrong and will be fixed with
   //!! https://github.com/angular/clutz/issues/551.
```
mprobst: fix comment?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: Looks like Martin did a pretty deep review, I would only add guard the new AST parsing behind the partial flag so that it doesn't affect current performance.

src/test/java/com/google/javascript/clutz/partial/missing_base.js
mprobst: > 4. We already have a test for referring to a type by its full name (`direct.ref.A`).

This is what I meant:

```js
// file A
goog.provide('ns.A');
/** @constructor */
ns.A = function() {...};

// file B
goog.module('mod');
const A = goog.require('ns.A');
```

Now if you compile file B in isolation, I think this code emits a bunch of `module$something` strings that might not match up with file A produces when compiled in isolation (as there's no module stuff involved there).

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1669,6 +1674,11 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
         return;
       }
       String displayName = type.getDisplayName();
+      // In partial mode, closure doesn't know the correct name of imported symbols, if the name
+      // matches one in the precomputed map, replace it with the original declared name
+      if (importRenameMap.containsKey(displayName)) {
```
mprobst: noResolvedType can also happen in non-incremental mode (if code is incorrect/missing deps), which would then cause an `NPE` here. Maybe always initialize the map?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
+package com.google.javascript.clutz;
+
+import com.google.javascript.jscomp.Compiler;
+import com.google.javascript.jscomp.SourceFile;
+import com.google.javascript.rhino.Node;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ImportRenameMapBuilder {
```
mprobst: docs!

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+  public Map<String, String> buildImportRenameMap(
```
mprobst: docs!

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+    Node expression = statement.getFirstChild();
+    return expression.isCall() && expression.getFirstChild().matchesQualifiedName("goog.module");
+  }
+
+  private boolean isGoogRequireAssignment(Node statement) {
+    if (!(statement.isConst() || statement.isVar() || statement.isLet())) {
+      return false;
+    }
+
+    if (statement.getFirstFirstChild() == null) {
```
mprobst: extract a local variable instead of calling it twice?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+    for (Node statement : moduleBody.children()) {
+      if (isGoogModuleCall(statement)) {
+        return statement.getFirstChild().getChildAtIndex(1).getString();
+      }
+    }
+    return null;
+  }
+
+  /**
+   * If we're running in incremental mode, closure gives imported symbols names in the current
```
mprobst: nit: avoid "we" in docs, use `Clutz` or whatever appropriate.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+   * module's namespace, instead of the namespace of the module we're importing from. This reads the
```
mprobst: I think this text would be easier to understand with an example of the names.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+      return null;
+    }
+
+    Map<String, String> importRenameMap = new HashMap<>();
+    Node moduleBody = astRoot.getFirstChild();
+    for (Node statement : moduleBody.children()) {
+      if (isGoogRequireAssignment(statement)) {
+        String importedModuleId = statement.getFirstFirstChild().getChildAtIndex(1).getString();
+        String variableName = statement.getFirstChild().getString();
+        String symbolIdInCurrentModule =
```
mprobst: would it maybe be a better code organization to split this code into detection and extraction of goog.require in one piece, and then handling the require in a second piece? This here duplicates the string replace and `module$...` stuff in the two branches, and at the same time splits the knowledge about the AST structure of a `goog.require` between this code and the `isGoog...` methods above.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,128 @@
...
+          // generates, so we have to extract it.
+          String variableName;
+          if (destructure.getFirstChild() != null) {
+            variableName = destructure.getFirstChild().getString();
+          } else {
+            variableName = originalName;
+          }
+          String symbolIdInCurrentModule =
+              "module$contents$" + importingModuleId.replace(".", "$") + "_" + variableName;
+          String symbolIdInDefiningModule =
```
mprobst: To expand on my comment about the required test: I think `"module$exports$bla"` is the symbol that's produced by a `goog.module`, but it is not the symbol that's produced by code using `goog.provide`. So if you have a `goog.module` that imports code using `goog.provide`, I think this bit here will to reference the right type.

The thing to test for is having multiple files that are compiled in incremental mode, and then piping the output to TypeScript compiler together.

src/test/java/com/google/javascript/clutz/partial/missing_base.d.ts
```diff
@@ -7,14 +7,22 @@ declare namespace ಠ_ಠ.clutz.module$exports$mising$extend {
   }
   class BTemplated_Instance extends ಠ_ಠ.clutz.direct.ref.ATemplated < string , number > {
   }
-  class D extends D_Instance {
+  class ClassExtendingMissingDestructuredRequire extends ClassExtendingMissingDestructuredRequire_Instance {
   }
-  //!! This emit is wrong and will be fixed with
-  //!! https://github.com/angular/clutz/issues/551.
-  class D_Instance extends ಠ_ಠ.clutz.module$contents$mising$extend_C {
+  class ClassExtendingMissingDestructuredRequire_Instance extends ಠ_ಠ.clutz.module$exports$missing$base_MissingDestructuredRequire {
```
mprobst: For case 4) mentioned above, I think this import string will not match the type string generated by a goog.provide. I suspect `ಠ_ಠ.clutz.module$exports$missing$base_MissingDestructuredRequire` will be `missing.base.MissingDestructuredRequire`. If I'm right, you might need to change the producing side to make these match.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: I've extracted the string creators, but I don't see a really solid way to make it better beyond that.  The only thing that might be worth extracting is the for loop that grabs the destructured object names, and that'd still need a for loop, so I don't think it's worth it.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Done.

src/test/java/com/google/javascript/clutz/partial/missing_base.js
LucasSloan: Got it.  Talked it over with Rado, and there's currently no way to distinguish imports of goog.provide() files from goog.module().  We think it's outside the scope of this PR to fix that, so I've filed an issue https://github.com/angular/clutz/issues/596.

src/test/java/com/google/javascript/clutz/partial/missing_base.d.ts
LucasSloan: filed https://github.com/angular/clutz/issues/596

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: filed https://github.com/angular/clutz/issues/596

And agreed that we need tests for compiling the output of incremental mode.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: Ah I see. That's a bit surprising that you assign the field twice, in two different locations. I guess the alternative (passing it into `ImportRenameMapBuilder`) isn't worth it.

Can you use `Collections.emptyMap()` up there to make it a bit more obvious that it is never used (and prevent accidental use)?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -7,7 +7,19 @@
...
+ * reference the original namespace (ie `module$exports$imported$file`). ImportRenameMapBuilder
+ * builds a map from the local name to the exported name so the original names can be substituted.
+ */
 public class ImportRenameMapBuilder {
+
+  /**
+   * Takes a collection of source files, parses them with the compiler, and walks the ast to find
+   * any `const variable = goog.require()` statements to build a map from the name closure generates
+   * in incremental mode to the exported name of the import.
+   */
```
mprobst: Can buildImportRenameMap be a static method? It's a bit of a code smell to call `new Foo().bar()` without using the instance at all. You could still construct an instance in here that uses a private ctor.

mprobst: `ImportRenameMapBuilder.buildImportRenameMap` is a bit of a mouth full and won't help readers much. What about just calling this `.build`?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -79,50 +88,71 @@ private String getGoogModuleId(Node astRoot) {
...
-    String abkljeri = astRoot.toStringTree();
+  private Map<String, String> buildImportRenameMap(String localModuleId, Node astRoot) {
     if (astRoot.getFirstChild() == null || !astRoot.getFirstChild().isModuleBody()) {
       return null;
     }
 
     Map<String, String> importRenameMap = new HashMap<>();
     Node moduleBody = astRoot.getFirstChild();
     for (Node statement : moduleBody.children()) {
+      // `const C = goog.require()`
```
mprobst: hint: move the docs into the if statement, then they get consistent indent

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -294,6 +294,8 @@ public static void main(String[] args) {
    */
   private final Set<String> typesUsed = new LinkedHashSet<>();
 
+  private Map<String, String> importRenameMap = new HashMap<>();
```
mprobst: guess what I'd like to see on this field? Starts with a `d` :-)

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+   * import, the symbol name is just the module name.
+   */
+  private String buildExportedSymbolName(String importedModuleId) {
+    return "module$exports$" + importedModuleId.replace(".", "$");
+  }
+
+  /**
+   * The exported symbol can take 2 forms - one where it refers to everything that the module
+   * exports and another where it refers to just one thing the module exports. If there's a
+   * destructuring import, the symbol name is the module name plus the original name.
```
mprobst: should this code somewhere reference the issue you filed?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+        }
+      }
+    }
+
+    return importRenameMap;
+  }
+
+  /**
+   * The exported symbol can take 2 forms - one where it refers to everything that the module
+   * exports and another where it refers to just one thing the module exports. If there's a normal
```
mprobst: it's not clear what a normal import here means. I think what you're trying to explain is the difference between an `exports = ...` vs `exports.foo = ...` export, right? If so, the terminology is named exports for the latter. For the former, there's no good single word as it doesn't have a match in ES6 modules.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -79,50 +88,71 @@ private String getGoogModuleId(Node astRoot) {
   }
 
   /**
-   * If we're running in incremental mode, closure gives imported symbols names in the current
-   * module's namespace, instead of the namespace of the module we're importing from. This reads the
-   * import statements to find the module id of the module that's being imported and builds a map
-   * from the names in the current module, to their correct counterparts.
+   * Does the actual work of walking over the AST, finding any goog.require() assignments or
```
mprobst: A good habit is starting comments on functions & methods with the name of the function, it sets the right voice for the sentence: `buildImportRenameMap does the actual...`.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -79,50 +88,71 @@ private String getGoogModuleId(Node astRoot) {
...
 
     return importRenameMap;
   }
+
+  /**
+   * The exported symbol can take 2 forms - one where it refers to everything that the module
+   * exports and another where it refers to just one thing the module exports. If there's a normal
+   * import, the symbol name is just the module name.
+   */
+  private String buildExportedSymbolName(String importedModuleId) {
```
mprobst: maybe use a different method name that reflects the situation it is used for, instead of overloading based on parameters?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: Done!

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: It sure can, and would have been my preference, but I was deferring to a preference you had expressed for instance methods over static methods.  Is that just a Java vs. Typescript thing?

LucasSloan: Sounds good!

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Done.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Added a TODO.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Rewrote it to reference the actual styles.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

LucasSloan:
PTAL.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Build takes a collection of source files, parses them with the compiler, and walks the ast to
+   * find any `const variable = goog.require()` statements to build a map from the name closure
+   * generates in incremental mode to the exported name of the import.
+   */
+  public static Map<String, String> build(Compiler compiler, Collection<SourceFile> sourceFiles) {
+    Map<String, String> importRenameMap = new HashMap<>();
+    for (SourceFile sf : sourceFiles) {
+      Node ast = compiler.parse(sf);
```
mprobst: parsing is the most expensive operation in Closure Compiler. Can you double check that this doesn't actually parse the file again, and maybe find a way to make that more obvious in the code?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
mprobst: In TS, I'd make this a top level function, not a static method.

mprobst:
LGTM overall, but please make sure you're not re-parsing sources, and add at least a comment (or find a way that makes it more obvious0.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: We were reparsing, but I've fixed that.

PR #592, opened 2017-11-23
Clarify build and usage in README

### LINKED ISSUE #589: Should DeclarationGenerator be public?
The README notes:

> You can use Clutz as a library from a Java based build, or execute the wrapper script produced by Gradle.

If the library should be usable from via Java, should `DeclarationGenerator` be made public?

whymarrh:
Closes #589

Clarifies that:
- the build needs to be from source (kinda obvious maybe)
- this isn't a Java library but instead a command-line tool

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ auto-complete) as if the imported code was written in TypeScript.
 
 ## Example usage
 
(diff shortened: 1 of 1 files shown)


mprobst:
Thanks!

PR #593, opened 2017-11-27
Update to README to include Unix-separator hint

Dan1ve:
When working with Windows and the Windows-style separators are used, the process fails with a rather unhelpful NPE, which took me a while to figure out. I updated the README to reflect this.


Dan1ve:
I signed it!

mprobst:
I wonder if we could alternatively fix the issue? Is this in gents, or in Clutz? Under what circumstances?

PR #603, opened 2017-12-05
mark partial overridden function parameters as optional

### LINKED ISSUE #552: Incremental clutz makes optional args non-optional when types left out due to @override
When incremental clutz goes over b.js

```javascript
class C extends B {
  /** @override */
  foo(a, b, c) {...}
}
```

it emits 

```typescript
declare class C extends B {
  foo(a: any, b: any, c: any);
}
```

However, if 'B' looks like:

```javascript
class B {
  /**
    * @param {string=} a
    * @param {string=} b
    * @param {string=} c
    */
  foo(a, b, c) {...}
}
```

Clutz invariably made optional arguments, non-optional, thus breaking consumers.

In this case there is no harm in skipping emit for simple @overrides since they are picked up from the base class.

evmar:
When a method @override's a superclass method, and we can't see the
superclass due to partial mode, we can't know whether the parameters
are optional or not.  Try to detect when this happens (when all the
parameters are type {?}) and defensively mark them all as optional.

This change is walking a fine line between multiple competing concerns.
For example, we don't want to rely on specifically @override because
of @inheritDoc, and we also want to do the right thing when the
superclass is visible.  See the test case for some examples.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2624,6 +2624,16 @@ class DeclarationGenerator {
       emit("}");
     }
 
+    private boolean allParametersUnknown(FunctionType ftype) {
+      for (Node param : ftype.getParameters()) {
+        JSType type = param.getJSType();
+        // Note: template types (e.g. the T in Array<T>) return true for isUnknownType,
+        // so we check that first.
+        if (type.isTemplateType() || !type.isUnknownType()) return false;
+      }
+      return true;
+    }
+
     private void visitFunctionParameters(FunctionType ftype) {
       visitFunctionParameters(ftype, true, Collections.<String>emptySet());
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2633,6 +2641,10 @@ private void visitFunctionParameters(
       if (emitTemplatizedTypes) {
         visitTemplateTypes(ftype, alreadyEmittedTemplateType, true);
       }
+      // In partial mode when all the parameters to the function are unknown, it might be the
+      // case that the function is overriding/implementing a superclass that is in an unseen input.
```
rkirov: superclass or interface (for the next soul that thinks they can be smart about this).

src/test/java/com/google/javascript/clutz/partial/override.js
```diff
@@ -0,0 +1,70 @@
...
+ * @template T
+ */
+function Template() {}
+
+/**
+ * The type of T in the callback should not be marked optional.
+ * @param {function(T): R} f
+ * @template R
+ */
+Template.prototype.callback = function(f) {}
```
rkirov: How about
`@param {function(?): void} f`
should we emit
callback(f: (a: any) => void)
or
callback(f: (a?: any) => void)

I think your change does 1, but we shuold track it if it creates problems down the road. We can try to be smart and know about emitting class methods vs random types.

PR #604, opened 2017-12-05
Add Date to the set of Global symbols.

### LINKED ISSUE #602: Incremental Clutz refers to wrong Date in DateLike
DateLike looks like:

```js
goog.forwardDeclare('goog.date.Date');

/**
 * @typedef {(Date|goog.date.Date)}
 */
goog.date.DateLike;
```

The reference to `Date` there is attempting to reference the global Date, but it ends up picking up the forward-declared `goog.date.Date` from the current namespace.

LucasSloan:
Added a test to repro issue with DateLike declaration.  Fixed by introducing the indirection of GlobalDate by adding Date to the list of Global symbols, and adding GlobalDate to closure.lib.d.ts.

Fixes #602

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -227,9 +227,27 @@ class DeclarationGenerator {
    * </pre>
    *
    * by replacing the second Error with GlobalError.
+   *
+   * <p>Closure has internal "externs" for some global symbols:
+   * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/rhino/jstype/JSTypeRegistry.java
+   * Which will be emitted in each file that references them in partial mode. For example, using
+   * Date in a file results in:
+   *
+   * <pre>
+   * declare namespace ಠ_ಠ.clutz {
+   *   class Date extends Date_Instance {
+   *   }
+   *   class Date_Instance {
+   *     private noStructuralTyping_: any;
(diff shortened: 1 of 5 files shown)


src/test/java/com/google/javascript/clutz/partial/date_like.js
```diff
@@ -0,0 +1,12 @@
+goog.module('date.like')
+goog.forwardDeclare('goog.date.Date');
```
rkirov: Add a comment with `//!!` that forward declare is there to match real issue in closure/date_like, but not fully relevant to the test. I.e. it should work the same with a regular goog.require.

rkirov: //!! comments are ignored during golden checking.

LucasSloan:
PTAL

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -227,9 +227,14 @@
    * </pre>
    *
    * by replacing the second Error with GlobalError.
+   *
+   * <p>Closure has internal "externs" for some of these symbols:
+   * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/rhino/jstype/JSTypeRegistry.java
+   * Adding a symbol to this list has the side effect of suppressing an unnecessary declaration of
+   * the global type in each file that references it in partial mode.
```
rkirov: Hmm, I couldn't quite follow that sentence. In my mind, 'adding a symbol to this list also allows for accessing global symbols that have been shadowed'.

Even better if you add an example like the Error above.

PR #605, opened 2017-12-06
hack around missing base class in partial mode for goog.deferred.Async

evmar:
This class @extends Thenable, which means it should inherit a TTE type for
its .then() implementation.  But in partial mode, we don't see the base
class so we can't see that type.

We considered making it just infer to some broad any-based type, but in
in this particular case (the implementation of .then), if you lose the
TTE types then you have to manually annotate all the parameters to your
callback, so we deemed it ok to have a specific hack to see how common
this problem is.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2446,7 +2446,23 @@ class DeclarationGenerator {
           break;
         }
       }
-      if (!hasTTE) return false;
+
+      // Horrible hack.  goog.async.Deferred has an @override of a TTE function, but when running
+      // in partialInput mode we can't see that.  Identify it by grabbing:
+      // 1) functions named .then()
+      // 2) that use @override
+      // 3) that have no declared parameters/return type.
+      boolean horribleHackForPartialModeWithOverrides = false;
+      JSDocInfo info = type.getJSDocInfo();
+      if (info != null) {
+        boolean isUntypedOverride =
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2446,7 +2446,22 @@ private boolean handleSpecialTTEFunctions(
...
         }
       }
-      if (!hasTTE) return false;
+
+      // Horrible hack.  goog.async.Deferred has an @override of a TTE function, but when running
+      // in partialInput mode we can't see that.  Identify it by grabbing:
+      // 1) functions named .then()
+      // 2) that use @override
+      // 3) that have no declared parameters/return type.
+      boolean horribleAsyncHack = false;
```
rkirov: sounds like the hack is async. `horribleHackForPartialModeWithOverrides`

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2446,7 +2446,22 @@ private boolean handleSpecialTTEFunctions(
...
+      JSDocInfo info = type.getJSDocInfo();
+      if (info != null) {
+        boolean isUntypedOverride = info.isOverride() && info.getParameterCount() == 0 && info.getReturnType() == null;
+        if (opts.partialInput && info != null && isUntypedOverride && propName.equals("then")) {
```
rkirov: drop info != null in here

PR #606, opened 2017-12-06
Don't emit ಠ_ಠ.clutz for missing symbols.

LucasSloan:
Since we declare everything in namespaces that start with ಠ_ಠ.clutz, any symbols that are under the ಠ_ಠ.clutz namespace will be resolved there, but if the missing symbols are in the platform externs (eg Node), then they won't be shadowed by the ಠ_ಠ.clutz namespace, as they were previously.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1682,7 +1682,7 @@ class DeclarationGenerator {
       if (importRenameMap.containsKey(displayName)) {
         displayName = importRenameMap.get(displayName);
       }
-      emit(Constants.INTERNAL_NAMESPACE + "." + displayName);
+      emit(displayName);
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1682,7 +1682,7 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
       if (importRenameMap.containsKey(displayName)) {
         displayName = importRenameMap.get(displayName);
       }
-      emit(Constants.INTERNAL_NAMESPACE + "." + displayName);
+      emit(displayName);
```
rkirov: Leave some trace in comment why we switched to 'display name' vs 'Constants.INTERNAL_NAMESPACE + "." + displayName.'

rkirov:
Added a comment and merged this separetely.

PR #607, opened 2017-12-06
Adds a test that matches the emit pattern of tsickle.

rkirov:
Tsickle has to do an odd workaround to append moduleId to each
goog.module.

Turns out this changes the clutz emit significantly, and we need to have
a test to detect changes in that emit.


src/test/java/com/google/javascript/clutz/goog_module_export_redefined.js
```diff
@@ -0,0 +1,10 @@
+goog.module('module.tsickle.like');
+
+//!! This odd pattern is what we do for all tsickle emitted .js.
+exports = {};
+exports.moduleId = 'foo/bar';
```
mprobst: Isn't it more like:

```
exports = {};
var module = {id: '...'};
```
?

Not sure if that's significant either way.

PR #608, opened 2017-12-07
Separate root and nonroot externs.

LucasSloan:


src/test/java/com/google/javascript/clutz/DepgraphTest.java
```diff
@@ -25,6 +25,6 @@ public void testParseFile() throws Exception {
     assertThat(depgraph.getNonroots())
         .containsExactly("javascript/closure/base.js", "javascript/closure/string/string.js")
         .inOrder();
-    assertThat(depgraph.getExterns()).containsExactly("javascript/common/dom.js").inOrder();
+    assertThat(depgraph.getNonrootExterns()).containsExactly("javascript/common/dom.js").inOrder();
```
rkirov: Add assertion on getRootExterns too

rkirov: Add a test for partialInput flag branch too.

src/main/java/com/google/javascript/clutz/Depgraph.java
```diff
@@ -38,7 +39,8 @@ Depgraph withNonrootsAsRoots() {
     Depgraph res = new Depgraph();
     res.roots.addAll(roots);
     res.roots.addAll(nonroots);
-    res.externs.addAll(externs);
+    res.rootExterns.addAll(rootExterns);
+    res.nonrootExterns.addAll(nonrootExterns);
```
mprobst: is this correct? it's oddly assymmetric.

src/test/java/com/google/javascript/clutz/DepgraphTest.java
LucasSloan: Done, done.

src/main/java/com/google/javascript/clutz/Depgraph.java
LucasSloan: @rkirov 

I've looked over the code, and currently nothing depends on the state of `rootExterns` or `nonrootExterns` after `withNonrootsAsRoots()` is called.  I agree with Martin that the less surprising state of affairs in the future is that all the externs are in `rootExterns`.

LucasSloan:
I patched this into g3, and it did fix the error about files not being found.

src/main/java/com/google/javascript/clutz/Depgraph.java
rkirov: yep, should be `res.rootExterns.addAll(nonrootExterns)'. This branch code is only executed when 'strictDeps' is off, and we usually have it on.

PR #611, opened 2017-12-08
Add a multifile partial clutz test.

LucasSloan:
Adds a test case for compiling 3 javascript files (1 importing from the other 2) individually, then running a DeclarationSyntaxTest on the result to ensure that Clutz produces correct output.

As a result, an error in ImportRenameMapBuilder was discovered where the imported symbols were being rename to modulename *underscore* symbolname, instead of the correct modulename *dot* symbolname.


src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
```diff
@@ -97,8 +97,11 @@ static String getExternFileNameOrNull(String testFileName) {
     File[] testFiles = getPackagePath().toFile().listFiles(filter);
     // Partial files live in 'partial' dir and run implicitly with the --partialInput option on.
     File[] testPartialFiles = getPackagePath().resolve("partial").toFile().listFiles(filter);
+    File[] testMultifilePartailFiles =
```
rkirov: Add a comment describing the purpose of multifile partial, similar to the one above. Call out specifically that each .js creates its own golden test *and* we do a singleton test on top.

src/test/java/com/google/javascript/clutz/MultiFileTest.java
```diff
@@ -83,6 +83,15 @@ public void skipEmitWithTypedefs() throws Exception {
         .generatesDeclarations(expected);
   }
 
+  @Test
+  public void multifilePartial() throws Exception {
+    String expected = DeclarationGeneratorTests.getTestFileText(input("singleton.d.ts"));
```
rkirov: nitpicing on the name: 'singleton' in a java code base might bring unwanted associations. How about 'total', that's my new word for old clutz mode.

src/test/java/com/google/javascript/clutz/multifilePartial/singleton.d.ts
```diff
@@ -0,0 +1,44 @@
+//!! The singleton file is the result of a global compilation of the files in
```
rkirov: global -> total

src/test/java/com/google/javascript/clutz/multifilePartial/singleton.d.ts
```diff
@@ -0,0 +1,44 @@
...
+//!! the multifilePartial directory, so that the global and partial compilations
+//!! can be compared
```
rkirov: ok for now, but long term this file will be less and less useful (if we remove total mode one day). An even better test (in another PR) would a usage.ts file that typechecks together with the three partial .d.ts.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -152,7 +152,7 @@ private static String buildWholeModuleExportSymbolName(String importedModuleId)
    * https://github.com/angular/clutz/issues/596
    */
   private static String buildNamedExportSymbolName(String importedModuleId, String originalName) {
-    return "module$exports$" + importedModuleId.replace(".", "$") + "_" + originalName;
+    return "module$exports$" + importedModuleId.replace(".", "$") + "." + originalName;
```
rkirov: does this need a change in tsickle?

LucasSloan: Yes - https://github.com/angular/tsickle/pull/671

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
LucasSloan: Added!

src/test/java/com/google/javascript/clutz/MultiFileTest.java
LucasSloan: Huh - what does singleton mean in java?

And changed.

src/test/java/com/google/javascript/clutz/multifilePartial/singleton.d.ts
LucasSloan: Done.

src/test/java/com/google/javascript/clutz/multifilePartial/singleton.d.ts
LucasSloan: That makes sense!

LucasSloan:
PTAL.

PR #613, opened 2017-12-09
Attach the Global* emit logic for missing type too

rkirov:
We already emit Global* for some global types that can get shadowed by
other namespace types. For example

goog.events.EventTarget;

/** @param {EventTarget} */
goog.events.listen

cannot be

namespace goog.events {
  interface EventTarget {}
  listen(t: EventTarget) {}
}

because the fn argument refers to the global one. Instead we have
GlobalEventTartget alias the global EventTarget and emit that name
instead.

With incremental clutz, we have to add that logic to the
emitNoResolvedType function too.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1685,6 +1685,7 @@ class DeclarationGenerator {
         emit(defaultEmit);
         return;
       }
+      if (maybeEmitGlobalAlias(type)) return;
       // When processing partial inputs, this case handles implicitly forward declared types
       // for which we just emit the literal type written along with any type parameters.
       NoResolvedType nType = (NoResolvedType) type;
@@ -2762,10 +2763,7 @@ class DeclarationGenerator {
 
     public Void emitObjectType(
         ObjectType type, boolean extendingInstanceClass, boolean inExtendsImplementsPosition) {
-      if (type.getDisplayName() != null && GLOBAL_SYMBOL_ALIASES.contains(type.getDisplayName())) {
-        emit("Global" + type.getDisplayName());
-        return null;
-      }
+      if (maybeEmitGlobalAlias(type)) return null;
       // Closure doesn't require that all the type params be declared, but TS does
       if (!type.getTemplateTypeMap().isEmpty()
           && !typeRegistry.getNativeType(OBJECT_TYPE).equals(type)) {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/partial/global_alias_missing.js
```diff
@@ -0,0 +1,6 @@
+goog.provide('ns.event');
+
+//!! Note that emitting just EventTarget would be wrong here
+//!! Assume there exists a ns.event.EventTarget.
```
LucasSloan: Is ns.event.EventTarget defined in another file?

rkirov: reworded the comment. PTAL

src/test/java/com/google/javascript/clutz/partial/global_alias_missing.js
```diff
@@ -0,0 +1,8 @@
+goog.provide('ns.event');
+
+//!! Assume there exists a ns.event.EventTarget, then
+//!! emitting just t: EventTarget would be wrong.
+//!! TypeScript will pick up ns.event.EventTarget instead
+//!! of the global EventTarget.
+/** @param {EventTarget} t */
```
LucasSloan: I still don't understand.  Why doesn't `EventTarget` refer to the local value?

Could you add a contrasting function that does refer to `ns.event.EventTarget`?  Or at least show how the local value would be defined?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1700,6 +1700,7 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
       if (importRenameMap.containsKey(displayName)) {
         displayName = importRenameMap.get(displayName);
       }
+      if (maybeEmitGlobalAlias(type)) return;
```
LucasSloan: Move this up to above `String displayName = type.getDisplayName();` - let's keep all the early return guards at the top.

src/test/java/com/google/javascript/clutz/partial/global_alias_missing.js
rkirov: Good idea, added f2 showing how namespace local Event target would look like

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: moved up.

rkirov:
PTAL

LucasSloan:
LGTM, but I notice the CI is red.

PR #618, opened 2017-12-13
Add a partial_goog_base.js file that is used for all incremental clutz.

rkirov:
Knowing that 'goog.abstractMethod' is a function allows for incremental
clutz to emit 'C.prototype.fn = goog.abstractMethod' as a method and not
a field. TypeScript asserts that methods can only override other
methods, so this is crucial knowledge for incremental clutz.

To achieve this we add an extra mock version of base.js -
mock_goog_base.js that will be passed to every partial clutz
compilation. It will not affect the output in other ways than allowing
inference works to work better in its presence.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/ProgramSubject.java
+++ b/src/test/java/com/google/javascript/clutz/ProgramSubject.java
@@ -26,9 +26,20 @@ import javax.annotation.Nullable;
 /** A subject that supports assertions on {@link DeclarationGenerator}'s results. */
 class ProgramSubject extends Subject<ProgramSubject, ProgramSubject.Program> {
 
-  /** A stripped down version of Closure's base.js for Clutz tests. */
-  private static final SourceFile CLUTZ_GOOG_BASE =
+  /**
+   * A stripped down version of Closure's base.js for Clutz tests. In real total clutz runs we
+   * always pass the real closure's base.js.
+   */
+  private static final SourceFile CLUTZ_GOOG_BASE_TOTAL =
       SourceFile.fromFile("src/test/java/com/google/javascript/clutz/base.js", UTF_8);
+
+  /**
+   * A even more stripped down version of Closure's base.js for incremental Clutz tests. In
+   * production we also use this file for every incremental run, because the original base.js is not
+   * passed in.
+   */
+  private static final SourceFile CLUTZ_GOOG_BASE_INCR =
+      SourceFile.fromFile("src/resources/partial_goog_base.js", UTF_8);
 
   public boolean withPlatform = false;
   public boolean partialInput = false;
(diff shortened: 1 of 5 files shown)


mprobst:
Terminology here might be confusing for future debuggers.

A mock is an implementation of some interface used in tests that has no
implementation logic and returns pre-configured values.
A fake is an implementation of some interface used in tests that has
implementation logic, but is not the real production logic.

Given that this is not something that's used in tests at all, maybe
something like `minimal_goog_base_for_incremental_clutz.js` would be a
better description of the file?

src/test/java/com/google/javascript/clutz/base.js
```diff
@@ -30,4 +35,4 @@ goog.isDef = function(val) {};
 
 goog.provide('goog.Uri');
 /** @constructor */
-goog.Uri =  function() {}
+goog.Uri =  function() {}
```
evmar: This edit looks unintentional.

src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -26,10 +26,22 @@
 /** A subject that supports assertions on {@link DeclarationGenerator}'s results. */
 class ProgramSubject extends Subject<ProgramSubject, ProgramSubject.Program> {
 
-  /** A stripped down version of Closure's base.js for Clutz tests. */
-  private static final SourceFile CLUTZ_GOOG_BASE =
+  /**
+   * A stripped down version of Closure's base.js for Clutz tests.
+   * In real total clutz runs we always pass the real closure's base.js
```
evmar: trailing period

src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -26,10 +26,22 @@
...
+   */
+  private static final SourceFile CLUTZ_GOOG_BASE_TOTAL =
       SourceFile.fromFile("src/test/java/com/google/javascript/clutz/base.js", UTF_8);
 
+  /**
+   *  A even more stripped down version of Closure's base.js for incremental
+   *  Clutz tests. In production we also use this file for every incremental run,
+   *  because the original base.js is not passed it.
```
evmar: typo: in

src/test/java/com/google/javascript/clutz/base.js
evmar: fixme

src/test/java/com/google/javascript/clutz/partial/abstract_method_partial.js
```diff
@@ -0,0 +1,19 @@
+goog.module('abst.method');
+
+/** @constructor */
+let C = function () {
+
+};
+
+/**
+ * It appears that this one is emitted correctly without partial_goog_base.js
```
evmar: trailing period, and rebaseline test.

src/test/java/com/google/javascript/clutz/partial/abstract_method_partial.js
```diff
@@ -0,0 +1,19 @@
...
+ * @return {number}
+ */
+C.prototype.methodWithTypes = goog.abstractMethod;
+
+/**
+ * But this one needs partial_goog_base.js
```
evmar: here too

src/test/java/com/google/javascript/clutz/ProgramSubject.java
rkirov: done

src/test/java/com/google/javascript/clutz/ProgramSubject.java
rkirov: done

PR #619, opened 2017-12-13
Fix import renaming for es6 class syntax.

LucasSloan:
Using es6 syntax instead of /** @constructor */ annotations results in a different display name for a missing type in partial mode.  In an es6 the type is `foo` instead of the annoation method's `module$contents$foo`.  This PR modifies ImportRenameMapBuilder to produce mappings for both types of display name.

This PR also fixes an issue when importing an object, which contains the type to be used (ie imported foo, but used as `class quux extends foo.bar`.  In that case the displayname is of the form `foo.bar` or `module$contents$foo.bar`, but the importRenameMap only has a rename for the base of the display name.  To fix, the logic is modified to look for replacements of the portion of the display name up to the first period (if any).

### DIFF
--- a/src/test/java/com/google/javascript/clutz/multifilePartial/default_object_exporter.js
+++ b/src/test/java/com/google/javascript/clutz/multifilePartial/default_object_exporter.js
@@ -1,0 +1,11 @@
+goog.module('default.object.exporter');
+exports = {};
+
+/**
+ * @constructor
+ */
+function BaseClass() {
+
+}
+
+exports.BaseClass = BaseClass;
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1698,8 +1698,11 @@ class DeclarationGenerator {
       String displayName = type.getDisplayName();
       // In partial mode, closure doesn't know the correct name of imported symbols, if the name
       // matches one in the precomputed map, replace it with the original declared name
-      if (importRenameMap.containsKey(displayName)) {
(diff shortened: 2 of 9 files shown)


src/test/java/com/google/javascript/clutz/multifilePartial/default_object_exporter.js
```diff
@@ -0,0 +1,10 @@
+goog.module('default.object.exporter');
+
+/**
+ * @constructor
+ */
+function BaseClass() {
+
+}
+
+exports = {BaseClass};
```
rkirov: Just `exports.BaseClass = BaseClass`, I think that's more common.

src/test/java/com/google/javascript/clutz/partial/test.js
```diff
@@ -0,0 +1,32 @@
+goog.module("missing.imported.base");
```
rkirov: This file needs a better name than 'test.js'. I think it is testing a goog.require aliasing inside a goog.module.

src/test/java/com/google/javascript/clutz/multifilePartial/default_object_exporter.js
LucasSloan: Rewrote it to look more like the output from tsickle.

src/test/java/com/google/javascript/clutz/partial/test.js
LucasSloan: I've renamed it to `es6_missing_imported_base.js` to reflect that it's the same as `missing_imported_base.js`, but using es6 classes instead of constructor annoations.

PR #621, opened 2017-12-14
Rename missing imported symbols in goog.scope.

LucasSloan:
Similar to having to rename symbols that are imported in goog.modules(), closure doesn't get the right name for symbols imported inside a goog.scope block.  This PR modifies ImportRenameMapBuilder to collect the rename mappings inside goog.scope blocks, in addition to the existing goog.modules.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
+++ b/src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
@@ -1,34 +1,65 @@
 package com.google.javascript.clutz;
 
 import com.google.javascript.rhino.Node;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
  * If Clutz is running in incremental mode, closure can't resolve imported symbols, so it gives them
- * names in the current namespace (ie `module$contents$current$file`), instead of names that
- * reference the original namespace (ie `module$exports$imported$file`). ImportRenameMapBuilder
- * builds a map from the local name to the exported name so the original names can be substituted.
+ * names based on the symbol that the import was assigned to (either
+ * `module$contents$current$file_symbol` or just `symbol`) instead of names that reference the
+ * original namespace (ie `module$exports$imported$file`). ImportRenameMapBuilder builds a map from
+ * the local name to the exported name so the original names can be substituted.
  */
 public class ImportRenameMapBuilder {
 
   /**
-   * Build takes a collection of source files, parses them with the compiler, and walks the ast to
-   * find any `const variable = goog.require()` statements to build a map from the name closure
-   * generates in incremental mode to the exported name of the import.
+   * Build takes a collection of parsed inputs and walks the ast to find any imports into local
+   * variables to build a map from the name closure generates in incremental mode to the exported
+   * name of the import.
    */
   public static Map<String, String> build(Collection<Node> parsedInputs) {
     Map<String, String> importRenameMap = new HashMap<>();
     for (Node ast : parsedInputs) {
+      // Symbols can be imported into a variable in a goog.module() file, so look for imports in the
+      // body of the goog module.
       String moduleId = getGoogModuleId(ast);
-      // Closure can only put the symbol name into the current module's namespace, if the source
-      // file is a module, so if it isn't a module, just bail
       if (moduleId != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -1,36 +1,67 @@
...
+      // body of the goog module.
       String moduleId = getGoogModuleId(ast);
-      // Closure can only put the symbol name into the current module's namespace, if the source
-      // file is a module, so if it isn't a module, just bail
       if (moduleId != null) {
-        importRenameMap.putAll(build(moduleId, ast));
+        importRenameMap.putAll(build(moduleId, ast.getFirstChild()));
+      }
+
+      // Or symbols can be imported into a variable in a goog.scope() block, so look for imports in
```
rkirov: top-level goog.scope block, techincally goog.scope can be used anywhere, but we dont want to count those.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -1,36 +1,67 @@
...
+      if (!googScopes.isEmpty()) {
+        for (Node googScope : googScopes) {
+          importRenameMap.putAll(build(null, googScope));
+        }
       }
     }
     return importRenameMap;
   }
 
+  private static List<Node> getGoogScopes(Node astRoot) {
```
rkirov: getTopLevelGoogScopes, at least I think that's what this method does.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -49,10 +80,11 @@ private static boolean isGoogRequireAssignment(Node statement) {
 
     return rightHandSide != null
         && rightHandSide.isCall()
-        && rightHandSide.getFirstChild().matchesQualifiedName("goog.require");
+        && (rightHandSide.getFirstChild().matchesQualifiedName("goog.require")
```
rkirov: is goog.require in goog.scope even a thing? Just drop it for now. I haven' seen such a pattern in g3.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -121,10 +155,13 @@ private static String getGoogModuleId(Node astRoot) {
             variableName = originalName;
           }
 
-          String localSymbolName = buildLocalSymbolName(localModuleId, variableName);
           String exportedSymbolName = buildNamedExportSymbolName(importedModuleId, originalName);
-          importRenameMap.put(localSymbolName, exportedSymbolName);
           importRenameMap.put(variableName, exportedSymbolName);
+          // If we're in a goog scope, there isn't a module id
+          if (localModuleId != null) {
```
rkirov: looks like a duplicated block, can it be outside the if-else?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Ok.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Sure.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: This method matches either a `goog.require` or a `goog.module.get`.  Which it will find depends on whether a goog.module body was passed to `build` or a goog.scope body.  I've added comments.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: I don't see a way - the two blocks are subtly different.

LucasSloan:
PTAL.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
rkirov: I see, 'goog.require' was there before. Makes sense.

PR #622, opened 2017-12-14
Fix emits in the goog namespace.

LucasSloan:
Clutz had the assumption that every namespace it emits is a subnamespace of the ಠ_ಠ.clutz namespace.  This caused 2 problems:

1. Clutz refered to other types within the goog namespace with the ಠ_ಠ.clutz prefix.
2. Clutz refered to types in the ಠ_ಠ.clutz namespace within the goog namespace without the ಠ_ಠ.clutz prefix.

This PR fixes the above by only ever refering to a symbol inside a namespace with the ಠ_ಠ.clutz prefix when refering to a global symbol (ie in closure.lib.d.ts) while in the goog namespace.  In every other case, Clutz can rely on Typescript looking for the symbol in the parent namespaces if the prefix is omitted.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -595,7 +595,7 @@ class DeclarationGenerator {
     emit(getUnqualifiedName(typedefName));
     emit("=");
 
-    new TreeWalker(compiler.getTypeRegistry(), provides, false).visitType(moduleType);
+    new TreeWalker(compiler.getTypeRegistry(), provides, false, false).visitType(moduleType);
     emit(";");
     emitBreak();
     emitNamespaceEnd();
@@ -650,7 +650,7 @@ class DeclarationGenerator {
           emit(":");
           TypedVar var = topScope.getOwnSlot(reservedProvide);
           if (var != null) {
-            TreeWalker walker = new TreeWalker(compiler.getTypeRegistry(), provides, false);
+            TreeWalker walker = new TreeWalker(compiler.getTypeRegistry(), provides, false, false);
             walker.visitType(var.getType());
           } else {
             emit("any");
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1287,10 +1292,17 @@ private String lastDottedPart(String input) {
...
     private final boolean isExtern;
 
-    private TreeWalker(JSTypeRegistry typeRegistry, Set<String> provides, boolean isExtern) {
+    private final boolean isGoogNamespace;
+
+    private TreeWalker(
+        JSTypeRegistry typeRegistry,
+        Set<String> provides,
+        boolean isExtern,
+        boolean isGoogNamespace) {
```
rkirov: We need one location in the code to explain what's special about goog base. Maybe this is that place.

src/test/java/com/google/javascript/clutz/base.js
```diff
@@ -31,3 +31,21 @@ goog.isDef = function(val) {};
 goog.provide('goog.Uri');
 /** @constructor */
 goog.Uri =  function() {}
+
+/**
+ * @param {Object} object Object whose static property is being exported.
+ */
+goog.functionUsingGlobalObject = function(object) {
```
rkirov: given that this is supposed to mirror base, why not just put the actual function that uses Object from goog base.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: Added a comment - does it work?

src/test/java/com/google/javascript/clutz/base.js
LucasSloan: I've replaced my abstractions with the actual versions.

LucasSloan:
PTAL

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -989,7 +995,7 @@ private void declareNamespace(
 
       ObjectType oType = symbol.getType().toMaybeObjectType();
 
-      if (oType != null) {
+      if (oType != null && !isGoogNamespace) {
```
rkirov: add a comment here, that we skip this pass for goog namespace because of global oddity. I think we should come back to this one day.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: looks good

PR #632, opened 2017-12-20
Add a null check in ModuleImport.

voithos:
The previous code immediately tried to use the retrieved module, which
is sometimes `null` (either because a dependency is missing from the
manifest, or perhaps also for newly-converted namespaces). There is
already an additional check in `validateImport` that provides much
better error output (which is now reachable).


voithos:
Sure - it hits this check, I believe: https://github.com/angular/clutz/blob/master/src/main/java/com/google/javascript/gents/ModuleConversionPass.java#L381

Not sure if there's a better way to structure things to make it more clear. The error message could potentially use a bit more elaboration (maybe something about checking the manifest you passed in, etc).

PR #635, opened 2017-12-22
Make gents not throw on missing rewritelog option.

### LINKED ISSUE #633: Gents crashes on hello world
I originally tried gents on my Closure-annotated project and it crashed on the first line of code. I tried to create a minimal reproduction, and ended up with this. I'm running OSX 10.11.6, Java 1.8.0_121.

```
Robin$ cat test.js
console.log('hello world!');
Robin$ 
Robin$ gents test.js
java.lang.NullPointerException
	at java.io.File.<init>(File.java:277)
	at com.google.javascript.gents.TypeScriptGenerator.generateTypeScript(TypeScriptGenerator.java:147)
	at com.google.javascript.gents.TypeScriptGenerator.main(TypeScriptGenerator.java:75)
Uncaught exception in gents, exiting.
```

Interestingly, if I specify `-o somedir`, correct output is still written to that directory. However, that seems to only work for some files -- the files in my project don't produce any output at all.

rkirov:
Just skip emitting if the user does not provide that option.

Closes #633

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
+++ b/src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
@@ -144,7 +144,9 @@ public class TypeScriptGenerator {
       }
     }
     try {
-      Files.write(result.moduleRewriteLog, new File(opts.moduleRewriteLog), UTF_8);
+      if (opts.moduleRewriteLog != null) {
+        Files.write(result.moduleRewriteLog, new File(opts.moduleRewriteLog), UTF_8);
+      }
     } catch (IOException e) {
       throw new IllegalArgumentException("Unable to write to file " + opts.moduleRewriteLog, e);
     }


src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -144,7 +144,10 @@ private void generateTypeScript() {
       }
     }
     try {
-      Files.write(result.moduleRewriteLog, new File(opts.moduleRewriteLog), UTF_8);
+      if (opts.moduleRewriteLog != null) {
+        Files.write(result.moduleRewriteLog, new File(opts.moduleRewriteLog), UTF_8);
+      }
+      // Skip writing if moduleRewriteLog is not used, instead of trying to guess a place to write to.
```
evmar: I think the comment isn't necessary.  If the user doesn't say where to write an extra file to, it's usually implied it won't be written.

rkirov: I am just never sure what are the expectations with CLI tools. Removed

PR #637, opened 2017-12-22
Gents translates ES6 classes with @interface annotation to interfaces.

### LINKED ISSUE #630: gents skips extends on a class-interface object
```javascript
/**
 * @interface
 * @extends {A}
 */
class B {}
```

emits just `class B`, no extends.

rkirov:
Some closure users prefer to write /** @interface */ class IFoo, instead
of the more common function syntax. The intent is nevertheless for this
to be a pure interface and not class, so gents will correspondingly
convert that.

Some complications arrise from adding @extends and ES6's class extend
keyword and method/field definitions.

Closes #630,627


rkirov:
Can I also let it be known that closure's untyped AST takes forever to work with. Spent hours staring at NPEs and stacktraces.

PR #641, opened 2017-12-23
Add a missing break statement in case block.

rkirov:


rkirov:
submitting TBR, this is trivial.

rkirov:
some java presubmit forbid me from checking this in :!

PR #644, opened 2017-12-28
Special case log.js's class aliases.

LucasSloan:
Closure's log.js has this pattern:

/**
 * @constructor
 * @final
 */
goog.log.Level = goog.debug.Logger.Level;

Where the goog.debug.Logger.Level isn't part of the partial compilation.  Closure gives us a type based only on the type comment, totally ignoring the aliased symbol, so there's no easy way to solve the issue in clutz without a closure change.  This pattern only seems to show up in log.js, so just add the relevant symbols to partial_goog_base.js, which causes closure to give a type that knows about the alias.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/partial/missing_logger_dot_level.js
+++ b/src/test/java/com/google/javascript/clutz/partial/missing_logger_dot_level.js
@@ -1,0 +1,11 @@
+goog.provide('goog.log.Level');
+
+goog.require('goog.debug.Logger');
+
+//!! goog.log.Level should be an alias for goog.debug.Logger.Level, not an empty
+//!! class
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/clutz/partial/missing_logger_dot_level.js
```diff
@@ -0,0 +1,10 @@
+goog.provide('goog.log.Level');
+
+goog.require('goog.debug.Logger');
+
+
```
evmar: Can you add a comment here about the sort of output you expect?  (That is, that Level is an alias for Logger.Level in the d.ts.)

PR #652, opened 2018-01-04
Enable goog.module and goog.provide files to interop in partial mode.

### LINKED ISSUE #596: Incremental Clutz can't distinguish imports from goog.module() files and goog.provide() files.
A goog.provide file:
```javascript
goog.provide('goog.example');
/** @constructor */
goog.example = function() {};
```
Produces this .d.ts:
```typescript
declare namespace ಠ_ಠ.clutz.goog {
  class example extends example_Instance {
  }
  class example_Instance {
  }
}
declare module 'goog:goog.example' {
  import alias = ಠ_ಠ.clutz.goog.example;
  export default alias;
}
```
A goog.module file:
```javascript
goog.module("goog.example");
/** @constructor */
const A = function() {};
exports.A = A;
```
Produces this .d.ts:
```typescript
declare namespace ಠ_ಠ.clutz.module$exports$goog$example {
  class A extends A_Instance {
  }
  class A_Instance {
  }
}
declare module 'goog:goog.example' {
  import alias = ಠ_ಠ.clutz.module$exports$goog$example;
  export = alias;
}
```

The goog.provide file generates the namespace `ಠ_ಠ.clutz.goog`, while the goog.module generates the namespace `ಠ_ಠ.clutz.module$exports$goog$example`.

A goog.module file could require `goog.example`:
```javascript
goog.module('importer');
const import = goog.require('goog.example');
```

In incremental mode, Clutz doesn't know if `goog.example` is from a goog.provide file or a goog.module file, so Clutz don't know to emit `ಠ_ಠ.clutz.goog` or `ಠ_ಠ.clutz.module$exports$goog$example`.

The two styles of namespace declaration need to be united, so incremental mode doesn't need to know about the style of its dependencies.

LucasSloan:
In a goog.module file, there's no indication if a symbol that's being goog.required() is from another goog.module file or a goog.provide file.  To compensate, clutz needs to rewrite goog.provided symbols to use the goog.module export style.  Complicating matters, the semantics of goog.provide/goog.require mean that requiring `foo` gives access to `foo.bar` and `foo.bar.baz`.  To enable that with the goog.module export style, each goog.provide symbol has to be emitted multiple times, so that each possible parent namespace has access to the symbol.

Fixes #596


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1174,6 +1211,11 @@ private void declareModule(
       // goog:goog cannot be imported.
       return;
     }
+    // In partial mode, emit clutz emits goog.provided symbols in goog.module/dollar style, so the
```
evmar: Typo: "emit clutz emits"

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1755,6 +1789,26 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
       }
     }
 
+    private String getDisplayNameFromType(JSType type) {
+      String displayName = type.getDisplayName();
+      // In partial mode, closure doesn't know the correct name of imported symbols, if the name
```
evmar: Sentence splice makes this a bit weird to read.  Maybe change hte second comma to a period?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1755,6 +1789,26 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
...
+      // matches one in the precomputed map, replace it with the original declared name
+      // The displayName can be of the form foo.bar, but the symbol that was goog required was just
+      // foo, so just replace the part of the display name before the first period
```
evmar: I didn't follow this example too well, because there's only one period in the name.  Maybe make it instead describe foo.bar.baz so you can see the difference between the periods?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1755,6 +1789,26 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
...
+      List<String> displayNameParts = new ArrayList<>(Arrays.asList(displayName.split("\\.")));
+      for (int i = 1; i <= displayNameParts.size(); i++) {
+        List<String> displayNamePrefix = displayNameParts.subList(0, i);
+        String baseDisplayName = Joiner.on(".").join(displayNamePrefix);
+        if (importRenameMap.containsKey(baseDisplayName)) {
+          displayNamePrefix.clear();
+          displayNamePrefix.add(importRenameMap.get(baseDisplayName));
```
evmar: This seems to do some work on displayNamePrefix but then never use it (?)

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -24,11 +24,14 @@
...
     for (Node ast : parsedInputs) {
-      // Symbols can be imported into a variable in a goog.module() file, so look for imports in the
-      // body of the goog module.
       String moduleId = getGoogModuleId(ast);
       if (moduleId != null) {
+        // Symbols can be imported into a variable in a goog.module() file, so look for imports in the
+        // body of the goog module.
         importRenameMap.putAll(build(moduleId, ast.getFirstChild()));
+      } else {
+        // Also get renames for symbols that are required, but not into a variable
```
evmar: I don't get "required, but not into a variable"

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -71,6 +74,27 @@ private static boolean isGoogModuleCall(Node statement) {
...
+      return false;
+    }
+
+    if (statement.getFirstChild() == null) {
+      return false;
+    }
+
+    Node googRequireCall = statement.getFirstChild();
+
+    return googRequireCall != null
```
evmar: You null-check the first child twice.  I think maybe you should drop the getFirstChild() == null "if" block.

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
```diff
@@ -77,7 +77,8 @@ public static TestSuite suite() throws IOException {
       if (input.getName().contains("_output_base")) {
         subject.emitBase = true;
       }
-      if (Arrays.asList("partial", "multifilePartial").contains(input.getParentFile().getName())) {
+      if (Arrays.asList("partial", "multifilePartial", "crossModuleImports")
```
evmar: It's kinda unpredictable from the outside that "crossModuleImports" implies "partial".  Maybe it should be "partialCrossModuleImports" or something?

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
```diff
@@ -107,10 +108,16 @@ static String getExternFileNameOrNull(String testFileName) {
         getPackagePath().resolve("multifilePartial").toFile().listFiles(filter);
     // Output base files live in the 'outputBase' dir and impilicitly have base.js in their roots
     File[] testOutputBaseFiles = getPackagePath().resolve("outputBase").toFile().listFiles(filter);
+    // Test files that live in the 'crossModuleImports' dir, and run with the --partialInput option
+    // The resulting .d.ts files are checked with a DeclarationSyntaxTest, and they're also
+    // compiled in a single run in MultiFileTest
```
evmar: Is this the reason to not put them alongside the other 'partial' tests -- that they're checked/compiled after?

src/test/java/com/google/javascript/clutz/crossModuleImports/goog_multi_provide_exporter.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+declare namespace ಠ_ಠ.clutz.module$exports$goog.multi.provide.exporter {
+  export import FirstMultiProvideClass = ಠ_ಠ.clutz.module$exports$goog$multi$provide$exporter$FirstMultiProvideClass;
+}
+declare namespace ಠ_ಠ.clutz.module$exports$goog$multi.provide.exporter {
+  export import FirstMultiProvideClass = ಠ_ಠ.clutz.module$exports$goog$multi$provide$exporter$FirstMultiProvideClass;
+}
+declare namespace ಠ_ಠ.clutz.module$exports$goog$multi$provide.exporter {
+  export import FirstMultiProvideClass = ಠ_ಠ.clutz.module$exports$goog$multi$provide$exporter$FirstMultiProvideClass;
```
evmar: This seems to be the same code repeated 3x

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: Rewritten - take another look?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: `displayNamePrefix` is a sublist of `displayNameParts`, so modifiying it modifies `displayNameParts`.  Added a comment.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Rewritten - take another look?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Whoops - done!

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
LucasSloan: Sounds good!

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
LucasSloan: Right, they need to be in their own directory to do the total compile as well as the syntax check.

src/test/java/com/google/javascript/clutz/crossModuleImports/goog_multi_provide_exporter.d.ts
LucasSloan: If you look very carefully, some dollar signs turned into dots.  That's the secret for allowing goog.provide() style imports, where importing `foo` gives you access to `foo.bar.baz`, even with the goog.module()/dollar sign style symbol names.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -24,11 +24,14 @@
...
     for (Node ast : parsedInputs) {
-      // Symbols can be imported into a variable in a goog.module() file, so look for imports in the
-      // body of the goog module.
       String moduleId = getGoogModuleId(ast);
       if (moduleId != null) {
+        // Symbols can be imported into a variable in a goog.module() file, so look for imports in the
+        // body of the goog module.
         importRenameMap.putAll(build(moduleId, ast.getFirstChild()));
+      } else {
+        // Also get renames goog.require() symbols in a goog.provide() file
```
evmar: typo: "also get renames" (?)

LucasSloan: Fixed.

LucasSloan:
`are you sure this does nothing in non-partial mode?`
Pretty sure, waiting on g3 run.
`is it worth having Rado review? I know little about clutz`
I don't think it's worth waiting a week, no.
`probably needs a run in g3 (for non-partial)`
Yep.

PR #655, opened 2018-01-09
Rename all references to symbols, not just top level ones.

LucasSloan:
To ensure compatibility between goog.module and goog.provide code, symbols have to be renamed.  Previously, only top level references to goog.provide symbols were renamed.  This PR renames other references in a given file (ie method params).

### DIFF
--- a/src/test/java/com/google/javascript/clutz/partial/static_method.d.ts
+++ b/src/test/java/com/google/javascript/clutz/partial/static_method.d.ts
@@ -1,0 +1,18 @@
+declare namespace ಠ_ಠ.clutz {
+  class module$exports$foo$bar$Class extends module$exports$foo$bar$Class_Instance {
+    static equals (c1 : module$exports$foo$bar$Class | null , c2 : module$exports$foo$bar$Class | null ) : boolean ;
+  }
+  class module$exports$foo$bar$Class_Instance {
+    private noStructuralTyping_: any;
+  }
+}
+declare namespace ಠ_ಠ.clutz.module$exports$foo.bar {
+  export import Class = ಠ_ಠ.clutz.module$exports$foo$bar$Class;
+}
+declare namespace ಠ_ಠ.clutz.module$exports$foo$bar {
+  export import Class = ಠ_ಠ.clutz.module$exports$foo$bar$Class;
+}
+declare module 'goog:foo.bar.Class' {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/partial/static_method.js
```diff
@@ -0,0 +1,16 @@
+goog.provide('foo.bar.Class');
+
```
evmar: Can you add docs to this about what it's testing, what the expected output should include?

LucasSloan: Added - do they answer your question adequately?

src/test/java/com/google/javascript/clutz/partial/static_method.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @constructor
+ */
+foo.bar.Class = function() {
+};
+
+//!! The method param references to foo.bar.Class should be renamed, along with
```
evmar: renamed to what?  (don't have to spell out exactly the target of it, but rather something "renamed to a dotted-module access" or whatever)

src/test/java/com/google/javascript/clutz/partial/static_method.d.ts
```diff
@@ -0,0 +1,19 @@
+declare namespace ಠ_ಠ.clutz {
+  class module$exports$foo$bar$Class extends module$exports$foo$bar$Class_Instance {
+    //!! The parameter type annotations match the (renamed) class name
```
evmar: Won't the test suite eat comments in here?

LucasSloan: Say more?

src/test/java/com/google/javascript/clutz/partial/static_method.js
LucasSloan: Done

src/test/java/com/google/javascript/clutz/partial/static_method.d.ts
evmar: Isn't this file effectively a golden generated by the test suite?  E.g. if you edit static_method.js, to rebaseline, don't you copy the output over this file?

LucasSloan: Good point - removed.

PR #656, opened 2018-01-12
use a sorted List for class template parameters

evmar:
We pass the template parameters of a class down through a whole
train of functions, ultimately to grab the "first" parameter for
special-casing Promises.  Unfortunately we were using a HashSet for
this, which meant that we could get the second parameter instead
of the first depending on how the HashSet worked out.

Rather than just making this a LinkedHashSet, I though I might as
well make it a List because that is a more appropriate type.  In
doing so I found one place where we were accidentally passing the
list of class template parameters in to a function that was expecting
something different.  I made my best guess at the behavior we actually
wanted there.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1012,7 +1012,7 @@ class DeclarationGenerator {
           } else {
             emit("function");
           }
-          treeWalker.visitProperty(propName, oType, false, false, Collections.<String>emptySet());
+          treeWalker.visitProperty(propName, oType, false, false, Collections.<String>emptyList());
         }
       }
     }
@@ -1407,7 +1407,7 @@ class DeclarationGenerator {
       visitTemplateTypes(ftype);
       emit("=");
       emit(typeName);
-      visitTemplateTypes(ftype, Collections.emptySet(), false);
+      visitTemplateTypes(ftype, Collections.emptyList(), false);
       emit(";");
       emitBreak();
       if (!ftype.isInterface()) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2184,7 +2184,7 @@ private void visitObjectType(
           false,
           Collections.<String>emptySet(),
           Collections.<String>emptySet(),
-          superClassFields);
+          Collections.<String>emptyList());
```
evmar: This is the one place I'm not sure about.  The param here is "classTemplateTypeNames", so I don't know why we were passing in the superClassFields before.

evmar:
This fixed a bug in how we generate .then() for goog.Promise but I'm not sure I can repro it in a test, because the bug depends on the ordering of HashSet.

LucasSloan:
I'm concerned that there was a reason we were using Set in the first place - did your archeology figure that out?  It might be safer to use a [LinkedHashSet](https://docs.oracle.com/javase/7/docs/api/java/util/LinkedHashSet.html).

mprobst:
I think this should be fine not to use a set. Template parameters defined on one class should have non-clashing names.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: That's a bug, but likely we wanted `superClassFields` as 'skipNames', so we don't double emit inherited fields. That said I think we might have had this bug for long time, so at least add a comment.

evmar:
Manually merged.

PR #658, opened 2018-01-16
Skip emitting symbols from tsickle's extern file.

rkirov:
The file named 'closure_externs.js' contains types that are in
lib.d.ts, but not in closure's externs. Therefore there is no need for
clutz to emit them.


evmar:
Manually merged.

PR #659, opened 2018-01-16
Fix for importing symbol S from a module named something.S.

rkirov:
Ideally, we shouldn't be generating null nodes, but I could not
reproduce that case from the user reports.

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
@@ -238,6 +238,9 @@ public final class ModuleConversionPass implements CompilerPass {
             && node.getNext().getFirstChild().matchesQualifiedName("goog.require")) {
           Node importedNode = node.getFirstChild();
           // For multiple destructuring imports, there are multiple full local names.
+          // This does not support renaming imports of the sort
+          // const {a: b} = goog.require
+          // TODO(rado): add support for that pattern.
           ArrayList<String> namedExports = new ArrayList<String>();
           while (importedNode != null) {
             namedExports.add(importedNode.getString());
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -26,6 +26,11 @@ protected GentsCodeGenerator(
 
   @Override
   protected void add(Node n, Context ctx) {
+    // Other CodeGenerators seem to safely assume n is non-null. So having a null n is
+    // odd. That said, I know this happens for some users and cannot reproduce it yet, so for now
+    // just continue emitting instead of crashing.
+    // TODO(rado): reproduce having a null node and fix at the generator source.
+    if (n == null) return;
```
evmar: Isn't there some standard way to dump a stack trace?

rkirov: I can get to the null node reproducibly in g3, but because this is during the emit phase it is hard to trace back to site of a actual AST mutation that introduced it.

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -515,13 +515,10 @@ void convertDestructuringRequireToImportStatements(Node n, ModuleImport moduleIm
...
 
-    // import {value as localName} from "./file"
+    // import {localName} from "./file"
     Node importSpec = new Node(Token.IMPORT_SPEC);
-    // import {a as b} only when a != b
     for (String localName : moduleImport.localNames) {
-      if (!moduleImport.moduleSuffix.equals(localName)) {
-        importSpec.addChildToBack(IR.name(localName));
-      }
+      importSpec.addChildToBack(IR.name(localName));
```
rkirov: This still looks very odd to me, because we don't do anything for renaming imports.

rkirov:
I think I got a much better fix, adding bowen for review.

PR #664, opened 2018-01-24
Use fake incremental numbers for numeric enums.

### LINKED ISSUE #663: Clutz should use individual enum values to create *.d.ts files
- What are you trying to do?
I have a proto enum that looks like:
```proto
enum Type {
  A = 1;
  B = 5;
}
```

Then on Typescript I want to be able to use `checkExhaustive` on that proto enum like this:
```ts
function checkExhaustive(a: never): never{ throw new Error() };

printMessage(type: Type):string {
  switch(type) {
    case Type.A: return 'Letter A';
    case Type.B: return 'Letter B'; 
    default: throw checkExhaustive(type);
  }
}
```

Example code that shows issue:
https://www.typescriptlang.org/play/#src=function%20x(a%3A%20never)%3A%20never%7Bthrow%20new%20Error()%7D%0D%0A%0D%0Adeclare%20namespace%20Cart%20%7B%0D%0A%20%20enum%20Type%20%7B%0D%0A%20%20%20%20A%2C%0D%0A%20%20%20%20B%2C%0D%0A%20%20%7D%0D%0A%0D%0A%20%20enum%20NumberedType%20%7B%0D%0A%20%20%20%20A%20%3D%201%2C%0D%0A%20%20%20%20B%20%3D%205%2C%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0Atype%20X%20%3D%20Cart.Type%0D%0A%0D%0Afunction%20y(p%3A%20Cart.Type)%20%7B%0D%0A%20%20switch%20(p)%20%7B%0D%0A%20%20%20%20case%20Cart.Type.A%3A%0D%0A%20%20%20%20case%20Cart.Type.B%3A%0D%0A%20%20%20%20%20%20break%3B%0D%0A%20%20%20%20default%3A%0D%0A%20%20%20%20%20%20x(p)%3B%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0Afunction%20z(p%3A%20Cart.NumberedType)%20%7B%0D%0A%20%20switch%20(p)%20%7B%0D%0A%20%20%20%20case%20Cart.NumberedType.A%3A%0D%0A%20%20%20%20case%20Cart.NumberedType.B%3A%0D%0A%20%20%20%20%20%20break%3B%0D%0A%20%20%20%20default%3A%0D%0A%20%20%20%20%20%20x(p)%3B%0D%0A%20%20%7D%0D%0A%7D

- What was the result?
I currently get "Argument of type 'Type' is not assignable to parameter of type 'never'."

- What did you want/expect?
I should be able to use `checkExhaustive`!

- More information
So I was playing with the code in the playground link shared above and I realized that the reason it doesn't work is because clutz doesn't specify the enum value for each individual item.
My guess is that if I don't set those numbers in the compiler, it knows the values are 0,1,2,3... and so on, but when the type comes from a declared namespace, then the compiler can't tell for sure because it might be 0,2,4,7... or whatever.

Is there a reason that clutz doesn't include the enum number in the generated declarations files?

Goodwine:
The value shouldn't really matter, but one should note that these values are not the real ones.
It would be possible to do myEnum=3 even if 3 is not in the enum when doing this is how real protos behave anyways.

Fixes #663

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1665,8 +1665,11 @@ class DeclarationGenerator {
         emit("{");
         emitBreak();
         indent();
+        int i = 0;
         for (String elem : sorted(type.getElements())) {
           emit(elem);
+          emit("=");
+          emit(String.valueOf(++i));
           emit(",");
           emitBreak();
         }
--- a/src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
+++ b/src/test/java/com/google/javascript/clutz/aliased_enums.d.ts
@@ -8,7 +8,7 @@ declare module 'goog:nested.bar.HahaEnum' {
 }
 declare namespace ಠ_ಠ.clutz.nested.baz {
   enum Enum {
-    A ,
+    A = 1 ,
   }
 }
 declare module 'goog:nested.baz.Enum' {
--- a/src/test/java/com/google/javascript/clutz/enum.d.ts
+++ b/src/test/java/com/google/javascript/clutz/enum.d.ts
@@ -11,8 +11,8 @@ declare module 'goog:some.ObjectValuedEnum' {
 }
 declare namespace ಠ_ಠ.clutz.some {
   enum SomeEnum {
-    A ,
-    B ,
+    A = 1 ,
+    B = 2 ,
   }
 }
 declare module 'goog:some.SomeEnum' {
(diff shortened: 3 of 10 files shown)


rkirov:
I am not sure how I feel about the fake numbers. It can throw off people that are looking into the generated .d.ts files. Are you sure this is not a TypeScript bug? Have you looked into their bug tracker about it?

Goodwine:
I just found it, indeed it is considered a typescript compiler bug: Microsoft/TypeScript/issues/16977

Goodwine:
I'll remove my PR for now, I'll see later if this can be done in a smarter way because it doesn't look like that TS bug has had any change in a while.

PR #665, opened 2018-01-24
Fix the paths so the DeclarationGeneratorTests work on Windows.

LucasSloan:


src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -131,14 +131,14 @@ StringSubject diagnosticStream() {
     Set<String> nonroots = new LinkedHashSet<>();
     for (File nonroot : actual().nonroots) {
       sourceFiles.add(SourceFile.fromPath(nonroot.toPath(), UTF_8));
-      nonroots.add(nonroot.getPath());
+      nonroots.add(nonroot.getPath().replace(System.getProperty("file.separator"), "/"));
```
rkirov: It would be better to fix it when the paths are constructed. Not sure how one does that in java, but in python all path joins must go through a library function path.join.

That said we should also be running tests on windows if we are serious about windows support, and at this point I just don't think it is worth the effort.

Also, according to https://github.com/angular/clutz/pull/593 the inputs also have to use '/' even on windows for clutz to work property.

PR #667, opened 2018-01-29
Pass in a list of module ids to treat as goog.provide when generating the import rename map.

LucasSloan:


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -129,12 +135,19 @@ public int parseArguments(Parameters params) throws CmdLineException {
   )
   String skipEmitRegExp = null;
 
+  @Option(
+    name = "--googProvides",
+    usage = "file containing a list of module ids to treat as goog.provide style modules"
```
rkirov: 'module ids' can be confusing, as they don't come from modules. I would say 'namespaces names that we know come from goog.provides (not goog.modules)'.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -129,7 +131,8 @@ private static String getGoogModuleId(Node astRoot) {
    * goog.module.get() assignments or destructuring assignments, parsing them, and generating the
    * mappings from local symbol names to exported symbol names.
```
rkirov: extend the comment with method local description of the purpose of knownGoogProvides

PR #670, opened 2018-01-30
Add a test to confirm that the closure fix for unresolved type union works in clutz.

### LINKED ISSUE #625: Union of two unresolved types is optimized into a single type
```javascript
/**
 * @param {Foo|Bar} a
 */
```

is emitted as `a: Foo`

This is a closure bug, so it would be best to fix in closure.

LucasSloan:
Fixes #625

### DIFF
--- a/src/test/java/com/google/javascript/clutz/partial/type_union.js
+++ b/src/test/java/com/google/javascript/clutz/partial/type_union.js
@@ -1,0 +1,8 @@
+goog.provide("func");
+goog.require("Foo");
+goog.require("Bar");
+
+/**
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/javascript/clutz/partial/type_union.js
```diff
@@ -0,0 +1,8 @@
+goog.provide("func");
+goog.require("Foo");
+goog.require("Bar");
+
+/**
+ * @param {Foo|Bar} a
```
rkirov: !Foo | !Bar to make the test tighter

PR #671, opened 2018-01-30
Fix for enum aliasing unknown enums.

### LINKED ISSUE #669: Enum lost when aliasing
```js
goog.module('enum_alias');

const OtherEnum = goog.require('not_visible');

/** @enum {string} */
const Enum = OtherEnum;

exports = Enum;
```

Produces 
```js
  var module$exports$enum_alias : any ;
```
which loses that `Enum` is also usable as a type.

rkirov:
Fixes #669


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1382,6 +1382,13 @@ private void walk(TypedVar symbol, String emitName) {
                     + " but type not found in Closure type registry.");
           }
         }
+        // The enum alias of an unknown type is present in the type registry, but it's type is still
```
evmar: its

PR #672, opened 2018-01-31
always qualify clutz names, add Node to global aliased symbols

evmar:
This reverts commit 9303d0c468d3fbcad557259e6ccab3ec694dc6d9,
"Don't emit ಠ_ಠ.clutz for missing symbols. (#609)"

We cannot rely on emitting unqualified symbols and having TypeScript
resolve them because they can be ambiguous.  Instead, just double down
on our aliasing workaround for global symbols.


src/test/java/com/google/javascript/clutz/partial/type_union.d.ts
```diff
@@ -1,5 +1,5 @@
 declare namespace ಠ_ಠ.clutz {
-  function func (a : Foo | Bar ) : void ;
+  function func (a : ಠ_ಠ.clutz.Foo ) : void ;
```
rkirov: ohoh, Bar disappeared?

PR #673, opened 2018-02-01
Support destructuring goog.require() variables on a subsequent line.

### LINKED ISSUE #668: Some `module$contents` are not rewritten when the come from destructuring
The following snippet:

```javascript
goog.module('my.module');

const otherModule = goog.require('some.other.module');

const {C} = otherModule;

/** @type {C} */
let x;

exports.x = x;
```

Emits `  var x : module$contents$my$module_C | null ;`, but we really want `var x : module$exports$some$other$module.C | null ;`. Note that our rewritter gets confused by the destructuring that happens on a separate line. However, for completely different reasons we need to keep that two line approach.

LucasSloan:
fixes #668

### DIFF
--- a/src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
+++ b/src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
@@ -90,6 +90,23 @@ public class ImportRenameMapBuilder {
             || rightHandSide.getFirstChild().matchesQualifiedName("goog.module.get"));
   }
 
+  /** Matches destructing from a variable ie `const {foo, bar: baz} = quux;` */
+  private static boolean isVariableDestructuringAssignment(Node statement) {
+    if (!(statement.isConst() || statement.isVar() || statement.isLet())) {
+      return false;
+    }
+
+    if (!statement.getFirstChild().isDestructuringLhs()) {
+      return false;
+    }
+
+    Node destructuringAssignment = statement.getFirstChild();
+
+    Node rightHandSide = destructuringAssignment.getChildAtIndex(1);
+
+    return rightHandSide.isName();
+  }
+
   /**
    * Matches either `const {foo} = goog.require()` or `const {foo} = goog.module.get()` depending on
    * if statement is in a goog.module or a goog.scope.
@@ -167,8 +184,10 @@ public class ImportRenameMapBuilder {
           // generates, so we have to extract it.
           String variableName;
           if (destructured.getFirstChild() != null) {
+            // Renaming
             variableName = destructured.getFirstChild().getString();
           } else {
+            // No rename
             variableName = originalName;
           }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -90,6 +90,25 @@ private static boolean isImportAssignment(Node statement) {
             || rightHandSide.getFirstChild().matchesQualifiedName("goog.module.get"));
   }
 
+  /** Matches `const {foo} = bar` */
+  private static boolean isVariableDestructuringAssignment(Node statement) {
+    String arewuipaeug = statement.toStringTree();
```
rkirov: ?

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -90,6 +90,25 @@ private static boolean isImportAssignment(Node statement) {
             || rightHandSide.getFirstChild().matchesQualifiedName("goog.module.get"));
   }
 
+  /** Matches `const {foo} = bar` */
```
rkirov: Does it also match `const {foo, bar} = buz`? Expand comment.

rkirov: Further comments seem to say that you also support `const {foo: bar} = buz`.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -185,6 +204,36 @@ private static String getGoogModuleId(Node astRoot) {
...
+            variableName = originalName;
+          }
+
+          // The statements are iterated in order, so the statement with the goog.require() is already
+          // processed, so the value is already in the import rename map.
+          String exportedSymbolName =
+              importRenameMap.get(destructuredVariable) + "." + originalName;
+
+          importRenameMap.put(variableName, exportedSymbolName);
+          // If we're in a goog scope, there isn't a module id
```
rkirov: Feels like this check has to be moved higher up in the processing logic. I think (could be wrong) that one cannot have goog.require within a goog.scope.

The only two patterns I am familiar with is:
old
```
goog.require('a.b.c');
goog.scope(function() {
var c = a.b.c;
});
```
or new

```
goog.module(a.b.c);
const c = goog.require('a.b.c');
```

we are only trying to fix the new one, unless I forgot something.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
```diff
@@ -185,6 +204,36 @@ private static String getGoogModuleId(Node astRoot) {
...
+        String destructuredVariable = statement.getFirstChild().getChildAtIndex(1).getString();
+
+        for (Node destructured : statement.getFirstFirstChild().children()) {
+          String originalName = destructured.getString();
+          // Destructuring can use the original name `const {A} = goog.require("foo.a")` or rename
+          // it `const {A: RenamedA} = ...`, and closure uses whichever in the symbol name it
+          // generates, so we have to extract it.
+          String variableName;
+          if (destructured.getFirstChild() != null) {
+            variableName = destructured.getFirstChild().getString();
```
rkirov: because of closures' messy AST, it is not clear which branch is the renaming destructuring. Add a one line comment.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Done.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Whoops!

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Done.

src/main/java/com/google/javascript/clutz/ImportRenameMapBuilder.java
LucasSloan: Goog.scope lets you write goog.module code inside a goog.provide file.  You can write:

```
goog.scope(function() {
var c = goog.require('a.b.c');
});
````

Which is what this is covering.

PR #674, opened 2018-02-01
Add goog.provide to the Logger symbols in partial_base.

rkirov:
This fixes the issue that Logger starts to appear in goog.debug
namespace. However, now we have to exclude partial_base from
debug/logger compilations, which also sucks :( WIP

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -17,6 +17,7 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
@@ -221,6 +222,25 @@ class DeclarationGenerator {
           "while",
           "with");
 
+  /**
+   * Incremental clutz cannot infer class aliases like:
+   *
+   * <pre>
+   *   /** @constructor *
+   *   an.KlassAlias = some.Missing;
+   * </pre>
+   *
+   * Closure reports an.KlassAlias as regular empty class. In order to emit the correct clutz
+   * description for now we work-around this issue by hardcoding some known aliases.
+   *
+   * <p>So far this works only for classes with no generic type parameters.
+   */
+  private static final ImmutableMap<String, String> KNOWN_CLASS_ALIASES =
+      ImmutableMap.of(
+          "goog.log.Logger", "goog.debug.Logger",
+          "goog.log.Level", "goog.debug.Logger.Level",
+          "goog.log.LogRecord", "goog.debug.LogRecord");
+
   private static final String GOOG_BASE_NAMESPACE = "goog";
 
   /**
@@ -408,6 +428,7 @@ class DeclarationGenerator {
 
   void generateDeclarations() {
     List<SourceFile> sourceFiles = new ArrayList<>();
+
     for (String source : opts.arguments) {
       sourceFiles.add(SourceFile.fromFile(source, UTF_8));
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -408,7 +408,13 @@ private boolean shouldEmitTypedefByName(JSType realType) {
 
   void generateDeclarations() {
     List<SourceFile> sourceFiles = new ArrayList<>();
+    Boolean isVisitingClosureDebug =
+        opts.partialInput
+            && (opts.arguments.contains("javascript/closure/debug/logrecord.js")
+                || opts.arguments.contains("javascript/closure/debug/logger.js")
```
rkirov: This blows. Any better ideas?

src/resources/partial_goog_base.js
```diff
@@ -14,6 +14,11 @@ var goog = goog || {};
 goog.abstractMethod = function() {
 };
 
+// Without these lines the symbols will appear on goog.debug.
```
evmar: What symbols?

src/test/java/com/google/javascript/clutz/partial/goog_debug.js
```diff
@@ -0,0 +1,8 @@
+goog.provide('goog.debug');
+
+//!! This test is testing that the workaround for goog.debug.Logger
+//!! tested by missing_logger_dot_level, does not affect other files
+//!! like debug.js
```
evmar: I don't understand this comment, can you rethink it?

rkirov:
Sorry, I wrote this in a rush to catch a bus. I expanded the comments, start with this comment - https://github.com/angular/clutz/pull/674/files#diff-0e5e178a3d202663d2681b1be0c5ed0b

Also it is a WIP because I hate the workaround - can you come up with something better?

LucasSloan:
It might be better to try and tackle the issue that caused those symbols to be added to partial_goog_base, rather than add another hack on top.

rkirov:
Attempted a different solution. Has some good features - not touching partial_base, no hardcoding of filenames, but also some downsides - cannot handle templated types or interfaces.

Going to sleep on it and decide one of the two tomorrow :/

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1418,28 +1435,40 @@ private boolean isNewableFunctionType(FunctionType type) {
...
      */
-    private void visitTypeValueAlias(String unqualifiedName, ObjectType otype) {
-      String typeName = Constants.INTERNAL_NAMESPACE + "." + otype.getDisplayName();
+    private void visitTypeValueAlias(
+        String unqualifiedName, @Nullable ObjectType otype, @Nullable String alternativeAliasName) {
+      if ((otype == null) == (alternativeAliasName == null)) {
+        throw new IllegalArgumentException(
+            "Visiting symbol "
+                + unqualifiedName
+                + " with aliased type and alternative name or neither.");
```
rkirov: also known as I wish java had union types.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -221,6 +222,21 @@
           "while",
           "with");
 
+  /**
+   * Incremental clutz cannot infer class aliases like:
+   * <pre>
+   *   @constructor
```
evmar: This really should be /** @constructor */
which can't work in this comment but maybe you can avoid the "*/" part of it

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
evmar: I think I'd overload the function maybe.

rkirov:
Updated PTAL

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -221,6 +222,23 @@
           "while",
           "with");
 
+  /**
+   * Incremental clutz cannot infer class aliases like:
+   * <pre>
+   *   ** @constructor *
```
evmar: You can even write
"/**"
just not the closing one

PR #675, opened 2018-02-03
rationalize handling of platform types

evmar:
See individual commits.

This partially reverts commit 9303d0c, then fixes it a different way.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -897,19 +766,32 @@ private boolean isDefaultExport(TypedVar symbol) {
...
+
+    // We're asked about e.g. "Array.from", which we should treat as a platform extern.
+    // The PlatformSymbols lists just contain the toplevel name like "Array", so strip after
+    // the first dot.
+    int dotPos = symbolName.indexOf('.');
+    if (dotPos > 0) {
+      symbolName = symbolName.substring(0, dotPos);
+    }
+
+    // Don't emit externs for Closure types that nobody uses.
```
rkirov: This line goes against the method name 'isPlatformExtern', can we change the method name, or at least put a big warning there, that for the purposes of optimization we kinda lie here.

If we make externs_list incremental too one day, we should revisit this.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2847,13 +2717,13 @@ public void emitPrivateValue(String emitName) {
...
     public Void emitObjectType(
         ObjectType type, boolean extendingInstanceClass, boolean inExtendsImplementsPosition) {
-      if (maybeEmitGlobalAlias(type)) return null;
       // Closure doesn't require that all the type params be declared, but TS does
       if (!type.getTemplateTypeMap().isEmpty()
           && !typeRegistry.getNativeType(OBJECT_TYPE).equals(type)) {
         return emitTemplatizedType(
             typeRegistry.createTemplatizedType(type), false, inExtendsImplementsPosition);
       }
+      if (maybeEmitGlobalType(type)) return null;
```
rkirov: odd that this had to move

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -0,0 +1,1716 @@
...
+   * }
+   * </pre>
+   *
+   * Adding a symbol to this list also suppresses the duplicate externs.
+   */
+  public static final ImmutableSet<String> GLOBAL_SYMBOL_ALIASES =
+      ImmutableSet.of("Error", "Event", "EventTarget", "Object", "Date");
+
+  /**
+   * Set of all symbols that are defined in a platform externs.js that we don't have matching
```
rkirov: Technically not "all symbols".

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -0,0 +1,1716 @@
...
+   * TypeScript type for. We could in theory just let Clutz generate types for these, but in
+   * practice no users have needed them yet and it saves file space. It should be safe to remove
```
rkirov: Sentence repeated in the top of the class. I think it is not needed here.

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -0,0 +1,1716 @@
...
+          "webkitIDBTransaction",
+          "webkitIDBVersionChangeEvent",
+          "webkitMediaStream",
+          "webkitOfflineAudioContext",
+          "webkitPannerNode",
+          "webkitRequestFileSystem",
+          "webkitResolveLocalFileSystemURI",
+          "webkitURL");
+
+  /** Set of all symbols exposed by TypeScript's lib.*.d.ts. */
```
rkirov: can you at least throw the little ts extractor in a gist here

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -0,0 +1,1716 @@
...
+ * defines them in a lib.d.ts file. When a Closure file references a platform type like HTMLElement
+ * or platform global like encodeURIComponent, rather than referring to a type that Clutz derives
+ * from the externs we'd rather it refer directly to the TypeScript type.
+ *
+ * <p>Note that doing this is not strictly correct (the two may have slightly different definitions)
+ * but this is likely what users expect and it reduces the output size.
+ *
+ * <p>This class maintains all the metadata about these types; for example, Closure and TypeScript
+ * have different names for the interface that is the 'style' property on an element, so this class
+ * has a map from Closure name to TypeScript name.
```
rkirov: add general guidance on adding and removing symbols in this class

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
evmar: Without it some tests fail because we write Array in place of any[].

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
evmar: Renamed, not super happy with the name.

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
evmar: Can you elaborate?  I think the local comments are sufficient maybe?

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
evmar: Done.

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
evmar: Whcih sentence?

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
evmar: done

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
rkirov: Imagine x months in the future: lib.d.ts and platform externs have changes. How does one go about updating this lists.

Can we write a simple flow script

- if Symbol X appears in lib.d.ts and
  - it is not in closure and someone needs to use it google add it to ...
  - it is in closure, add it to ...
  - it is not is closure and noone needs to use it do nothing.

- if symbol x appears in closure externs
 - ...

Maybe I am overthinking it ...

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
rkirov: nevermind, I thought that 'not strictly correct' and 'in theory' referred to the same exception. Now I see I was wrong.

PR #676, opened 2018-02-03
Goog.require ambiguity rewriting now works for generic type params.

rkirov:
The rewriter logic is now hooked in one more symbol emit code path.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1746,15 +1746,8 @@ class DeclarationGenerator {
         emit(defaultEmit);
         return;
       }
-      String displayName = type.getDisplayName();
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1774,6 +1767,18 @@ private void emitNoResolvedType(NoType type, String defaultEmit) {
       }
     }
 
+    private String maybeRewriteImportedName(String displayName) {
+      // In partial mode, closure doesn't know the correct name of imported symbols, if the name
+      // matches one in the precomputed map, replace it with the original declared name
+      // The displayName can be of the form foo.bar, but the symbol that was goog required was just
+      // foo, so just replace the part of the display name before the first period
```
evmar: Should this be jsdoc on the function itself?

PR #677, opened 2018-02-06
Make the knownGoogProvides whitelist work with destructuring imports.

LucasSloan:
Previously all symbols in the knownGoogProvides list were aliased to the bare namespace, but if the import is destructured, need to alias to namespace + symbol name.


src/test/java/com/google/javascript/clutz/partialCrossModuleTypeImports/goog_legacy_namespace_exporter.js
```diff
@@ -0,0 +1,11 @@
+goog.module("goog.legacy.namespace.exporter");
+goog.module.declareLegacyNamespace();
```
rkirov: add a comment about the role of declareLegacyNamespace in the test. Would this test work if this was just a goog.provide('goog.legacy.namespace.exporter')?

LucasSloan: Added a comment explaining that the declare legacy namespace isn't load bearing, it's just the example I came across.

PR #678, opened 2018-02-06
Add a boolean that can be modified to add the contents of importRenameMap in the dts output as comments.

LucasSloan:


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -343,6 +344,7 @@ public static void main(String[] args) {
    * is used to store the mappings from the closure supplied names to the correct names.
    */
   private Map<String, String> importRenameMap = Collections.emptyMap();
+  private final boolean PRINT_IMPORT_RENAME_MAP = false;
```
rkirov: Document what is this for.

PR #679, opened 2018-02-07
Refactor the known class aliases code to handle goog modules with class aliases.

LucasSloan:
The goog.module requires a type annotation to work, and the known class aliases map is read in from a file instead of being hard coded.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -13,9 +13,11 @@ import com.google.javascript.jscomp.parsing.Config;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
 import org.kohsuke.args4j.Argument;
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -142,6 +144,12 @@ public int parseArguments(Parameters params) throws CmdLineException {
   )
   String googProvidesFile = null;
 
+  @Option(
+    name = "--knownClassAliases",
+    usage = "file containing a map of names that we know are aliases"
```
rkirov: It would nice to describe the expected format for this file. Probably the usage string is too short for this, but maybe in a regular comment

src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -227,6 +250,18 @@ public CompilerOptions getCompilerOptions() {
         throw new RuntimeException("Error reading goog provides file " + googProvidesFile, e);
       }
     }
+
+    if (knownClassAliasesFile != null) {
+      try {
+        for (String line : Files.readLines(new File(knownClassAliasesFile), UTF_8)) {
+          String[] alias = line.split(",");
```
rkirov: add some basic checking here. If alias.length is != 2, should throw some meaningful error.

PR #682, opened 2018-02-09
add a whitelist for aliasing goog.provides

evmar:
If a file A has
  goog.provide('mything');
  mything = 3;
and a file B has
  goog.provide('mything.subnamespace');
  mything.subnamespace = 4;

Then the code is nonsense but clutz has a workaround anyway.

In partial mode we look at A without knowledge of B, so we cannot apply
the workaround without providing a global set of broken inputs like
these.  This change adds support for the global set and hooks it up to
a command-line flag.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -151,6 +151,12 @@ public class Options {
   /** Mappings should be comma separated, with no spaces, one mapping per line */
   String knownClassAliasesFile = null;
 
+  @Option(
+    name = "--collidingProvides",
+    usage = "file containing a list of names that we know conflict with namespaces"
+  )
+  String collidingProvidesFile = null;
+
   @Argument List<String> arguments = new ArrayList<>();
 
   Depgraph depgraph;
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -151,13 +151,20 @@ public int parseArguments(Parameters params) throws CmdLineException {
   /** Mappings should be comma separated, with no spaces, one mapping per line */
   String knownClassAliasesFile = null;
 
+  @Option(
+    name = "--aliasedNames",
```
rkirov: aliased referes more the the workaround than the original problem. I would call it "assignedNamespaces" (because foo.bar was supposed to be a namespace, but got assigned a "non-namespace" value. 
Or at least "needAliasedNames".

src/test/java/com/google/javascript/clutz/partial/symbol_alias.d.ts
```diff
@@ -0,0 +1,14 @@
+declare namespace ಠ_ಠ.clutz.symbol_alias {
+  var aliased : any;
+}
+declare module 'goog:symbol_alias.aliased' {
+  import alias = ಠ_ಠ.clutz.symbol_alias.aliased;
```
rkirov: needs the actual alias

evmar:
PTAL, renamed everything

PR #683, opened 2018-02-09
Better bare reexport

LucasSloan:
Read goog.require statements to generate possible export aliases, instead of passing in a map as an option.

Refactor ImportRenameMapBuilder to extend an abstract class, so AliasMapBuilder can rely on the same infrastructure.  Only supports reexports in goog.module() files, continues to rely on KNOWN_CLASS_ALIASES for goog.provide() class aliases.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
+++ b/src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
@@ -1,0 +1,89 @@
+package com.google.javascript.clutz;
+
+import com.google.javascript.rhino.Node;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Closure doesn't handle alias reexports like the following well:
+ *
+ * <p>`goog.module('bare.reexport'); const {Class} = goog.require('original.module'); exports =
+ * Class;`
+ *
+ * <p>It completely ignores the goog.require() and gives Class the unknown {?} type.
+ *
+ * <p>AliasMapBuilder looks for goog.require statements and generates mappings from the exported
+ * name of Class (here module$exports$bare$reexport) to the name of the imported symbol (here
+ * module$exports$original$module.Class), so if clutz finds a symbol that has no type, it can emit
+ * it as an alias to the correct symbol.
+ */
+public class AliasMapBuilder extends ImportBasedMapBuilder {
+  @Override
+  protected Map<String, String> build(
+      String localModuleId, Node moduleBody, Set<String> knownGoogProvides) {
+    Map<String, String> aliasMap = new HashMap<>();
+    if (localModuleId == null) {
+      //TODO(lucassloan): handle goog.module.get()
+      return aliasMap;
+    }
+
+    for (Node statement : moduleBody.children()) {
+      if (isImportAssignment(statement)) {
+        // `const C = goog.require()` or
+        // `const C = goog.module.get()`
+        String importedModuleId = statement.getFirstFirstChild().getChildAtIndex(1).getString();
+        String variableName = statement.getFirstChild().getString();
+
+        String exportedSymbolName = buildWholeModuleExportSymbolName(importedModuleId);
+        aliasMap.putAll(
+            buildNamedAndWholeModuleMappings(localModuleId, variableName, exportedSymbolName));
+
+      } else if (isImportDestructuringAssignment(statement)) {
+        // `const {C, Clazz: RenamedClazz} = goog.require()` or
+        // `const {C, Clazz: RenamedClazz} = goog.module.get()`
+        String importedModuleId =
+            statement.getFirstChild().getChildAtIndex(1).getChildAtIndex(1).getString();
+        for (Node destructured : statement.getFirstFirstChild().children()) {
+          String originalName = destructured.getString();
+          // Destructuring can use the original name `const {A} = goog.require("foo.a")` or rename
+          // it `const {A: RenamedA} = ...`, and closure uses whichever in the symbol name it
+          // generates, so we have to extract it.
+          String variableName;
+          if (destructured.getFirstChild() != null) {
+            // Renaming
+            variableName = destructured.getFirstChild().getString();
+          } else {
+            // No rename
+            variableName = originalName;
+          }
+
+          String exportedSymbolName = buildNamedExportSymbolName(importedModuleId, originalName);
+          aliasMap.putAll(
+              buildNamedAndWholeModuleMappings(localModuleId, variableName, exportedSymbolName));
+        }
+      }
+    }
+
+    return aliasMap;
+  }
+
+  /**
+   * To avoid parsing for the export statements, generate mappings for both styles of exports from
+   * goog.modules.
+   *
+   * <p>`exports = Foo` produces the name `module$exports$module$Name` `exports.Foo = Foo` produces
+   * the name `module$exports$module$Name.Foo`
+   */
+  private static Map<String, String> buildNamedAndWholeModuleMappings(
+      String localModuleId, String variableName, String exportedSymbolName) {
+    //TODO(lucassloan): handle knownGoogProvides
+    Map<String, String> mappings = new HashMap<>();
+    // exports = Foo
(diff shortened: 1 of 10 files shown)


src/test/java/com/google/javascript/clutz/partial/bare_reexport.d.ts
```diff
@@ -1,6 +1,6 @@
 declare namespace ಠ_ಠ.clutz {
-  type module$exports$bare$reexport = ಠ_ಠ.clutz.module$exports$original$module.Class ;
-  var module$exports$bare$reexport : typeof ಠ_ಠ.clutz.module$exports$original$module.Class ;
+  type ಠ_ಠ.clutz.module$exports$bare$reexport = module$exports$original$module.Class ;
```
evmar: This looks wrong, it has the eyes twice

LucasSloan: Good catch - fixed!

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1363,6 +1389,21 @@ private void visitTypeValueAlias(String unqualifiedName, ObjectType otype) {
       typesUsed.add(otype.getDisplayName());
     }
 
+    private void visitTypeValueAlias(String unqualifiedName, String emitName) {
```
rkirov: try reusing visitKnownTypeValueAlias

LucasSloan: Yep, that works - thanks for pointing it out!

LucasSloan:
PTAL - fixed your comment.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.google.javascript.rhino.Node;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * AliasMapBuilder builds a map from the imported name of a symbol (based on a goog.require) to the
+ * exported name of a symbol (based on the variable name and goog.module id), so symbols that are
+ * imported and reexported can be properly aliased.
```
rkirov: needs an example, I can't fully follow the text without an example here.

Word it like, given this code:
const A = goog.require('x.y.z') the map will contains A -> x.y.z,

doesnt' need to cover all cases.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+      if (isImportAssignment(statement)) {
+        // `const C = goog.require()` or
+        // `const C = goog.module.get()`
+        String importedModuleId = statement.getFirstFirstChild().getChildAtIndex(1).getString();
+        String variableName = statement.getFirstChild().getString();
+
+        String exportedSymbolName = buildWholeModuleExportSymbolName(importedModuleId);
+        //TODO(lucassloan): handle knownGoogProvides
+        // exports = Foo
+        aliasMap.put(buildWholeModuleExportSymbolName(localModuleId), exportedSymbolName);
```
rkirov: extract the putting of the two entries into aliasMap into a helper function, because it is duplicated.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+    for (Node statement : moduleBody.children()) {
+      if (isImportAssignment(statement)) {
+        // `const C = goog.require()` or
+        // `const C = goog.module.get()`
+        String importedModuleId = statement.getFirstFirstChild().getChildAtIndex(1).getString();
+        String variableName = statement.getFirstChild().getString();
+
+        String exportedSymbolName = buildWholeModuleExportSymbolName(importedModuleId);
+        //TODO(lucassloan): handle knownGoogProvides
+        // exports = Foo
```
rkirov: I am confused here because you are not actually looking for 'exports = Foo'. I understand that's the ultimate intention of the code, but it needs some more explanation.

Currently the map just contains module$exports$C = module$exports$x.y.z. I guess closure will resolve from exports.C to module$exports$C , but that needs explanation.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1307,7 +1329,11 @@ private void walk(TypedVar symbol, String emitName) {
           visitTypeValueAlias(symbol.getName(), (EnumElementType) registryType);
           return;
         }
-        visitVarDeclaration(getUnqualifiedName(emitName), type);
+        if (aliasMap.containsKey(emitName)) {
```
rkirov: This is my overlooking, but it would be nice to have a comment describing when do we hit this branch - 'missing provide'?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -194,6 +214,8 @@ public static void main(String[] args) {
    */
   private Map<String, String> importRenameMap = Collections.emptyMap();
 
+  private Map<String, String> aliasMap = Collections.emptyMap();
```
rkirov: Describe the need for aliasMap here or point to the documentation in the other class.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
LucasSloan: Reworked the comment - thoughts?

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
LucasSloan: Put a comment on the extracted method - thoughts?

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
LucasSloan: Done.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: I don't quite follow, but added a comment.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Closure doesn't handle alias reexports like the following well:
+ *
+ * `goog.module('bare.reexport');
+ * const {Class} = goog.require('original.module');
+ * exports = Class;`
```
rkirov: What happens in the following scenario:

```
const {Class} = goog.require('original.module');
const {Class2} = goog.require('original.module2');

exports = Class;
```

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -0,0 +1,88 @@
...
+ *
+ * It completely ignores the goog.require() and gives Class the unknown {?} type.
+ *
+ * AliasMapBuilder looks for goog.require statements and generates mappings from the
+ * exported name of Class (here module$exports$bare$reexport) to the name of the imported
+ * symbol (here module$exports$original$module.Class), so if clutz finds a symbol that
+ * has no type, it can emit it as an alias to the correct symbol.
+ */
```
rkirov: Much better comment.

LucasSloan: Thanks!

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
LucasSloan: There are 4 mappings in the map, but only one will be used, since only Class is used as an export.

PR #684, opened 2018-02-10
Fix interaction between importRewriter and Global aliases.

rkirov:
The import rewriter has to be run before the code that looks for global
aliases. For example:

const EventTarget = goog.require('goog.events.EventTarget');

/** @type {EventTarget} */

The name has to be expanded to goog.events.EventTarget, otherwise we
would rewrite it as GlobalEventTarget.


src/test/java/com/google/javascript/clutz/partial/platform_goog_require.js
```diff
@@ -0,0 +1,14 @@
+/**
+ * This tests that EventTarget is expanded to goog.events.EventTarget, while processing
+ * Thus not confused with the platform global EventTarget, and not rewriten to GlobalEventTarget.
```
evmar: Missing a period on line 2 or something?  Confusing trail off here

PR #685, opened 2018-02-10
Make sure the symbol is unknown type before assuming its an alias.

LucasSloan:
Ran some tests in g3, and discovered I needed a better conditional.


LucasSloan:
Done!

src/test/java/com/google/javascript/clutz/partial/import_without_reexport.js
```diff
@@ -0,0 +1,8 @@
+goog.module('import.without.reexport');
+
```
rkirov: Add a comment what this is testing. It is had to see why the import without reexport matters, without understanding current working of aliasMap.

Long-term I think we should just parse the 'export = ', it is too much of jump to assume that the last import is reexported.

LucasSloan: Added a comment.

Agreed that's the right solution.

PR #686, opened 2018-02-10
Parse the depgraph to determine which imports are goog.provides vs. goog.modules

LucasSloan:


src/main/java/com/google/javascript/clutz/Depgraph.java
```diff
@@ -27,6 +30,7 @@
   private final Set<String> nonroots = new LinkedHashSet<>();
   private final Set<String> rootExterns = new LinkedHashSet<>();
   private final Set<String> nonrootExterns = new LinkedHashSet<>();
+  private final Set<String> knownGoogProvides = new HashSet<>();
```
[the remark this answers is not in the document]
rkirov: also need documentation why do we like to collect that data from the depgraph.

src/main/java/com/google/javascript/clutz/Depgraph.java
```diff
@@ -118,12 +126,26 @@ private void collectFiles(boolean isRoots, List<List<?>> fileList) {
...
       boolean isExterns = false;
+      boolean isGoogModule = false;
+      List<String> provides = new ArrayList<>();
       for (List<?> tuple : fileProperties) {
         String key = (String) tuple.get(0);
         if ("is_externs".equals(key) && Boolean.TRUE.equals(tuple.get(1))) {
           isExterns = true;
           break;
         }
+        if ("load_flags".equals(key)) {
```
rkirov: add some comments how does the depgraph looks likes at this point. It's an odd format will be useful to have more docs around it.

src/main/java/com/google/javascript/clutz/Depgraph.java
```diff
@@ -135,6 +157,9 @@ private void collectFiles(boolean isRoots, List<List<?>> fileList) {
       } else {
         nonroots.add(fileName);
       }
+      if (!isGoogModule) {
```
rkirov: Name flag and the Set<String> consistently. Either both module or provide.

src/test/java/com/google/javascript/clutz/partial/platform_goog_require.d.ts
```diff
@@ -1,7 +1,7 @@
 declare namespace ಠ_ಠ.clutz.module$exports$platform$goog$require {
   class C extends C_Instance {
   }
-  class C_Instance extends ಠ_ಠ.clutz.goog.events.EventTarget {
+  class C_Instance extends ಠ_ಠ.clutz.module$exports$goog$events$EventTarget {
```
rkirov: We have to fix the test by adding a simple depgraph for it, or just mocking it. Otherwise, this is a bad test with this change.

src/main/java/com/google/javascript/clutz/Depgraph.java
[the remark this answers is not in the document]
LucasSloan: Done, done.

src/main/java/com/google/javascript/clutz/Depgraph.java
LucasSloan: Done.

src/main/java/com/google/javascript/clutz/Depgraph.java
LucasSloan: Done.

src/test/java/com/google/javascript/clutz/partial/platform_goog_require.d.ts
LucasSloan: Added a comment that that's orthogonal to the issue of ambiguity with GlobalEventTarget.

PR #690, opened 2018-02-13
update closure version to v20180204

evmar:
Two changes required:

- The test output changed for a platform extern; this was an intentional
  change in the upstream extern (the function param should have been {?}).

- In gents, set the "shorthand property" flag on all imports/exports.
  This is related to the AST for statments like:
    export {x as x};
  This flag is saying whether the code is ok with that being emitted as
    export {x};
  which we want in all cases.  (If it's an {x as y}, where x!=y, then
  the flag has no effect.)


src/test/java/com/google/javascript/clutz/implements_iterable_with_platform.d.ts
```diff
@@ -28,7 +28,7 @@ declare namespace ಠ_ಠ.clutz.implements_iterable {
   class ImplIterableIterator_Instance implements IterableIterator < string > {
     private noStructuralTyping_: any;
     [Symbol.iterator]():  IterableIterator < string > ;
-    next (a ? : string ) : IteratorResult < string > ;
+    next (a ? : any ) : IteratorResult < string > ;
```
rkirov: weird

PR #691, opened 2018-02-13
Only add entries to aliasMap if the imported variable is reexported.

LucasSloan:
Previously, AliasMapBuilder optimistically produced the possible aliases for an import, without checking for a corresponding export.  This caused exports to be treated as aliases even if they weren't.  By only adding entries for variables that are known to be reexported, only actual aliases will be treated as such.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
+++ b/src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
@@ -28,17 +28,19 @@ public class AliasMapBuilder extends ImportBasedMapBuilder {
       return aliasMap;
     }
 
+    // Loop over the statements, looking for import statements, and build a map from the local variable
+    // names to the original symbol name eg `const C = goog.require('a.b.c');` will result in the map
+    // containing 'C' -> 'module$exports$a$b$c'
+    Map<String, String> localVariableToImportedSymbolNameMap = new HashMap<>();
     for (Node statement : moduleBody.children()) {
       if (isImportAssignment(statement)) {
         // `const C = goog.require()` or
         // `const C = goog.module.get()`
         String importedModuleId = statement.getFirstFirstChild().getChildAtIndex(1).getString();
-        String variableName = statement.getFirstChild().getString();
+        String localVariableName = statement.getFirstChild().getString();
 
-        String exportedSymbolName = buildWholeModuleExportSymbolName(importedModuleId);
-        aliasMap.putAll(
-            buildNamedAndWholeModuleMappings(localModuleId, variableName, exportedSymbolName));
-
+        String importedSymbolName = buildWholeModuleExportSymbolName(importedModuleId);
+        localVariableToImportedSymbolNameMap.put(localVariableName, importedSymbolName);
       } else if (isImportDestructuringAssignment(statement)) {
         // `const {C, Clazz: RenamedClazz} = goog.require()` or
         // `const {C, Clazz: RenamedClazz} = goog.module.get()`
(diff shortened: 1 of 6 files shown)


src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -28,17 +28,18 @@
       return aliasMap;
     }
 
+    // Loop over the statements, looking for import statements, and build a map from the local variable
+    // names to the original symbol name
```
rkirov: add example:
' const C = goog.require('a.b.c');`
the map will contain: 'C' -> 'a.b.c'

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
```diff
@@ -49,43 +50,93 @@
...
-    mappings.put(buildNamedExportSymbolName(localModuleId, variableName), exportedSymbolName);
-
-    return mappings;
+  /** Matches `exports = foo;` */
+  protected boolean isWholeModuleExportAssignment(Node statement) {
+    if (!statement.isExprResult()) {
+      return false;
+    }
+
+    if (!statement.getFirstChild().isAssign()) {
```
rkirov: Previous such code in gents, ended up with NPEs, because some closure apis can return null. Can you check for null, or convince me that it can't happen for things like getFirstChild() and getChildAtIndex(1).

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -976,6 +980,29 @@ private void declareNamespace(
     }
   }
 
+  /**
+   * Special emit for known aliases See KNOWN_ALIASES comment why this workaround is need.
+   *
+   * <p>This shares some similarity with visitTypeValueAlias method above, but it has to hardcode
+   * one assumptions - the alias is for a class with no generics.
+   */
+  private void visitKnownTypeValueAlias(String unqualifiedName, String alternativeAliasName) {
```
rkirov: Any reason to move this?

src/test/java/com/google/javascript/clutz/partial/import_without_reexport.js
```diff
@@ -1,12 +1,9 @@
...
 //!! actually aliases
-//!! TODO(lucassloan): actually parse the `exports = foo` or `exports.foo = foo`
-//!! statements, to actually solve this problem
 goog.module('import.without.reexport');
 
 const {Class} = goog.require('original.module');
 
-/** @const {type.not.present} */
-const x = null;
+const x = unknown.function(Class);
```
rkirov: OOC, why change this test? Maybe it wasn't a good enough case for the real internal google error, but then keep the old assertion and add a new one

`exports.x = x` and `exports.y = y`.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
LucasSloan: Done.

src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
LucasSloan: I don't think we're vulnerable to NPEs here, since we're checking for the types of the parent node before we access the child nodes.  ImportBasedMapBuilder has a bunch of similar checks which work just fine without explicit null checks.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: Alas, yes.  I had to call it from declareNamespace(), which isn't part of TreeWalker.

src/test/java/com/google/javascript/clutz/partial/import_without_reexport.js
LucasSloan: Sure.

PR #693, opened 2018-02-14
Emit generic type params for global types.

LucasSloan:


src/test/java/com/google/javascript/clutz/partial/global_generic_type.js
```diff
@@ -0,0 +1,4 @@
+goog.provide('global.generic.type');
+
+/** @type {Map<string, string>} */
```
rkirov: Just !Map to make the test a little simpler

src/test/java/com/google/javascript/clutz/partial/global_generic_type.js
```diff
@@ -0,0 +1,4 @@
...
+global.generic.type = null;
```
evmar: Can you add a test where it uses Map without template params?

LucasSloan: Done!

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2880,27 +2880,30 @@ public Void emitObjectType(
...
+        // Assume it's safe to emit the name as-is.
+        emit(name);
+        foundGlobalType = true;
       }
 
-      String newName = PlatformSymbols.CLOSURE_TO_TYPESCRIPT.get(name);
-      if (newName != null) {
-        emit(newName);
-        return true;
+      if (foundGlobalType
```
rkirov: This is error-prone because, if the foundGlobalType wasn't checked first calling 'emit' would be very bad.

How about this function is renamed to 'maybeGlobalType' and keep it emit free. Move the eemitGenericTypeArguments to the two callers.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2880,27 +2878,19 @@ public Void emitObjectType(
      * If a symbol refers to a "platform" type (e.g. Promise) emit it specially here and return
      * true.
```
rkirov: Change function description. Specifically call out that it may return null if not a platform symbol.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2880,27 +2878,19 @@ public Void emitObjectType(
...
      */
-    private boolean maybeEmitGlobalType(String name) {
-      if (name == null) return false;
-
+    private String maybeGlobalType(String name) {
```
rkirov: `maybeRenameGlobalType` ?

src/test/java/com/google/javascript/clutz/partial/global_generic_type.js
```diff
@@ -0,0 +1,8 @@
+goog.provide('global.generic.type');
+goog.provide('global.non.generic.type');
+
```
rkirov: There is one more weird case where closure returns type arguments as different type than NoType. So add a test for:

/** @type {!SomeType<!Map<string, string>>  */

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: Sure.

src/test/java/com/google/javascript/clutz/partial/global_generic_type.js
LucasSloan: Done, took major surgery, take a look at the diff.

PR #696, opened 2018-02-16
Add the mapping IArrayLike -> ArrayLike to the CLOSURE_TO_TYPESCRIPT platform symbols map.

LucasSloan:
Also while I was at it, I moved the platform symbol mappings that were in emitTemplatizedType() to use maybeRenameGlobalType, instead of using custom logic.


src/test/java/com/google/javascript/clutz/partial/to_object.d.ts
```diff
@@ -0,0 +1,14 @@
+declare namespace ಠ_ಠ.clutz.goog.array {
+  /**
```
rkirov: can you simplify the test to the its core issue. You can point to the goog.array.toObject in a comment https://github.com/google/closure-library/blob/master/closure/goog/array/array.js#L22

LucasSloan: Done.

PR #698, opened 2018-02-20
Add forward declared noResolvedTypes to typesUsed.

LucasSloan:
Works around an issue where an aliased type used as a type parameter to an unknown type doesn't resolve to the original type.  Adding the type to typesUsed ensures that the type alias is emitted as a type.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1738,10 +1738,12 @@ class DeclarationGenerator {
     private void emitNoResolvedTypeAsumingForwardDeclare(ObjectType type) {
       String displayName = maybeRewriteImportedName(type.getDisplayName());
       String maybeGlobalName = maybeRenameGlobalType(displayName);
-      displayName =
-          maybeGlobalName == null
-              ? Constants.INTERNAL_NAMESPACE + "." + displayName
-              : maybeGlobalName;
+      if (maybeGlobalName == null) {
+        typesUsed.add(displayName);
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1738,10 +1738,12 @@ private void emitNoResolvedTypeOrDefault(NoType type, String defaultEmit) {
     private void emitNoResolvedTypeAsumingForwardDeclare(ObjectType type) {
       String displayName = maybeRewriteImportedName(type.getDisplayName());
       String maybeGlobalName = maybeRenameGlobalType(displayName);
-      displayName =
-          maybeGlobalName == null
-              ? Constants.INTERNAL_NAMESPACE + "." + displayName
-              : maybeGlobalName;
+      if (maybeGlobalName == null) {
+        typesUsed.add(displayName);
```
rkirov: I am pretty sure it is safe to add the global types to used too, because the types used pass ignored externs. Can you try adding 'typesUsed.add' as the last statement in the function, similar to:

https://github.com/angular/clutz/blob/master/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java#L1003

LucasSloan: I think it'd be safe add the global types as well.  However, this if block (and the ternary it replaces) is doing 2 pieces of logic:

1. Is the displayName the maybeGlobalName?
2. Should the displayName be prefixed with the look of disapproval?

typesUsed assumes that the type isn't absolute, so I can't use the post-processed display name.  Given that, and that global types don't need to be added to types used, makes this the least disruptive thing I could think of.

PR #699, opened 2018-02-21
Support specialCaseMapEntries in incremental mode.

### LINKED ISSUE #697: specialCaseMapEntries does not work with incremental clutz
It appears we have special cased Map entries:
https://github.com/angular/clutz/blob/master/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java#L2472

but that hack doesn't work any longer with incremental clutz. Needs revisiting, don't see obvious reason why it can't work with iclutz.

rkirov:
specialCaseMapEntries does some type inspection to make sure this is
indeed the special cased method. Those were write with the assumption
that externs are present, which is not true in incremental mode.

Same checks are now done using type names instead of resolved symbols.
There is no loss of safety.

Fixes #697


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2493,14 +2493,20 @@ private void emitProperty(
...
-      TemplatizedType ttType = propertyType.getReturnType().toMaybeTemplatizedType();
-      ImmutableList<JSType> ttypes = ttType.getTemplateTypes();
-      if (!isTemplateOf(ttType, "IteratorIterable")
+
+      ImmutableList<JSType> ttypes = null;
+      JSType returnType = propertyType.getReturnType();
+
+      // In incremental mode this type is noResolveType, while in total mode it is TemplatizedType
+      // They both have getTemplateTypes, but neither extends the other in the class heirarchy.
+      if (returnType.isTemplatizedType())
```
evmar: I think the rule is curlies if it spans more than one line.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2493,14 +2493,20 @@ private void emitProperty(
...
         return false;
       }
-      TemplatizedType ttType = propertyType.getReturnType().toMaybeTemplatizedType();
-      ImmutableList<JSType> ttypes = ttType.getTemplateTypes();
-      if (!isTemplateOf(ttType, "IteratorIterable")
+
+      ImmutableList<JSType> ttypes = null;
+      JSType returnType = propertyType.getReturnType();
+
+      // In incremental mode this type is noResolveType, while in total mode it is TemplatizedType
```
evmar: noResolvedType

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2493,14 +2493,20 @@ private void emitProperty(
...
+      // They both have getTemplateTypes, but neither extends the other in the class heirarchy.
+      if (returnType.isTemplatizedType())
+        ttypes = ((TemplatizedType) returnType).getTemplateTypes();
+      if (returnType.isNoResolvedType()) ttypes = ((NoResolvedType) returnType).getTemplateTypes();
```
evmar: Maybe "else if" here to make it clearer that it's checking a few props of returnType?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: of course. It used to be single line but googleJavaFormatter agressively puts everything on new line.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #700, opened 2018-02-21
Add an option to set compiler tracer mode from command line.

roydor:
This enables outputting tracer data from the js compiler which will let us improve performance by tweaking compiler or jvm options.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -142,6 +142,16 @@ public class Options {
   )
   String collidingProvidesFile = null;
 
+  @Option(
+    name = "--tracer_mode",
+    hidden = true,
+    usage =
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -142,6 +142,16 @@ public int parseArguments(Parameters params) throws CmdLineException {
   )
   String collidingProvidesFile = null;
 
+  @Option(
+    name = "--tracer_mode",
+    hidden = true,
+    usage =
+        "Shows the duration of each compiler pass and the impact to "
```
rkirov: Interesting, I have never used this. Does it have any other documentation you can point off from there (i.e. official closure docs)?

roydor: Done

PR #701, opened 2018-02-22
Fix emit for declareLegacyNamespace() files.

LucasSloan:
declareLegacyNamespace() files treat their exports as goog.provide, but the depgraph treats them as goog.modules.  To resolve this ambiguity, emit aliases for their exports in the goog.module style.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -677,6 +688,29 @@ private void processUnprovidedTypes(Set<String> provides, Set<String> transitive
     }
   }
 
+  /**
+   * If any inputs declare a legacy namespace, emit aliases for their exports in goog.module style.
```
rkirov: OOC, why it is better to emit both goog.provide and goog.module (through alias), vs just emit solely in goog.module style.

I have a hunch that codewise what you currently have is easier, but want to hear your opinion here.

src/main/java/com/google/javascript/clutz/LegacyNamespaceReexportMapBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+    if (localModuleId == null) {
+      return reexportMap;
+    }
+
+    if (!isLegacyNamespaceModule(moduleBody)) {
+      return reexportMap;
+    }
+
+    for (Node statement : moduleBody.children()) {
+      if (isWholeModuleExportAssignment(statement)) {
```
rkirov: I feel like I have seen this code in AliasMapBuilder, is refactoring commonalities feasable?

rkirov: In general it would be nice to hide the `getFirstChild().getChildAtIndex(1)` type of AST mangling inside more meaningfully named reusable methods like what you did with isNamedExportAssignment.

src/test/java/com/google/javascript/clutz/partial/declare_legacy_namespace.js
```diff
@@ -0,0 +1,7 @@
+
+goog.module('declare.legacy.namespace');
+goog.module.declareLegacyNamespace();
+
+class Class {}
+
+exports = Class;
```
rkirov: add a test for named export too, code does have a branch on that.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: We have to emit both styles, since declareLegacyNamespace() allows you to import in a goog.provide() file without the goog.scope() and goog.module.get() workaround.  There's a more complicated thing we could do, which is replace the module$contents$foo initial definitions of the types with module$exports$foo definitions, but that'd be much harder than this.

src/test/java/com/google/javascript/clutz/partial/declare_legacy_namespace.js
LucasSloan: We already have one - goog_legacy_namespace_exporter.d.ts (below) is the result of just such a circumstance.

src/main/java/com/google/javascript/clutz/LegacyNamespaceReexportMapBuilder.java
LucasSloan: Done!

src/test/java/com/google/javascript/clutz/partial/declare_legacy_namespace.js
```diff
@@ -0,0 +1,7 @@
+
+goog.module('declare.legacy.namespace');
+goog.module.declareLegacyNamespace();
+
+class Class {}
```
rkirov: Because you are using `visitKnownTypeValueAlias` without any guarantees that the symbols are really are of the type-and-value variety (classes, enums), I don't expect this code to work correctly with interfaces. That said, lets leave it for a follow-up PR, and add a TODO (or even add interface in this test and a comment in the output "this is wrong, will fix later").

LucasSloan: Added a comment.

PR #702, opened 2018-02-24
Fixes emit in partial mode on module local typedefs used in generics.

rkirov:
Usually closure inlines typedefs, but in partial mode with generics
involved it only resolves the name to $module$contents. Our secondary
types used pass has to deal with finding the typedef definition and
emitting it correctly.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/partial/inner_typedef_npe.d.ts
+++ b/src/test/java/com/google/javascript/clutz/partial/inner_typedef_npe.d.ts
@@ -1,0 +1,15 @@
+declare namespace ಠ_ಠ.clutz {
+  class module$exports$odd$npe$typedef extends module$exports$odd$npe$typedef_Instance {
+  }
+  class module$exports$odd$npe$typedef_Instance {
+    private noStructuralTyping_: any;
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/partial/inner_typedef_npe.d.ts
```diff
@@ -0,0 +1,15 @@
...
+    private noStructuralTyping_: any;
+    foo ( ) : ಠ_ಠ.clutz.some.type < ಠ_ಠ.clutz.module$contents$odd$npe$typedef_ATypeDef > ;
+  }
+}
+declare module 'goog:odd.npe.typedef' {
+  import alias = ಠ_ಠ.clutz.module$exports$odd$npe$typedef;
+  export default alias;
+}
+declare namespace ಠ_ಠ.clutz {
+  type module$contents$odd$npe$typedef_ATypeDef = { a : string } ;
```
LucasSloan: What would we have seen with the old emit here?

rkirov: This block would have been missing. Clutz would have crashed with NPE because emitMissingTypes would be working with a TypedVar object that has `.getType()` returning `null` as it does for `typedefs`.

rkirov:
PTAL, refactored around the existing `declareTypedef` code to support this usecase

PR #703, opened 2018-02-27
Incremental clutz should not add root depgraph externs to the compila…

rkirov:
…tion externs.

Depgraph roots can be exaggerated due to a mechanism of "exporting" for
js_library bazel rules. Both sources and externs should only be
filtering with the depgraph roots and not adding new targets. So far
this was true for sources and this change will make it true for externs.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/OptionsTest.java
+++ b/src/test/java/com/google/javascript/clutz/OptionsTest.java
@@ -31,24 +31,35 @@ public class OptionsTest {
               "javascript/closure/other_file_not_in_depgraph.js",
               "javascript/closure/string/string.js",
               "blaze-out/blah/my/blaze-out-file.js",
+              "--depgraphs",
+              DepgraphTest.DEPGRAPH_PATH.toFile().toString(),
+              "--depgraphs_filter_sources",
+            });
+    // Arguments are filtered by what appears in the depgraph.
+    assertThat(opts.arguments)
+        .containsExactly(
+            "javascript/closure/string/string.js", // kept as a root in depgraph
+            "blaze-out/blah/my/blaze-out-file.js") // kept as a non-root in depgraph
+        .inOrder();
+  }
+
+  @Test
+  public void testFilterSourcesWithExternsInDepgraphs() throws Exception {
+    Options opts =
+        new Options(
+            new String[] {
+              "javascript/closure/string/string.js",
               "--externs",
               "extern1.js",
               "extern2.js",
               "--depgraphs",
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/clutz/OptionsTest.java
```diff
@@ -146,4 +146,20 @@ public void testGoogProvides() throws Exception {
...
+  @Test
+  public void testPartialInputIgnoresDepgraphRootExternsIfNotPassedToExterns() throws Exception {
+    // Due to "exported" libraries, what the depgraph considers "root" can be incorrect for the
+    // purposes of incremental clutz. Arguments and Externs lists should only be filtered down
+    // with depgraph info and never extended.
+    Options opts =
+        new Options(
+            new String[] {
+              "--partialInput",
+              "javascript/common/dom.js",
```
LucasSloan: You've removed this from the depgraph - I think you can just replace with foo.js.

src/test/java/com/google/javascript/clutz/OptionsTest.java
```diff
@@ -104,11 +104,11 @@ public void testExternsShouldBeUnique() throws Exception {
         new Options(
             new String[] {
               "--externs",
-              "javascript/common/dom.js",
+              "my/root/extern.js",
               "--depgraphs",
               DepgraphTest.DEPGRAPH_PATH.toFile().toString(),
```
LucasSloan: Add a comment saying that depgraph has my/root/extern.js and my/nonroot/extern.js

src/test/java/com/google/javascript/clutz/OptionsTest.java
```diff
@@ -46,7 +46,7 @@ public void testFilterSourcesDepgraphs() throws Exception {
             "javascript/closure/string/string.js", "blaze-out/blah/my/blaze-out-file.js")
         .inOrder();
     assertThat(opts.externs)
-        .containsExactly("javascript/common/dom.js", "extern1.js", "extern2.js")
+        .containsExactly("my/root/extern.js", "my/nonroot/extern.js", "extern1.js", "extern2.js")
```
LucasSloan: I don't like this test.  Its name says it's about filtering sources, but it has a bunch of other assertions.  Could you rename the test or split it up?

rkirov: Generally, random refactoring is best left for other PRs, but these are tests so pretty harmless.

Split into sources assertions and extern assertions. Output assertions are not needed (not affected by depgraph_filter_sources).

rkirov:
PTAL

PR #705, opened 2018-03-01
Only visit members of object literals where assignments are of names, not expressions.

LucasSloan:


src/test/java/com/google/javascript/clutz/partial/declare_legacy_namespace_object_literal_export.js
```diff
@@ -5,4 +5,10 @@ class Class1 {}
...
 class Class2 {}
 
-exports = {Class1, RenamedClass2: Class2};
\ No newline at end of file
+exports = {
+  Class1,
+  RenamedClass2: Class2,
+  //!! Inline expression won't show up in the declareLegacyNamespace, since it
+  //!! doesn't have a predictable name
+  Class3: class {}
```
rkirov: Use something other than class expr, `aConstant: 2 + 2` or `aFunction: function() {}`

PR #706, opened 2018-03-14
Fix emit for optional properties on export objects.

LucasSloan:
We use visitProperty() for visiting object properties and for visiting certain namespace members.  If we're visiting a namespace member that is a type union including undefined, the old code would incorrectly use the question mark syntax for optionality.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1010,7 +1010,8 @@ class DeclarationGenerator {
           } else {
             emit("function");
           }
-          treeWalker.visitProperty(propName, oType, false, false, Collections.<String>emptyList());
+          treeWalker.visitProperty(
+              propName, oType, false, false, true, Collections.<String>emptyList());
         }
       }
     }
@@ -2355,7 +2356,12 @@ class DeclarationGenerator {
           continue;
         }
         visitProperty(
-            propName, objType, isStatic, forceProps.contains(propName), classTemplateTypeNames);
+            propName,
+            objType,
+            isStatic,
+            forceProps.contains(propName),
+            false,
+            classTemplateTypeNames);
       }
     }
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/partial/optional_export_object_property.js
```diff
@@ -0,0 +1,17 @@
+goog.module('optional.export.object.property');
+goog.module.declareLegacyNamespace();
```
rkirov: just goog.provide if you really need a provide here. Even better if just a goog.module works for the repro.

src/test/java/com/google/javascript/clutz/partial/optional_export_object_property.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * @return {number|undefined}
+ */
+function getNumberOrUndefined() {}
```
rkirov: fn is not exported thus not seen in the output .d.ts. I am guessing you needed it to get around closure's type inference, but explain that in a comment (use //!! to skip emit on it).

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2479,7 +2493,8 @@ private void emitProperty(
             && unionType
                 .getAlternatesWithoutStructuralTyping()
                 .stream()
-                .anyMatch(JSType::isVoidType)) {
+                .anyMatch(JSType::isVoidType)
+            && !isNamespace) {
```
rkirov: Took me second to realize that 'var optional?: number' in not a valid ts syntax. Add a comment why isNamespace check is needed for syntactic correctness.

src/test/java/com/google/javascript/clutz/partial/optional_export_object_property.js
LucasSloan: Unfortunately not.  Neither of those get the right code path through clutz.

src/test/java/com/google/javascript/clutz/partial/optional_export_object_property.js
LucasSloan: Yep.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #712, opened 2018-03-19
Remove [Symbol.iterator()] from incremental clutz.

rkirov:
It appears that records extending unknown records spuriously trigger
that branch. Since the use case of extending records is a lot more
common than Iterator classes/interfaces, I am turning off this branch
for incremental clutz.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2394,23 +2394,31 @@ class DeclarationGenerator {
      *
      * <p>JSCompiler does not understand nor represent Symbol properties, so we cannot just emit the
      * property in the loop above, and must guess on the actual return type of the iterator method.
-     *
-     * <p>TODO(rado): This code will almost never work with iclutz, because IterableIterator and the
-     * class that implements it would be part of different complitation. Investigate whether this
-     * can be syntactically resolved.
      */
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2396,21 +2396,25 @@ private void visitProperties(
...
       JSType implemented;
       String returnType;
       if (iteratorIterableType != null && instanceType.isSubtype(iteratorIterableType)) {
         implemented = iteratorIterableType;
         returnType = "IterableIterator";
-      } else if (iterableType != null && instanceType.isSubtype(iterableType)) {
+        // It appears that iterableType is always defined. However, a simple extending unknown
+        // record is reported as subtype of iterable for unknown reasons, so in partial mode we
+        // do not emit the iterator signature.
+      } else if (iterableType != null
```
LucasSloan: The type doesn't have any information that it's a record?

rkirov: It does. Are you suggesting narrowing the exception to record types only? I guess I can try that tomorrow.

rkirov: Spend some time looking into better checks specifically for interface/record, but for some unknown reason `instanceType.isInterface()` returns `false`. Added some comments and tests for interfaces too, but the internal logic is the same.

PR #714, opened 2018-03-22
Fixes the regression on the static-instance split for classes.

LucasSloan:
Also reworks support for declareLegacyNamespace() to ensure that there's an alias for the instance half.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -714,9 +714,12 @@ class DeclarationGenerator {
           googModuleStyleName = e.getKey();
         }
         emitNamespaceBegin(namespace);
-        // TODO(lucassloan): not technically correct, since this emits an alias as if it's a class
-        // with a type and value. Change to emit the alias only for type or value as necessary.
-        visitKnownTypeValueAlias(googModuleStyleName, e.getValue());
+        TreeWalker treeWalker =
+            new TreeWalker(compiler.getTypeRegistry(), new HashSet<>(), false, false);
+        JSType type = compiler.getTopScope().getOwnSlot(e.getValue()).getType();
+        if (type != null && isDefiningType(type)) {
+          treeWalker.visitTypeValueAlias(googModuleStyleName, type.toMaybeObjectType());
+        }
         emitNamespaceEnd();
       }
     }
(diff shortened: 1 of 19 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1069,6 +1070,123 @@ private void visitKnownTypeValueAlias(String unqualifiedName, String alternative
...
+      // TS type aliases are only useful in type positions.
+      // To emulate closure alias semantics, introduce also an aliased constructor
+      emit("var " + unqualifiedName);
+      emit(":");
+      emit("typeof");
+      emit(emitName);
+      emit(";");
+      emitBreak();
+    }
+    if (!otype.isInterface() && otype.isFunctionType()) {
```
LucasSloan: Most of this green block was just copy and pasted to make it visible, but this if statement is new.

rkirov: Ad-hoc pulling some visit* methods out of TreeWalker is not a good idea. Can you introduce a new TreeWalker entry point so that we can to `new TreeWalker(...).specialWalk()`. It would make it easier for me to review this change too.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -714,9 +714,12 @@ private void declareLegacyNamespaceAliases() {
           googModuleStyleName = e.getKey();
         }
         emitNamespaceBegin(namespace);
-        // TODO(lucassloan): not technically correct, since this emits an alias as if it's a class
-        // with a type and value. Change to emit the alias only for type or value as necessary.
-        visitKnownTypeValueAlias(googModuleStyleName, e.getValue());
```
rkirov: I have concerns that previously we could emit here purely based on the name, whereas now we require type information. If there is a case where the type info is not present for some reason we could potentially regress here. 

That said, I don't have a good idea how could this happen, so let's see what happens during presubmit.
No action requred.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1779,7 +1801,8 @@ private void visitTypeDeclaration(JSType type, boolean isVarArgs, boolean isOpti
      * the literal string that was in the original code. Special care is taken for templatized
      * types.
      */
-    private void emitNoResolvedTypeOrDefault(NoType type, String defaultEmit) {
+    private void emitNoResolvedTypeOrDefault(
```
rkirov: Add some documentation for `emitInstanceForObject`.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
LucasSloan: Yep, needs a presubmit.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

PR #716, opened 2018-03-22
Treat goog.forwardDeclare() as an import for the purposes of building the import rename map.

LucasSloan:


src/test/java/com/google/javascript/clutz/partial/legacy_namespace_forward_declare.js
```diff
@@ -0,0 +1,13 @@
+goog.module('legacy.namespace.forward.declare');
+goog.module.declareLegacyNamespace();
```
rkirov: I keep forgetting, is the `declareLegacyNamespace` crucial to this test? Can you annotate the .d.ts output like 

//!! this block is needed for ...
declare namespace ಠ_ಠ.clutz.legacy.namespace.forward {

//!! this block is needed for
declare namespace ಠ_ಠ.clutz { ... }

src/test/java/com/google/javascript/clutz/partial/legacy_namespace_forward_declare.js
```diff
@@ -0,0 +1,13 @@
...
+
+const MissingForwardDeclare = goog.forwardDeclare('missing.forward.Declare');
```
rkirov: Does it matter if 'missing.forward.Declare' is a goog.module or goog.provide? Is this fix assuming one or the other?

src/test/java/com/google/javascript/clutz/partial/legacy_namespace_forward_declare.js
LucasSloan: It is necessary, but in a tricky way.  I added a comment to the .d.ts.

src/test/java/com/google/javascript/clutz/partial/legacy_namespace_forward_declare.js
LucasSloan: Our usual methodology for determining that will still work.  In this case, since we passed in no depgraphs, it defaults to goog.module.

src/test/java/com/google/javascript/clutz/partial/legacy_namespace_forward_declare.d.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+declare namespace ಠ_ಠ.clutz {
+  class module$contents$legacy$namespace$forward$declare_ClassExtendingMissingForwardDeclare extends module$contents$legacy$namespace$forward$declare_ClassExtendingMissingForwardDeclare_Instance {
+  }
+//!! Without including forward declares as imports, ಠ_ಠ.clutz.module$exports$missing$forward$Declare_Instance would be ಠ_ಠ.clutz.module$contents$missing$forward$Declare
+  class module$contents$legacy$namespace$forward$declare_ClassExtendingMissingForwardDeclare_Instance extends ಠ_ಠ.clutz.module$exports$missing$forward$Declare_Instance {
+    constructor ( ) ;
+  }
+}
+declare namespace ಠ_ಠ.clutz {
```
rkirov: I meant add //!! for this block, something like "//!! Upstream consumers expect goog.module-shaped import, but we have generated goog.provide-shaped. This block links the two."

But maybe that is too obvious?

rkirov:
Does this mean 'goog.forwardDeclare' just works when there is no 'legacyNamespace'. Would a test for that help?

LucasSloan:
Nevermind, I wasn't looking hard enough at the diff when I tested goog.module.  It fails for that as well, I'll update the test.

PR #717, opened 2018-03-23
Allow clutz to handle ES6 generator functions.

rkirov:
In closure generator functions implicitly are defined as functions that
return a Generator<T> type. In TypeScript, the closest thing to
Generator type is the IterableIterator.

The difference boils down to the the two methods return and throw, that
are part of the Generator interface in closure, but in TS they are
optional methods on the Iterator<T> type.

This change makes it so that we emit Genarator as IterableIterator in
the .d.ts. Tests added for both partial and total mode.

Special care for not emitting builtins Iterator and Generator, which
appear even if the externs they are defined in are not passed to the
compilation.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/generator_with_platform.d.ts
+++ b/src/test/java/com/google/javascript/clutz/generator_with_platform.d.ts
@@ -1,0 +1,10 @@
+declare namespace ಠ_ಠ.clutz.module$exports$generators_total {
+  //!! AFAICT, closure just infers ? and we faithfully translate that to 'any'.
+  function inferredG ( ) : any ;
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/javascript/clutz/generator_with_platform.d.ts
```diff
@@ -0,0 +1,10 @@
+declare namespace ಠ_ಠ.clutz.module$exports$generators_total {
+  //!! AFAIKT, closure just infers ? and we faithfully translate that to 'any'.
```
evmar: AFAICT (no K)

PR #718, opened 2018-03-23
Fix an npe in declareLegacyNamespacesAliases().

LucasSloan:


src/test/java/com/google/javascript/clutz/partial/optional_export_object_property.d.ts
```diff
@@ -5,5 +5,3 @@ declare module 'goog:optional.export.object.property' {
   import alias = ಠ_ಠ.clutz.optional.export.object.property;
   export = alias;
 }
-declare namespace ಠ_ಠ.clutz {
```
rkirov: what happened here?

LucasSloan: I moved the lines that created the namespace block inside the if statement.

LucasSloan:
Added a test, PTAL

src/test/java/com/google/javascript/clutz/partial/declare_legacy_namespace_goog_define.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace ಠ_ಠ.clutz.legacy.namespace.goog {
```
rkirov: Ops, we probably should be emitting this, add '//!! TODO'

PR #719, opened 2018-03-23
Extract enum values with AST and put them in the declaration files.

### LINKED ISSUE #663: Clutz should use individual enum values to create *.d.ts files

Goodwine:
Fixes #663

This should allow using the `checkExhaustive` function in switch statements for JSPB enums.
I will follow up with doing the same thing for string enums if this PR is ok.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -23,6 +23,7 @@ import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.Sets;
+import com.google.common.collect.Streams;
 import com.google.common.io.Files;
 import com.google.javascript.jscomp.AbstractCommandLineRunner;
 import com.google.javascript.jscomp.CompilerInput;
@@ -67,8 +68,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import javax.annotation.Nullable;
 import org.kohsuke.args4j.CmdLineException;
 
@@ -1399,7 +1402,7 @@ class DeclarationGenerator {
       } else {
         maybeEmitJsDoc(symbol.getJSDocInfo(), /* ignoreParams */ false);
         if (type.isEnumType()) {
-          visitEnumType(emitName, emitName, (EnumType) type);
+          visitEnumType(emitName, emitName, (EnumType) type, symbol.getNode());
           return;
         }
         if (isTypedef(type)) {
@@ -1712,7 +1715,8 @@ class DeclarationGenerator {
       emitBreak();
     }
 
-    private void visitEnumType(String symbolName, String qualifiedName, EnumType type) {
+    private void visitEnumType(
+        String symbolName, String qualifiedName, EnumType type, Node node) {
       // Enums are top level vars, but also declare a corresponding type:
       // <pre>
       // /** @enum {ValueType} */ var MyEnum = {A: ..., B: ...};
@@ -1750,8 +1754,13 @@ class DeclarationGenerator {
         emit("{");
         emitBreak();
         indent();
-        for (String elem : sorted(type.getElements())) {
+        Map<String, String> elements = Streams.stream(node.getNext().children())
+            .collect(Collectors.toMap(
+              Node::getString, n->String.valueOf(n.getFirstChild().getDouble())));
+        for (String elem : sorted(elements.keySet())) {
           emit(elem);
+          emit("=");
+          emit(elements.get(elem));
           emit(",");
           emitBreak();
         }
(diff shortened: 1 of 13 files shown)


Goodwine:
I will follow up with doing the same thing for string enums if this PR is ok.
@rkirov, this is a follow up for my failed attempt at #664

src/test/java/com/google/javascript/clutz/provide_single_class.js
```diff
@@ -54,6 +54,6 @@ foo.bar.Baz.NestedClass = function() {};
 
 /** @enum */
 foo.bar.Baz.NestedEnum = {
-  A: 1,
+  XD: 1,
```
rkirov: what was the purpose of change to XD here? Maybe order? add a comment

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1754,8 +1758,15 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
         emit("{");
         emitBreak();
         indent();
-        for (String elem : sorted(type.getElements())) {
+        Map<String, String> elements =
+            Streams.stream(node.getNext().children())
+                .collect(
+                    Collectors.toMap(
+                        Node::getString, n -> String.valueOf(n.getFirstChild().getDouble())));
+        for (String elem : sorted(elements.keySet())) {
```
rkirov: why sorted? Isn't it best to emit in original order?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2883,25 +2899,33 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
       // TODO(martinprobst): This curiously duplicates visitProperty above. Investigate the code
       // smell and reduce duplication (or figure out & document why it's needed).
 
-      Set<NamedTypePair> innerProps = new TreeSet<>();
+      Map<NamedTypePair, Node> innerProps = new TreeMap<>();
```
rkirov: Add a comment that reminds us that 'Node' so far is only used for enum emitting, otherwise it is not clear why we collect it.

src/test/java/com/google/javascript/clutz/provide_single_class.js
Goodwine: Initially for testing because I wasn't sure changes were being picked up, but then I decided to leave it to test order. (but I can revert this if I remove sorting given the message below)

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
Goodwine: It was sorted before, so I just kept the same functionality, I can remove this if you want.

I *think* it was being sorted because the values came from a `Set` and maybe they were generated in different orders, I guess that's not needed anymore

rkirov: ah, right. Ok to keep it as it is then.

rkirov:
Oh, I just remembered. I think it is valid closure to use non-literal initializers (though rare). Can you add a test for those and make sure clutz doesn't barf on them:

```
/** @enum {number} */
let E = {
 A: 1 + 1,
 B: f(0)
}
```

Btw, we have some code in gents to deal with the AST bits for enums (poke around).

Goodwine:
I'll try to apply the changes you suggested. I actually had the checks in place before, but I reverted them because I didn't think that that was valid closure code.

However I sync'd my fork to HEAD and for whatever reason tests don't pass even when checking out to upstream/master :/

Goodwine:
PTAL

- I added the comment you mentioned about sorting enum values (XD before B in JS file, but then sorted on D.TS file)
- I added the test cases for functions and expressions and updated the code to only emit when the value node `isNumber`
- My git-fu is not great, I couldn't figure out how to revert the bad merge and I ended up wrecking the commit history on my branch, this probably doesn't matter for the pull request except that previous comments seem to have been lost :/

rkirov:
No worries, I will clean up the PR when merging and squash into a single commit. LGTM

PR #720, opened 2018-03-23
Global types don't have the static/instance split, so don't apply the…

LucasSloan:
… _Instance suffix to them.


LucasSloan:
Sure!

rkirov:
Alternatively we could have introduced _Instance for Event too, for completeness, but this is probably a better fix.

PR #721, opened 2018-03-23
add Float32Array and 64 to list of global workarounds

evmar:
This library defines goog.vec.Float32Array
  https://github.com/google/closure-library/blob/17b6f4d1142d931dcdc5845475af288d5d13d447/closure/goog/vec/float32array.js#L28
while this one in the same namespace wants to refer to the global one
  https://github.com/google/closure-library/blob/17b6f4d1142d931dcdc5845475af288d5d13d447/closure/goog/vec/vec.js#L43
so it causes shadowing in Clutz.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/PlatformSymbols.java
+++ b/src/main/java/com/google/javascript/clutz/PlatformSymbols.java
@@ -63,7 +63,15 @@ public class PlatformSymbols {
    * Adding a symbol to this list also suppresses the duplicate externs.
    */
   public static final ImmutableSet<String> GLOBAL_SYMBOL_ALIASES =
-      ImmutableSet.of("Element", "Error", "Event", "EventTarget", "Object", "Date");
+      ImmutableSet.of(
+          "Date",
+          "Element",
+          "Error",
+          "Event",
+          "EventTarget",
+          "Float32Array",
+          "Float64Array",
+          "Object");
 
   /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -63,7 +63,15 @@
...
-      ImmutableSet.of("Element", "Error", "Event", "EventTarget", "Object", "Date");
+      ImmutableSet.of(
+          "Date",
+          "Element",
+          "Error",
+          "Event",
+          "EventTarget",
+          "Float32Array",
+          "Float64Array",
+          "Object");
```
rkirov: Don;t you need to add them here too - https://github.com/angular/clutz/blob/master/src/resources/closure.lib.d.ts#L16

evmar: Hah, what I had worked because it effectively made the code into implicit any :(

src/resources/closure.lib.d.ts
```diff
@@ -2,6 +2,8 @@
 // All clutz namespaces are below ಠ_ಠ.clutz, thus
 // this acts as global.
 declare namespace ಠ_ಠ.clutz {
+  type GlobalDate = Date;
+  var GlobalDate: typeof Date;
```
evmar: I re-alphabetized this and put each pair as type then value.

PR #728, opened 2018-04-09
Upgrade to Truth 0.40 library.

rkirov:
Upstreaming from an internal clean up to Truth library usage.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/ProgramSubject.java
+++ b/src/test/java/com/google/javascript/clutz/ProgramSubject.java
@@ -96,7 +96,7 @@ class ProgramSubject extends Subject<ProgramSubject, ProgramSubject.Program> {
           && expected.equals(expectedClean)) {
         Files.write(stripped, golden, Charsets.UTF_8);
       } else {
-        failComparing("compilation result doesn't match", expected, stripped);
+        check().that(stripped).isEqualTo(expected);
       }
     }
   }
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ dependencies {
   compile 'com.google.guava:guava:20.0'
 
   testCompile 'junit:junit:4.11'
-  testCompile 'com.google.truth:truth:0.39'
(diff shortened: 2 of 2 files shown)


src/test/java/com/google/javascript/clutz/ProgramSubject.java
```diff
@@ -96,7 +96,7 @@ void generatesDeclarations(File golden) throws IOException {
           && expected.equals(expectedClean)) {
         Files.write(stripped, golden, Charsets.UTF_8);
       } else {
-        failComparing("compilation result doesn't match", expected, stripped);
```
evmar: failComparing makes it show a diff, does the new thing?

I think I made a similar change elsewhere, where I found that failComparing had just moved locations in the API

rkirov: It shows as:

```
com.google.javascript.clutz.DeclarationGeneratorTests > colliding_provide.js FAILED
    expected:<...rt default alias;
    }
    [// some new text]> but was:<...rt default alias;
    }
    []>
```

and I still get my see diff link in Intellij.

rkirov: Tested old emit for the same error:
```
    compilation result doesn't match expected:<...rt default alias;
    }
    [// some new text]> but was:<...rt default alias;
    }
    []>
``` 
basically the same just text has changed.

evmar: Can you make the CL more descriptive? e.g. "failComparing is deprecated" or some such?

evmar: Also the internal CL says we need to change truth versions for this, have we done that?

rkirov: Upgraded to Truth 0.40 (thought it didn't seem needed for the change). PTAL

PR #729, opened 2018-04-13
Replace deprecated traverseEs6 method with traverse.

rkirov:
No functional change.


MatrixFrog:
Yeah, try again after the next MOE push (which should be about one day) or failing that, the next release (which should be in a couple weeks). This isn't urgent, I just want to clean up deprecated stuff rather than forget about it and leave it around indefinitely.

Issue #643, opened 2017-12-25
Nullable function aliases incorrectly translated


rkirov:
When gents sees

```javascript
/** @type {?function():string} */
let f = null;
```
It translates the type to `() => string|null`. However, due to union binding tighter than `=>` in TypeScript, means that this is read as 'a non-nullable function that returns string or null'.

For example, the following is a type error in TS:
```typescript
export type FnT = () => string | null;
let f: FnT = null;
```

The correct emit would be `(() => string) | null`.

bowenni:
Not only nullable functions but also undefinable functions.
```js
function()|undefined
```
emits
```js
() => AnyDuringMigration | undefined
```
instead of
```js
(() => AnyDuringMigration) | undefined
```

PR #737, opened 2018-05-08
Fix precedence of "=>" and "|".

bowenni:
Fixes https://github.com/angular/clutz/issues/643


src/test/java/com/google/javascript/gents/singleTests/fn_types.js
```diff
@@ -47,3 +47,9 @@ var restParamsTyped = function(n, br) {};
  * @type {function(number, boolean=, ...): number}
  */
 const complex = function(n, o, r) { return n; };
+
+/** @type {?function():string} */
+let f1 = null;
+
+/** @type {function()|string} */
+const f2 = 'string';
```
rkirov: Add something trickier like

`function():string|function():string` -> `() => string | () => string`

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -159,6 +169,14 @@ boolean maybeOverrideCodeGen(Node n) {
           return true;
         }
         return false;
+      case FUNCTION_TYPE:
```
rkirov: I don't understand why the code that adds "(" and the one that adds ")" have to be so far apart. How can you convince me that we produce balanced parens?

src/test/java/com/google/javascript/gents/singleTests/fn_types.js
bowenni: That's just `(function():string)|(function(): string)`, which is `function():string`.
Did you mean `function():(string|function():string)`?

I added test cases for both anyway.

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
bowenni: The way `maybeOverrideCodeGen` works is that if it returns true then no further code generation on the node is needed. We cannot use that for a `FUNCTION_TYPE` node otherwise we have to duplicate all the code gen logic for `FUNCTION_TYPE` which contains many invocations to private APIs.
So we have to rely on the default code gen for `FUNCTION_TYPE`, i.e. `maybeOverrideCodeGen` must return false for `FUNCTION_TYPE`. Therefore, to add a pair of `()` we need to add `(` before the code gen and `)` after the code gen.

I added some comments explaining this.

Issue #650, opened 2018-01-03
Gents doesn't emit parameter properties when the parameters have default values


bowenni:
We emit
```ts
class C {
  constructor(public readonly a: string) {}
}
```
for
```js
class C {
  /** @param {string} a */
  constructor(a) {
    /** @const */
    this.a = a;
  }
}
```
But when the parameter `a` has a default value we don't emit parameter properties any more.
For example, given
```js
class C {
  /** @param {string} a */
  constructor(a = 'Hello') { // Default value
    /** @const */
    this.a = a;
  }
}
```
we emit
```ts
class C {
  a: any;
  constructor(a = 'Hello') {
    this.a = a;
  }
}
```
instead of the below desired output:
```ts
class C {
  constructor(public readonly a: string = 'Hello') {}
}
```

Issue #731, opened 2018-04-25
gents does not convert rest parameters correctly


rkirov:
The following pattern:

```
/**
 * @param {...string|Array<string>} args
 */
function (...args) { ... }
```
got translated to 

```
function(...args) {...}
```
without any type annotations, which triggered 'noImplicitAny' errors.

PR #738, opened 2018-05-08
Add support for collapsing parameter properties with default values.

bowenni:
Fixes https://github.com/angular/clutz/issues/650

One side-effect of this change is that it adds type annotations to function parameters with default values which fixes the implicity any errors.

Another side-effect: since this change adds type annotations, an optional parameter with default values will be migrated as
```
f(p?: type = value) {
```
which is a ts compiler error. But I think that's not too bad. There's no reasons to use both, I think.
Besides, we previously completely throw away the type annotation in this case anyways:
```
f(p = value) {
```


bowenni:
https://github.com/angular/clutz/issues/731 is closely related to this issue (rest parameters instead of parameters with default values).

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -162,6 +162,19 @@ public void visit(NodeTraversal t, Node n, Node parent) {
             maybeSetInlineTypeExpression(parent, n, bestJSDocInfo, true);
           }
           break;
+          // If a DEVAULE_VALUE is in a PARAM_LIST we type annotate its first child which is the
+          // actual parameter.
+        case DEFAULT_VALUE:
```
rkirov: default values can occur in ES6 destructuring too:

let {x = 0} = {};

Are sure this makes sense for those cases?

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -308,12 +308,6 @@ public void visit(NodeTraversal t, Node n, Node parent) {
             if (!nodeAfterDefault.isName()) {
               continue;
             }
-            // It appears that adding ACCESS_MODIFIERs to Default params do not come out though
-            // the CodeGenerator, thus not safe to remove the declaration.
-            // TODO(rado): fix in emitting code and remove this line.
```
rkirov: Thanks, for cleaning up my TODOs!

src/test/java/com/google/javascript/gents/singleTests/module_class3.js
```diff
@@ -5,7 +5,7 @@ class Klass {
 
   /** 
    * @param {number} n
-   * @param {Array<?>=} list
```
rkirov: Why delete the '='? I looked around and it appears closure allows one to write this with or without the '=', but the one with '=' is more common.

bowenni: Reverted.

src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -309,16 +309,16 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
-            if (param.isDefaultValue()) {
-              continue;
-            }
             String paramName = nodeAfterDefault.getString();
             @Nullable
             JSTypeExpression paramType =
                 constructorJsDoc == null ? null : constructorJsDoc.getParameterType(paramName);
+            // The class member declaration on "this" will not have "=" in the JSDoc most of the time. Therefore we need to remove the "=" from the JSDoc of the parameter declaration so we can compare the types and collapse the parameter property.
```
rkirov: comment goes over 100+ chars

bowenni: Done.
`gradle googleJavaFormat` stopped working for me.

bowenni: Now I look at that, the code passed on travis so maybe our formatter setup is not working.

PR #746, opened 2018-05-14
Do not consider simpler default obj lit exports as default exports.

### LINKED ISSUE #527: exporting an object literal can be translated better than default exports
This pattern

```javascript
goog.module('a.b.c');
const x = 0;
exports = {x: x};
```

is translated to 

```javascript
const x = 0;

export const c = {
  x: x
};
```

Because that is how gents handles default exports in general.

However, this output would be preferable and closer to the original intent: 

```javascript
export const x = 0;
```

Clearly, this can only be done for static object literal with a simple `{nameA: nameA, nameB: nameB}` (names need to match in order to maintain export symbol).

Reported by @mgr01, from what I hear this is common enough to be worth the extra effort in rewriting.

rkirov:
Gents tries to not use default exports to account for an internal style
rule to not use default exports. However, some default exports are
trivial obj literals that just safe typing for the author.

class C {}
class D {}

exports = {C, D};

After this change gents will translate this file as if it was writen in
the equivalent style of:

exports.C = C;
exports.D = D;

Closes #527

### DIFF
--- a/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/ModuleConversionPass.java
@@ -636,30 +636,29 @@ public final class ModuleConversionPass implements CompilerPass {
     Node rhs = assign.getLastChild();
     JSDocInfo jsDoc = NodeUtil.getBestJSDocInfo(assign);
 
-    ExportedSymbol symbolToExport =
-        ExportedSymbol.fromExportAssignment(rhs, exportedNamespace, exportedSymbol, fileName);
-
     if (lhs.matchesQualifiedName(exportedNamespace)) {
       rhs.detach();
-      Node exportSpecNode;
-      if (rhs.isName() && exportsToNodes.containsKey(symbolToExport)) {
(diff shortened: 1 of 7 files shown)


rkirov:
Travis breaking because of https://github.com/angular/clutz/pull/747

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -678,6 +677,44 @@ void convertExportAssignment(
...
+      if (!child.getString().equals(child.getFirstChild().getString())) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private void moveExportStmtToADeclKeyword(
+      Node assign, Node exprNode, ExportedSymbol symbolToExport) {
+    // Rewrite the AST to export the symbol directly using information from the export
```
bowenni: Should this be a comment for the function, instead of an implementation comment?

bowenni: `exprNode` is unused?

PR #749, opened 2018-05-22
npm audit suggested fix

domesticmouse:
Node package manager found a vulnerability and suggested a `npm audit` which lead to the following suggested fix.

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -11,18 +11,16 @@
       "dev": true
     },
     "balanced-match": {
-      "version": "0.4.2",
-      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-0.4.2.tgz",
-      "integrity": "sha1-yz8+PHMtwPAe5wtAPzAuYddwmDg=",
-      "dev": true
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/balanced-match/-/balanced-match-1.0.0.tgz",
+      "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c="
     },
     "brace-expansion": {
-      "version": "1.1.6",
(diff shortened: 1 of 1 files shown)


npm-shrinkwrap.json
```diff
@@ -11,18 +11,16 @@
...
-      "version": "1.1.6",
-      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.6.tgz",
-      "integrity": "sha1-cZfX6qm4fmSDkOph/GbIRCdCDfk=",
-      "dev": true,
+      "version": "1.1.11",
+      "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
+      "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
       "requires": {
-        "balanced-match": "0.4.2",
+        "balanced-match": "^1.0.0",
```
mprobst: I think we want precise matches here. With `^` more likely than not our build might break because it pulls in some upgraded version that's not actually compatible.

Is it possible to either run the tool such that it doesn't produce floating deps, or just replace all `"^` with `"`?

domesticmouse: I removed the `^` versions and then reverted the package versions that regressed. This minimized diff now passes `npm audit`.

domesticmouse:
I have no idea how this change is causing the travis failures. Halp?

mprobst:
@domesticmouse halping, or at least trying to. The symptom (empty strings in tests) is typically what happens if we cannot find clang-format for some reason. I'm patching your change and seeing what's going on.

domesticmouse:
SGTM, thanks for digging into this!

For some reason, this whole structure reminds me of my mis-spent youth playing with [Mouse Trap](https://en.wikipedia.org/wiki/Mouse_Trap_(game)).

PR #751, opened 2018-06-05
Use both 'this.' and 'window.' for normalization of repeated global symbols

rkirov:
For some reason, I cannot reproduce the issue in our test harness, but
for internal Google users, when compiling over the standard typings
clutz emits:

declare namespace o_O.clutz.this {
...
}

This is invalid TS syntactically.

Previously we fixed a similar issue with
https://github.com/angular/clutz/commit/5df4937a262580fff8cfac05729027d8b157a9d1.
but we missed this instance.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -802,12 +802,11 @@ class DeclarationGenerator {
       // Some extern symbols appear twice, once unprefixed, and once prefixed with window or this.
       // Skip the second one, if both exist.
       TypedVar symbol = it.next();
-      if (symbol.getName().startsWith("window.") || symbol.getName().startsWith("this.")) {
-        String normalizedName = symbol.getName().replaceAll("^(window|this)\\.", "");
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -948,6 +945,18 @@ private boolean shouldAvoidGeneratingExterns(String filePath, String symbolName)
     return false;
   }
 
+  /**
+   * Strip window. or this. from full symbol names. Technically, window. and this. symbols are
+   * tracked separately in the type system from the global symbols, but they largely overlap both in
+   * TS and Closure.
+   */
+  private String normalizeWindowGlobals(String name) {
+    if (name.startsWith("window.") || name.startsWith("this.")) {
```
evmar: The 'if' check isn't needed here, right?  replaceAll will only match if these are true?

rkirov: right, of course. Bad refactoring.

rkirov:
Indeed, it does!

PR #752, opened 2018-06-06
Fix gents not to throw when visiting a typedef alias.

rkirov:
Paterns of the shape:

class C {}
/** @typedef {...} */
C.Alias = Other;

where throwing off gents, because we were trying emit twice: once for
Alias and once for Other. Fix by not considering Other in such cases.

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -120,12 +120,23 @@ public final class TypeConversionPass implements CompilerPass {
           // Typedef of simple types
           createTypeAlias(n, parent);
           break;
-        case NAME:
         case VAR:
         case LET:
         case CONST:
           createTypeAlias(n, parent);
           break;
+        case NAME:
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -120,12 +120,23 @@ public void visit(NodeTraversal t, Node n, Node parent) {
...
           createTypeAlias(n, parent);
           break;
+        case NAME:
+          // NAME token can occur in many locations. Only create an alias for ones that are direct
+          // children of statements.
+          // Without this check, gents will try to create two aliases for code like:
+          // /** @typedef {...} */
+          // Foo.Bar = Buz;
+          // Because of Bar and Buz.
+          if (parent.getToken() == Token.EXPR_RESULT && parent.getChildCount() == 1) {
```
bowenni: nit: Use `parent.isExprResult()` instead?

rkirov: done

PR #753, opened 2018-06-07
Accept --dependenciesManifest and --sourcesManifest

bowenni:
This allows the node wrapper to pass manifest files instead of all files to gents, which causes E2BIG failures.

### DIFF
--- a/src/main/java/com/google/javascript/gents/Options.java
+++ b/src/main/java/com/google/javascript/gents/Options.java
@@ -48,6 +48,24 @@ public class Options {
   String moduleRewriteLog = null;
 
   @Option(
+    name = "--dependenciesManifest",
+    usage =
+        "the path to a manifest file containing all dependencies\n"
+            + "This option overrides the explicitly passed files",
+    metaVar = "DEPENDENCIES_MANIFEST"
+  )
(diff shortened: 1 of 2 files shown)


bowenni:
@rkirov 
Disallows the overriding behavior now. PTAL.
npm-shrinkwrap change is in https://github.com/angular/clutz/pull/754.

src/main/java/com/google/javascript/gents/Options.java
```diff
@@ -136,7 +154,37 @@ public CompilerOptions getCompilerOptions() {
...
-    srcFiles.addAll(arguments);
+
+    if (filesToConvert.size() != 0 && sourcesManifest != null) {
+      throw new CmdLineException(
+          parser, "Don't specify a sources manifest file and source files at the same time.");
+    }
+    if (arguments.size() != 0 && dependenciesManifest != null) {
+      throw new CmdLineException(
+          parser,
+          "Don't specify a dependencies manifest file and dependency files at the same time.");
```
rkirov: "dependency files as arguments"

bowenni: Done.

PR #756, opened 2018-06-12
Fix a problem when migrating multiple sources that destructuring import the same module.

bowenni:
Turns out that I never understood gents 😞 

`root` is different from the sourceFile object from TypeScript. `root` is not the root node of a single
source file. Instead all source files are direct children of this "root" node. That means when the `DestructuringCollector` stores imported modules it's mixing the modules across multiple source files. The fix is to store the imported modules, per file.


rkirov:
LGTM, also https://github.com/angular/clutz/pull/757 should fix the clutz breakage.

PR #758, opened 2018-06-15
Make Clutz translate string Enums properly.

bowenni:


rkirov:
This might be hard to land internally. Make sure to do a global presubmit.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1802,58 +1804,85 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
...
       }
 
-      boolean isNumericEnum = type.getElementsType().getPrimitiveType().equals(numberType);
-      if (isNumericEnum) {
-        emit("enum");
+      JSType primitiveType = type.getElementsType().getPrimitiveType();
+      if (maybeStringOrNumericEnum(primitiveType, unqualifiedName, node)) {
+        return;
+      }
+
```
rkirov: add a comment here '// since this is an enum of type different from string literals or numbers, we cannot emit it as TS enum. Instead we emit it as a var/type alias pair'

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1802,58 +1804,85 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
...
-          @Nullable Node n = elements.get(elem);
-          if (n != null && n.isNumber()) {
+      // The current node points to the enum type declaration, this means that the next node will
+      // be the OBJECTLIT containing all enum key and value pairs. However, globally declared
+      // enums that are indirectly provided will instead be pointing to the parent of the
+      // OBJECTLIT parent.
+      Node objectOfAllMembers = node.getNext() != null ? node.getNext() : node.getFirstChild();
+
+      // Peek an enum member. If the Closure string enum's value is not literal don't emit anything
+      // and fall back to the safe conversion.
```
rkirov: what if the n-th member is non-literal?

```
{ A: 'foo', B: 'bar', C: f() }
```
I think you need to check them all.

PR #760, opened 2018-06-21
Improve Clutz string enums translation

bowenni:
Try to emit literal type as much as possible.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1756,16 +1756,16 @@ class DeclarationGenerator {
 
     private void visitTypeAlias(
         JSType registryType, String unqualifiedName, boolean emitNeverBrand) {
-      visitTypeAliasHelper(registryType, unqualifiedName);
+      emitTypeAliasPrefix(registryType, unqualifiedName);
       // emit a brand to prevent accidental compatibility of values with an enum.
       if (emitNeverBrand) emit("&{clutzEnumBrand: never}");
       emit(";");
       emitBreak();
     }
 
-    private void visitTypeAlias(
+    private void visitTypeAliasForMixedStringEnums(
         JSType registryType, String unqualifiedName, Set<String> literalTypes) {
-      visitTypeAliasHelper(registryType, unqualifiedName);
+      emitTypeAliasPrefix(registryType, unqualifiedName);
       emit("&{clutzEnumBrand: never}");
       // Inline all literal types in the type alias.
       for (String n : literalTypes) {
@@ -1775,7 +1775,7 @@ class DeclarationGenerator {
       emitBreak();
     }
 
-    private void visitTypeAliasHelper(JSType registryType, String unqualifiedName) {
+    private void emitTypeAliasPrefix(JSType registryType, String unqualifiedName) {
       emit("type");
       emit(unqualifiedName);
       emit("=");
@@ -1834,13 +1834,15 @@ class DeclarationGenerator {
         return;
       }
 
-      // Since this is an enum of type different from string literals or numbers, we cannot emit it
-      // as TS enum. Instead we emit it as a var/type alias pair. We also inline all literal types
-      // in the type alias so this type can be compared and assigned by the literal types (because
-      // they don't have the clutzEnumBrand).
+      // We checked the types of all members in the enums above. Since it's not all string literals
+      // or numbers, we cannot emit a TypeScript enum. Instead, we emit it as a var/type alias pair.
+      // Specially for string enums, we still try to emit as many literal types as possible to make
+      // it close to the original intent from Closure. To do that we also inline all literal types
+      // in the type alias so this type alias can be compared and assigned by the literal types
+      // (because the literal types don't have clutzEnumBrand).
       if (primitiveType.equals(stringType)) {
         Set<String> literalTypes = collectAllLiterals(elements);
-        visitTypeAlias(primitiveType, unqualifiedName, literalTypes);
+        visitTypeAliasForMixedStringEnums(primitiveType, unqualifiedName, literalTypes);
       } else {
         visitTypeAlias(primitiveType, unqualifiedName, true);
       }
@@ -1853,8 +1855,9 @@ class DeclarationGenerator {
       for (String elem : sorted(type.getElements())) {
         emit(elem);
         emit(":");
-        // For string enums that have some literal values and some calculated values try to use the
-        // literal values as types as much as possible.
+        // For string enums that have some literal values and some calculated values, we  try to use
+        // the literal values as types as much as possible. For calculated values there's nothing
+        // much we can do. Just back off and use the type alias.
         Node n = elements.get(elem);
         if (primitiveType.equals(stringType) && n.isString()) {
           emit("'" + n.getString() + "'");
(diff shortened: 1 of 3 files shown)


bowenni:
This doesn't quite work because of [this](http://www.typescriptlang.org/play/#src=type%20P%20%3D%20string%26%7BclutzEnumBrand%3A%20never%7D%3B%0D%0Adeclare%20const%20foo%3A%20%7B%0D%0A%20%20A%3A%20'1'%2C%0D%0A%20%20B%3A%20P%2C%0D%0A%7D%3B%0D%0Adeclare%20function%20f(v%3A%20P)%3A%20void%3B%0D%0A%2F%2F%20fails%20because%20foo.A%20isof%20type%20literal%20'1'%0D%0A%2F%2F%20which%20is%20missing%20the%20brand%0D%0Af(foo.A)%3B).

Three ways forward:
1. Drop the brand for string enums that have both calculated values and string literals
2. Add the branch to each string literal
3. Inline all string literals in the type

bowenni:
Implemented approach 3.

(1) is type unsafe because losing the brand means [arbitrary strings can be assigned](http://www.typescriptlang.org/play/#src=type%20P%20%3D%20string%3B%0D%0Adeclare%20const%20foo%3A%20%7B%0D%0A%20%20A%3A%20'1'%2C%0D%0A%20%20B%3A%20P%2C%0D%0A%7D%3B%0D%0Alet%20x%3A%20P%20%3D%20'1'%3B%0D%0A%2F%2F%20No%20error!%0D%0Alet%20y%3A%20P%20%3D%20'2'%3B).
(2) will not work for similar reasons.

mprobst:
I introduced the brand to prevent assignability from string to enums and vice-versa.

```ts
enum A {
  B = 'b'
}

let a: A = A.B;
let astr: string = 'b';

a = astr; // error
astr = a;
```

In TS, `a` is not assignable from `astr`, and the same holds for Closure enums. With our previous emit, Closure enums ended up being just strings (or objects etc), and were assignable. That caused a ton of very funky code that just passed around strings, getting that back under control was a major effort.

As long as you don't regress on that, I'm happy with whatever :-)

bowenni:
Ack.
Eventually I didn't drop the brand so no regression on the type safety (but improvements!).
google3 global presubmit is also green.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1756,16 +1756,32 @@ private void visitTypeAlias(JSType registryType, TypedVar symbol) {
...
+      emit("&{clutzEnumBrand: never}");
+      // Inline all literal types in the type alias.
+      for (String n : literalTypes) {
+        emit("|'" + n + "'");
+      }
+      emit(";");
+      emitBreak();
+    }
+
+    private void visitTypeAliasHelper(JSType registryType, String unqualifiedName) {
```
rkirov: Helper is not a very descriptive name. Maybe 'emitTypeAliasPrefix'?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1804,14 +1820,31 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
...
+
+      JSType primitiveType = type.getEnumeratedTypeOfEnumObject();
+      if (maybeStringOrNumericEnum(primitiveType, unqualifiedName, objectOfAllMembers, elements)) {
         return;
       }
 
       // Since this is an enum of type different from string literals or numbers, we cannot emit it
-      // as TS enum. Instead we emit it as a var/type alias pair.
-      visitTypeAlias(type.getElementsType().getPrimitiveType(), unqualifiedName, true);
+      // as TS enum. Instead we emit it as a var/type alias pair. We also inline all literal types
```
rkirov: 'Specifically for string enums, we also inline ...'

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1756,16 +1756,32 @@ private void visitTypeAlias(JSType registryType, TypedVar symbol) {
...
-      visitType(registryType, true, false);
+      visitTypeAliasHelper(registryType, unqualifiedName);
       // emit a brand to prevent accidental compatibility of values with an enum.
       if (emitNeverBrand) emit("&{clutzEnumBrand: never}");
       emit(";");
       emitBreak();
     }
 
+    private void visitTypeAlias(
+        JSType registryType, String unqualifiedName, Set<String> literalTypes) {
```
rkirov: add a comment that intentionally we always emit the brand here. Also, this is really only safe for string literal types, add that to the javadoc too.

rkirov: The more I think about it, I don't think this should be an overload of 'visitTypeAlias', but rather 'visitTypeAliasForMixedStringEnums'. I don't see how the function will be useful for other type alias declarations.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1820,8 +1853,15 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
       for (String elem : sorted(type.getElements())) {
         emit(elem);
         emit(":");
-        // No need to use type.getMembersType(), this must match the type alias we just declared.
-        emit(unqualifiedName);
+        // For string enums that have some literal values and some calculated values try to use the
+        // literal values as types as much as possible.
+        Node n = elements.get(elem);
+        if (primitiveType.equals(stringType) && n.isString()) {
+          emit("'" + n.getString() + "'");
```
rkirov: does this makes sense if primitiveType != stringType, is this valid closure

/** @enum {string|number} */
let x = {A: 0, B: "fsds"}

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1831,26 +1871,32 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
       emitBreak();
     }
 
+    private Set<String> collectAllLiterals(Map<String, Node> elements) {
+      Set<String> literalTypes = new HashSet();
```
rkirov: literal initializers, not types. We emit types but when we collect them they are just initializers.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
bowenni: Reworded a bit.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
bowenni: Renamed the function and skipped the javadoc because the function name is more descriptive now.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
bowenni: This is actually a common pattern in google3 :scream: 

I don't understand your question here. `emit("'" + n.getString() + "'");` doesn't make sense if primitiveType != stringType. That's why the line above checked the type of primitiveType.

Anyways, I added a test case for this mixed enums.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
Added javadoc too.

PR #761, opened 2018-06-22
Replace deprecated FileTraversal API.

rkirov:


rkirov:
I am not even sure it is a different order, and it doesn't matter. However, there are also clutz breakages because of new externs.

PR #763, opened 2018-06-25
Emit static properties of interfaces.

mprobst:
In Closure, JS code can legally define static properties on interfaces:

    /** @interface */
    class Foo {
      static bar() { return 'a'; }
    }

This is also allowed on interfaces not using the newer class syntax, but
just prototype declarations.

In TypeScript, this can be represented as a namespace object that
matches the interface's name:

    // A normal interface definition.
    interface A {
      foo(): void;
    }

    namespace A {
      // Declare a static function in A
      export function bar() {
        return 'bar';
      }
    }

    // Can invoke bar on A.
    A.bar();

    // B can still impement A with only the `foo` method defined.
    class B implements A {
      foo() {}
    }

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1084,8 +1084,16 @@ class DeclarationGenerator {
           } else {
             emit("function");
           }
+          boolean forcePropDeclaration = false;
+          boolean isStatic = forcePropDeclaration;
+          boolean isNamespace = true;
           treeWalker.visitProperty(
-              propName, oType, false, false, true, Collections.<String>emptyList());
+              propName,
+              oType,
+              isStatic,
+              forcePropDeclaration,
+              isNamespace,
+              Collections.<String>emptyList());
         }
       }
     }
@@ -1608,6 +1616,7 @@ class DeclarationGenerator {
       //
       // Emit original name class before Instance in order to match the already emitted JSDoc.
       final boolean emitInstance = !ftype.isInterface();
+      Set<String> staticProps = getTypePropertyNamesToEmit(ftype, true);
       if (emitInstance) {
         emit("class");
         emit(name);
@@ -1620,7 +1629,19 @@ class DeclarationGenerator {
         emitBreak();
         // we pass an empty set because static function can never refer to a TemplateType defined
         // on the class
-        visitProperties(ftype, true);
+        visitStaticProperties(ftype, staticProps, /* isInNamespace*/ false);
+        unindent();
+        emit("}");
+        emitBreak();
+      } else if (!staticProps.isEmpty()) {
+        // This is an interface, but in Closure, interfaces can still have static properties
+        // defined on them. Emit those in a namespace that matches the interface's name.
+        emit("namespace");
+        emit(name);
+        emit("{");
+        indent();
+        emitBreak();
+        visitStaticProperties(ftype, staticProps, /* isInNamespace*/ true);
         unindent();
         emit("}");
         emitBreak();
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1084,8 +1084,16 @@ private void declareNamespace(
           } else {
             emit("function");
           }
+          boolean forcePropDeclaration = false;
```
mprobst: FYI no change intended, this is just to improve readability.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2303,7 +2324,45 @@ private void visitRecordType(ObjectType type) {
       emit("}");
     }
 
-    private Set<String> getSortedPropertyNamesToEmit(final ObjectType type) {
+    /**
+     * Returns the (sorted) set of properties that should be emitted within types, i.e. on classes,
+     * interfaces, or in object types.
+     *
+     * @param isStatic if true, skips properties that are likely namespaces during extern
```
mprobst: this is a bit odd, but matches previous behaviour, I only moved the code around. I guess we had a good reason for that?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2494,42 +2551,43 @@ private void addRelatedInterfaces(ObjectType instance, Set<ObjectType> interface
...
+            objType,
+            !isInNamespace,
+            false,
+            isInNamespace,
+            Collections.<String>emptyList());
+      }
     }
 
-    private void visitProperties(
+    private void visitInstanceProperties(
```
mprobst: it turns out this was only ever called directly with isStatic=false, so I refactored to split the two methods and call `visitProperty` directly. That allowed passing down the `propNames` for the static case, which in turn allows us to check conveniently beforehand if there are any properties we need to emit, so that we can avoid empty `namespace` declarations for interfaces that do not have static props. I also had to pull up some conditional checks into `getTypePropertyNamesToEmit`, I think by and large it's a bit more sensible now.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: comment on line 2337 and extern_static_namespace.externs.js example. That said, if I commented out that line all tests pass, so maybe we fixed it some other way.

Externs is significant because "namespaces' in extersn are declared without a 'goog.provide'.

src/test/java/com/google/javascript/clutz/interface.d.ts
```diff
@@ -1,4 +1,8 @@
 declare namespace ಠ_ಠ.clutz {
+  namespace interface_exp {
+    function staticMethod ( ) : number ;
```
rkirov: isn't this basically the same test as 'interface_static_method.d.ts'. Let's just keep one of the two.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: This function is getting so small, it almost feels like it should be inlined, but I guess it is nice to keeps for symmetry with visitStaticProperties  which has more meat.

mprobst: I like the symmetry, without it it's a bit unclear why this code path exists at all.

src/test/java/com/google/javascript/clutz/interface.d.ts
mprobst: done.

mprobst:
FYI I found that we double-emitted the static properties if and only if they were declared on a type that was immediately `goog.provide`d, in `walkInnerSymbols`, preceded by a comment about how this duplicates `visitProperties`. I dropped that code, so this change found some more cleanup possibilities.

PR #764, opened 2018-06-26
Emit value aliases for interfaces with static props.

mprobst:
While #763 more systematically emitted static properties of interfaces,
users in some situations could still not access them. More specifically,
if an interface with static properites was re-exported (aliased), Clutz
would only emit a `type Alias = Actual` re-export, but not a
corresponding variable `var Alias: typeof Actual`.

This change fixes the problem by emitting `var` aliases for interfaces
with static properties.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -154,6 +154,8 @@ class DeclarationGenerator {
 
   private static final String MODULE_PREFIX = "module$exports$";
 
+  private static final Splitter DOT_SPLITTER = Splitter.on('.');
+
   public static void main(String[] args) {
     Options options = null;
     try {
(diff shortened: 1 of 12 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1543,9 +1543,11 @@ private void visitTypeValueAlias(String unqualifiedName, ObjectType otype) {
       visitTemplateTypes(otype, Collections.emptyList(), false);
       emit(";");
       emitBreak();
-      if (!otype.isInterface()) {
+      if (!otype.isInterface() || !getTypePropertyNamesToEmit(otype, true).isEmpty()) {
```
rkirov: What happens if the original declaration is in a different js_library, i.e. all clutz is seeing is:

```
goog.module('alias_for_interface');

var AliasedInterface = goog.require('aliased_interface');
exports = AliasedInterface;
```

I think we already emit the 'var' declaration because isInterface is false on unknown types, but double-check and maybe comment?

mprobst: Good point, done, and your expectation is correct.

PR #765, opened 2018-06-27
Strip initial new line added by gents.

rkirov:
Not clear why and when does this happen, but for now just
unconditionally strip it, as it causes too much noise.


src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -225,6 +225,11 @@ public CodeGenerator getCodeGenerator(Format outputFormat, CodeConsumer cc) {
                 .setOutputTypes(true)
                 .build();
 
+        // For whatever reason closure always prefixes the emit with an empty new line.
+        // Technically, this will not respect if the original .js intentionally had a new line
+        // to begin with, but I would argue that's a bad style.
+        if (tsCode.startsWith("\n")) tsCode = tsCode.substring(1);
```
bowenni: nit: Remove all leading `\n`s?

I feel if we care more about avoiding the bugs caused by the leading empty newline we should remove all leading empty newlines. If we care more about respecting the original .js we should remove all newlines added by Closure.

rkirov: Done, now we remove only the ones added by closure. Take another look.

src/main/java/com/google/javascript/gents/TypeScriptGenerator.java
```diff
@@ -225,6 +225,19 @@ public CodeGenerator getCodeGenerator(Format outputFormat, CodeConsumer cc) {
                 .setOutputTypes(true)
                 .build();
 
+        // For whatever reason closure always prefixes the emit with an empty new line.
+        // Technically, this will not respect if the original .js intentionally had a new line
```
bowenni: Comments are out-of-date. We now respect the intentional newlines in the original .js files.

rkirov: duh, fixed.

PR #766, opened 2018-07-02
Handle static props using an @template this.

mprobst:
Previously, this would crash because of the popular "let me use an empty
list here" issue.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2786,7 +2786,7 @@ class DeclarationGenerator {
         }
         // Avoid re-emitting template variables defined on the class level if method is not static.
         List<String> skipTemplateParams =
-            isStatic ? Collections.<String>emptyList() : classTemplateTypeNames;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2796,6 +2796,8 @@ private void emitProperty(
         // In TypeScript one can use the reserved 'this' type, without templatization.
         // Detect the pattern here and remove the templatized type from the emit.
         if (typeOfThis != null && typeOfThis.isTemplateType()) {
+          // skipTemplateParams might be an immutable list.
+          skipTemplateParams = new ArrayList<>(skipTemplateParams);
```
rkirov: Why not just make it mutable at the beginning?

mprobst: It's being passed in from multiple paths, many of which pass an empty list. Our handling here is pretty messy I'm afraid (and arguably so is Java's collection library here). I think a good general rule of thumb is to expect incoming collections to be immutable, and create copies when you need to change them.

PR #767, opened 2018-07-04
Fix Clutz string enum: escape strings.

bowenni:


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1885,7 +1886,7 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
         // much we can do. Just back off and use the type alias.
         Node n = elements.get(elem);
         if (primitiveType.equals(stringType) && n.isString()) {
-          emit("'" + n.getString() + "'");
+          emit("'" + escapeJava(n.getString()) + "'");
```
mprobst: you're emitting JS - can you comment why `escapeJava` is the right thing to do here, or at least approximately right?

Given the function might not (or might - I don't know!) match our semantics, I think it'd also be fair to just write a very small for loop that escapes the \' characters in the string and be done with it.

Now that I think of it, Closure should have some API somewhere to emit escaped JS strings, maybe that's accessible?

src/test/java/com/google/javascript/clutz/enum.d.ts
```diff
@@ -1,3 +1,12 @@
+declare namespace ಠ_ಠ.clutz.some {
+  enum EscapedEnum {
+    A = '\\' ,
```
mprobst: are you dealing with the non-identifier enum key in a different CL?

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
bowenni: Closure has an [API to escape string](https://github.com/google/closure-compiler/blob/407cb25bcf5c4afd23737ebc1414c04f84a31829/src/com/google/debugging/sourcemap/Util.java#L32) but it's not public.
I'm using `escapeEcmaScript` instead.

src/test/java/com/google/javascript/clutz/enum.d.ts
bowenni: Yes the other fix was submitted in https://github.com/angular/clutz/pull/768

PR #768, opened 2018-07-04
Fix Clutz string enum: don't emit string enums if a key is numeric.

bowenni:


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1927,9 +1927,13 @@ private boolean maybeStringOrNumericEnum(
       }
 
       // Look at all enum members. If any of the Closure string enum's values is not literal don't
-      // emit anything and fall back to the safe conversion.
+      // emit anything and fall back to the safe conversion. Also, if any of the Closure string
+      // enum's keys starts with a digit it's invalid in TS.
```
rkirov: Expand a bit that even enum E { '1': ...} is invalid, that was surprise to me.

Added a test case for it.

PR #769, opened 2018-07-12
 Change exported alias in declared module to match unqualified name

SLaks:
TypeScript auto-suggests imports based on the name that the module exports.
Exporting the actual Closure namespace name makes auto-imports work as expected for default imports (before this, it would suggest every module for `alias`).

I split this into two commits for easier reviewing.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1286,19 +1286,21 @@ private void declareModule(
     emitNoSpace("' {");
     indent();
     emitBreak();
+    // This name is offered for auto-import.
```
mprobst: Maybe expand on this a bit? "Use the proper type name as the alias name, so that TypeScript language service offers it as an auto-import"?

SLaks: Done

PR #770, opened 2018-08-09
visit class constructors when gathering function parameters

evmar:
This is needed by an upcoming change in JSCompiler, which exposes
'class' nodes to Clutz.  The test doesn't really check it, but
hopefully it will at least exercise the code a bit once the JSCompiler
change is in.


evmar:
CC @concavelenz

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -3097,18 +3113,21 @@ private void visitFunctionParameters(
...
           emit("...");
         }
+
+        String pName = "";
         if (names != null && names.hasNext()) {
-          emitNoSpace(names.next());
-        } else {
-          String pName;
+          pName = names.next();
+        }
```
rkirov: Good time to add a comment here. Something like "When we can't pick a reasonable name from the AST, but the type object tells us there are parameters with known types, we generate names like p1, p2, ...".

Also add tests for destructuring functions/methods:
```
foo([a, b]) {...}
bar({x, y}) {...}
```

src/test/java/com/google/javascript/clutz/goog_module_class.js
```diff
@@ -1,6 +1,11 @@
 goog.module('module.Foo');
 
 class A {
+  /**
+   * @param {string} x
+   * @param {...string} rest
+   */
+  constructor(x, ...rest) {}
 }
```
rkirov: I think our test runs here are not properly testing what John is working towards internally - not transpiling to ES5 internally. The question is there a option that we can use to test that, or it is still work in progress, thus not available to upstream maven builds.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -3073,22 +3090,21 @@ private void visitFunctionParameters(
       Iterator<Node> parameters = ftype.getParameters().iterator();
       Iterator<String> names = null;
       Node functionSource = ftype.getSource();
-      if (functionSource != null) {
+      if (functionSource != null && functionSource.isClass()) {
```
rkirov: Question for John, is there or going to be a 'ClassType'? This function is odd, because we use it both for functions and classes, but maybe classes should have been handled separately altogether.

src/test/java/com/google/javascript/clutz/goog_module_class.js
rkirov: Looking at the internal change, it looks like there is no way to test the proposed change through the public api, because it requires touching - https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/TranspilationPasses.java#L111

I guess we have a bit of a deadlock, so we should just land this untested, so that john can land his change and which point this test will become the correct one.

PR #773, opened 2018-08-23
Special case AngularJS promise.then

bowenni:
Keeps Clutz emitted AngularJS promise type close to the TS AngularJS type declaration.

When the callback function of `.then()` returns a `promise<never> | T` the TS AngularJS type declaration is smart enough to prune `promise<never>` and return just `promise<T>` (see [AngularJS .d.ts on DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/374598de675780b1312912d43dff88f9fbfcf0d8/types/angular/index.d.ts#L1201)). This PR attempts to port that functionality to Clutz.

The TypeScript [built-in promise](https://github.com/Microsoft/TypeScript/blob/v2.9.2/lib/lib.es5.d.ts#L1336) doesn't have this so don't prune `never` for any `.then` functions.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2944,8 +2944,9 @@ class DeclarationGenerator {
       // fixed.
       String classTemplatizedType =
           className.equals("ಠ_ಠ.clutz.goog.Promise") ? " any" : className + " < RESULT >";
+      // The AngularJS promise type should match the TypeScript type declaration since they describe
+      // the same runtime.
       if (propName.equals("then")) {
-        // Prune promise<never> off AngularJS IPromise.then return type.
         if (className.equals("ಠ_ಠ.clutz.angular.$q.Promise")) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2945,14 +2945,28 @@ private String getSignatureForInstanceTTEFn(
...
       if (propName.equals("then")) {
-        return "then < RESULT > (opt_onFulfilled ? : ( (a : "
-            + templateVarName
-            + " ) => "
-            + classTemplatizedType
-            + " | RESULT ) | null , "
-            + "opt_onRejected ? : ( (a : any ) => any ) | null) : "
-            + classTemplatizedType
-            + " ;";
+        // Prune promise<never> off AngularJS IPromise.then return type.
```
rkirov: This comment is too specific. Describe the general aim here - match typings of angular.js.

bowenni:
I sent you a CL with global presubmit (green-ish).

Issue #771, opened 2018-08-20
NullPointerException when encountering namespaced `Object` typedef


rafaelweinstein:
foo.js:
```
const foo = {};

/**
 * @constructor
 */
foo.Bar = function() {
};

/**
 * @typedef {{
 *   baz: string,
 * }}
 */
foo.Bar.Bat;
```

processed via `./build/install/clutz/bin/gents --convert foo.js -o out`

outputs:
```
java.lang.NullPointerException
	at com.google.javascript.gents.TypeConversionPass.convertTypeAlias(TypeConversionPass.java:473)
	at com.google.javascript.gents.TypeConversionPass.process(TypeConversionPass.java:69)
	at com.google.javascript.gents.TypeScriptGenerator.generateTypeScript(TypeScriptGenerator.java:192)
	at com.google.javascript.gents.TypeScriptGenerator.generateTypeScript(TypeScriptGenerator.java:121)
	at com.google.javascript.gents.TypeScriptGenerator.main(TypeScriptGenerator.java:75)
Uncaught exception in gents, exiting.
```

rafaelweinstein:
cc @evmar

evmar:
Hm that stack trace points into a comment:
https://github.com/angular/clutz/blob/9856fbe08e3c8c59724201255868b08fa34f7859/src/main/java/com/google/javascript/gents/TypeConversionPass.java#L473

cc @bowenni does this mean anything to you?

PR #774, opened 2018-08-23
Fix a NPE issue when an inner typedef isn't in any modules.

bowenni:
Fixes https://github.com/angular/clutz/issues/771

In the `CollectModuleMetadata` pass we collect all modules in a map `fileToModule`.
When visiting inner typedef we store them in the file module so `TypeAnnotationPass` can look it up and fix all references across multiple files importing the typedef. If the inner typedef isn't in any file modules there are no references to fix. So a no-op is fine.

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeConversionPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeConversionPass.java
@@ -93,7 +93,7 @@ public final class TypeConversionPass implements CompilerPass {
           // has its members converted in TypeAnnotationPass; or (2) a type alias.
           // Most class inner @typedef meant @record in closure but they were added before @record
           // was supported. Also in TypeScript interfaces are preferred to type alias because of
-          // better error reporting and ability to extend. However simple types such as string are
+          // better error reporting and extendability. However simple types such as string are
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/javascript/gents/TypeConversionPass.java
```diff
@@ -482,8 +482,11 @@ private void convertTypeAlias() {
       String newTypeName = entry.getValue();
       String filename = typesToFilename.get(oldTypeName);
       FileModule module = fileMap.get(filename);
-      // TypeAnnotationPass will convert the global type name to the local type name using this mapping.
-      module.importedNamespacesToSymbols.put(oldTypeName, newTypeName);
+      if (module != null) {
```
rkirov: not easy to write a test?

bowenni: not hard (fork inner_types test https://github.com/angular/clutz/blob/master/src/test/java/com/google/javascript/gents/singleTests/inner_types.js and remove the first line) but I think the value is limited.

rafaelweinstein:
Thx, @bowenni

PR #777, opened 2018-08-30
Support aliases through module objects.

rkirov:
Now clutz understands the pattern:

```
goog.module('bare.named.reexport');

const Mod = goog.require('original.module');

exports.x = Mod.x;
```

and generates proper reexports. By necessity of incremental mode it
generates a type and var reexport of x, even if it is only a value or
only a type.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/partial/module_reexport_with_module_obj.js
+++ b/src/test/java/com/google/javascript/clutz/partial/module_reexport_with_module_obj.js
@@ -1,0 +1,10 @@
+goog.module('bare.named.reexport');
+
+const Mod = goog.require('original.module');
+
+exports.Class = Mod.Class;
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/javascript/clutz/partial/module_reexport_with_module_obj.js
```diff
@@ -0,0 +1,10 @@
+goog.module('bare.named.reexport');
+
+const Mod = goog.require('original.module');
+
+exports.Class = Mod.Class;
+//!! Note, that we will incorrectly generate a type reexport for fn
+//!! even though it is only a value. Since we have partial info
+//!! in clutz, we pick the conservative view that everything can be
+//!! class an emit a type and a value alias.
```
evmar: typo: an->and

PR #783, opened 2018-09-18
Support @protected and @package visibility.

mprobst:


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1226,11 +1232,15 @@ private boolean isPrototypeMethod(TypedVar other) {
     return false;
   }
 
+  /**
+   * Returns true if the given property on the given object should be emitted. That is, it is
+   * visible or a constructor.
+   */
   private boolean isEmittableProperty(ObjectType obj, String propName) {
```
rkirov: ouch, we had the name backwards :/

PR #786, opened 2018-10-31
update closure compiler version

evmar:
Last update was in Feb 2018.  This puts us on Oct 2018.

Realistically clutz is developed against tip of tree closure, but this
puts the 'release' build tests closer to reality.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,7 +62,7 @@ dependencies {
   // clutz runs against closure head inside google, so default to that, but also test against
   // the latest release for external users
   if (System.getenv().get('CI_MODE') == 'Released') {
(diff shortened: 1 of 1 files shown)


evmar:
Hm, I'm not sure if this is what we want -- I'm not clear on whether this test is showing us that nightly fails (?)

bowenni:
This makes both `Released` and `Nightly` CI modes run the pinned version.
Also, I don't understand why we allow Clutz to fail on Released version of CC but require Nightly version to pass. https://github.com/angular/clutz/blob/master/.travis.yml#L18-L19

My mental model is that Released (pinned) should be required and Nightly should be optional.

evmar:
PTAL, I rewrote it to just be a minor version update

PR #789, opened 2018-10-31
document our versioning strategy

evmar:
We only really care to test against closure at HEAD, because that is
where our main user is.

Also move some developer docs into DEVELOPER.md.


evmar:
I think I now understand our version support strategy

PR #792, opened 2018-11-06
Switch emitting properties from 'var x', to 'const x'.

rkirov:
Technically, this change is more restricive for cases where people were
modifying the exports, but I am hoping that case is rare and we can fix
those examples as they come.


rkirov:
I will run a global presubmit on this one, likely in some cases we would need to still emit `let`, but `const` should be the default.

mprobst:
What's the benefit that switching gives us here? Can you call that out?

rkirov:
My motivation is making clutz emit read more like idiomatic typings .d.ts. Nothing major, if it turns out to be too hard (too many ppl rely on mutable exports) will stay with 'var' (or maybe change to 'let').

rkirov:
Indeed, it ended up quite common for ppl to use `goog.provide('a.b.c')` as totally mutable and overwriting `a.b.c`. Changed the PR to use `let` (instead of `var`).

mprobst:
@rkirov what are the scoping rules for `let` in TS namespaces? At least for `export let`, they are conflicts, aren't they?

rkirov:
Yes, there is a conflict if you write:

```
declare namespace a.b {
  let c: number;
}
namespace a.b.c {
  let d: number;
}
```

But that is the same if you have `var`s.

PR #795, opened 2018-11-20
Remove splitting classes into C (static side) and C_instance

rkirov:
This was needed to support @constructor es5 classes that had custom
semantics in closure. These days most code is using es6 classes, so the
work-around is not needed and it greatly simplifies emitted .d.ts.


rkirov:
BTW, the commit has a lot better message now, but github does not update the PR message after force pushing the branch. You have to click through to see the commit message.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2586,7 +2525,13 @@ private void visitObjectType(
           ((ObjectType) instanceType).getOwnPropertyNames(),
           superClassFields,
           classTemplateTypeNames);
-      // Statics are handled in INSTANCE_CLASS_SUFFIX class.
+
+      // Interface statics were emitted in an sibling namespace prior to this point, because
```
mprobst: this needs more docs, it describes a situation, but not why the code below has to run.

Also, this method would benefit from breaking it up into sub steps, this is a pretty long read.

mprobst:
@rkirov your commit message should mention that we're intentionally accepting that we produce code that contains TS level errors. Your example also is missing an `@extends`, I believe.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
rkirov: Improved the comment. Will refactor later, as this PR is already on the large side.

rkirov:
"With this change some closure code can produce .d.ts files that are not valid TS." is where I say that. Isn't that clear?

mprobst:
@rkirov sorry, must have missed that. That's clear indeed.

PR #796, opened 2018-11-20
Fix for clutz not failing on circular inferred objects.

rkirov:
The cannonical example being:

```
/** @const */
let a = {};
a.b = a;
```

For now, just detect simple loops (property points to the parent), and
avoid emitting the property. Instead mark the failed to emit property
with a comment.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2381,32 +2381,51 @@ class DeclarationGenerator {
       }
     }
 
+    /**
+     * One can create recursive record types in closure using type inference. Since we emit them
+     * inline, we cannot express that without introducing aliases. To avoid infinite recursion, we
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2385,23 +2385,33 @@ private void visitRecordType(ObjectType type) {
       emit("{");
       Iterator<String> it = getEmittablePropertyNames(type).iterator();
       while (it.hasNext()) {
+
         String propName = it.next();
+        JSType propType = type.getPropertyType(propName);
+
+        if (propType == type) {
```
mprobst: Couldn't you easily have a `Set` of seen objects that acts as a stack, i.e. that you add something into as you're visiting, then remove as you leave?

rkirov: sure, done.

PR #797, opened 2018-11-24
Fix test of async_await_with_platform for latest Closure Compiler

teppeis:
from `com.google.javascript:closure-compiler:1.0-20181121.190252-418`


mprobst:
Thanks (and sorry for the delay, Rado is on vacation)

PR #799, opened 2018-11-29
Consume ES_NEXT from clutz

rictic:
This enables parsing code like async generators and for await.

All internal tests pass. See downstreaming change at cl/223155956

### DIFF
--- a/src/main/java/com/google/javascript/clutz/Options.java
+++ b/src/main/java/com/google/javascript/clutz/Options.java
@@ -186,6 +186,7 @@ public class Options {
     // Late Provides are errors by default, but they do not prevent clutz from transpiling.
     options.setWarningLevel(DiagnosticGroups.LATE_PROVIDE, CheckLevel.OFF);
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/Options.java
```diff
@@ -186,7 +186,7 @@ public CompilerOptions getCompilerOptions() {
     // Late Provides are errors by default, but they do not prevent clutz from transpiling.
     options.setWarningLevel(DiagnosticGroups.LATE_PROVIDE, CheckLevel.OFF);
 
-    options.setLanguage(LanguageMode.ECMASCRIPT_2017);
+    options.setLanguage(LanguageMode.ECMASCRIPT_NEXT);
```
mprobst: Is ES_NEXT the "floating latest", or just an alias for some version? Maybe add a comment explaining that?

rictic: It's the floating latest. Added a comment.

PR #801, opened 2018-12-03
Emit noStructuralTyping fields in subclasses.

mprobst:
Previously, Clutz would only emit a `noStructuralTyping_` field in a
class if it didn't have a superclass. That prevents structural typing
across entirely unrelated types, but it still allows structural typing
based assignments in related types extending the same base class. For example:

    class Parent { private noStructuralTyping_: {}; }
    class Child1 extends Parent { someField: number; }
    class Child2 extends Parent { someField: number; }

    let c = new Child1();
    c = new Child2();  // succeeds because Child1 matches Child2.

This is undesirable in Clutz for two reasons:

* Closure implements a nominal type system, Clutz emit for usage from
  TypeScript should match that.
* Clutz generally does not emit private fields. This means even types
  that wouldn't be structurally compatible in TypeScript, with an
  otherwise fully structural type system, appear assignment compatible.

This change fixes the problem by emitting a "structural type busting"
field in every class. The fields are named after the globally-unique
name of their containing type, to avoid collisions. There remains a
small chance of collision assuming `Foo.Bar` and `Foo_Bar` both exist,
but that's deemed acceptable - clients should ignore errors in the
generated code anyway.


mprobst:
This not quite ready for review, still lacking some test updates. Publishing it now for easier testing.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2462,11 +2469,13 @@ private void visitObjectType(
...
-      if (type.isNominalConstructor()
-          && !type.isInterface()
-          // But only for non-extending classes (TypeScript doesn't like overriding private fields)
-          && getSuperType(type) == null) {
-        emit("private noStructuralTyping_: any;");
+      if (type.isNominalConstructor() && !type.isInterface()) {
+        emit("private noStructuralTyping_");
+        // TypeScript does not allow overriding private properties from superclasses, so make sure
+        // to emit a private property name that's specific to this type.
+        String suffix = type.hasDisplayName() ? escapeForJSProperty(type.getDisplayName()) : "";
```
evmar: What does it mean if the type doesn't have a display name?  Maybe we shouldn't behave nominally in this case?

src/test/java/com/google/javascript/clutz/DeclarationGeneratorTests.java
```diff
@@ -130,7 +131,9 @@ static String getExternFileNameOrNull(String testFileName) {
             .flatMap(List::stream)
             .map(Path::toFile);
 
-    return Stream.concat(jsFiles, zipFiles).collect(Collectors.toList());
+    List<File> fileList = Stream.concat(jsFiles, zipFiles).collect(Collectors.toList());
+    Collections.sort(fileList);
+    return fileList;
```
evmar: I don't get this part, ... ?

mprobst: This was just to make test execution reproducible and make it easier to alphabetically go through the list of failures.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
mprobst: I cannot construct a test case where we emit a Clutz declaration for a type, but the type has no display name. Any type we reach to emit, we reach through the global symbol table, and anything in that table by definition has a reachable name, which Closure uses as the display name.

So this is just defensive coding, but I think it cannot happen in practice (I know, famous last words...).

mprobst:
> Do we have any way of estimating the code size (compiler perf) impact of this? I guess it's one extra member per subclass?

I was wondering about that, too, but we don't, I think. We can watch our dashboards, though I suspect we won't see this in the noise.

Arguing from first principles, indeed this will be one extra private member per subclass. Some Closure APIs are subclass heavy, e.g. protocol buffers, but on the other hand, those also come with many members per subclass already. So we might be adding <10% members or so with this change. How that relates to effective compiler performance (memory pressure on our cache, etc), I don't know. I'd actually suspect that the `is assignable to` relation is much easier to check for the compiler with nominal types, so we might be consuming more memory, but speeding up checking. But that's just speculation.

PR #802, opened 2018-12-03
Improve this type

### LINKED ISSUE #709: clutz outputs invalid type for goog.ui.Component.getEventHandler()
It should be
`getEventHandler(): ಠ_ಠ.clutz.goog.events.EventHandler<this>|null;`
instead of this (T isn't defined)
`getEventHandler(): ಠ_ಠ.clutz.goog.events.EventHandler<T>|null;`

teppeis:
fixes #709 

## New features

Additionally, these forms are supported:

```js
/**
 * @return {!GenericClass<T>}
 * @this {T}
 * @template T
 */
A.prototype.foo = function() {
  return new GenericClass(this);
};

/**
 * @return {!Array<T>}
 * @this {T}
 * @template T
 */
A.prototype.array = function() {
  return [this];
};

/**
 * @return {!Object<string, T>}
 * @this {T}
 * @template T
 */
A.prototype.object = function() {
  return {foo: this};
};

/**
 * @return {{foo: T}}
 * @this {T}
 * @template T
 */
A.prototype.record = function() {
  return {foo: this};
};


/**
 * @param {string} str
 * @this {Function}
 */
A.func = function(str) {
};
```

are converted to

```ts
  class A {
    array < T > (this : T ) : T [] ;
    foo < T > (this : T ) : GenericClass < T > ;
    object < T > (this : T ) : { [ key: string ]: T } ;
    record < T > (this : T ) : { foo : T } ;
    func (this : Function , str : string ) : void ;
  }
```

But non-templatized `this` in a function type is not supported yet.
```js
/**
 * @type {function(this: Date)}
 */
var foo = function() {};
```

## Changes

```js
/**
 * @return {THIS}
 * @this {THIS}
 * @template THIS
 */
C.prototype.foo = function() {
  return this;
};
```
is currently converted to 
```ts
  class C {
    foo ( ) : this ;
  }
```
but in this PR, it is converted to
```ts
  class C {
    foo < THIS > (this : THIS ) : THIS ;
  }
```

Because polymorphic `this` types are invalid in some places, but templatized `this` is all valid. They are equivalent. For simplicity, I convert all to templatized `this`.

```ts
declare namespace ಠ_ಠ.clutz.nsThis {
  class A {
    // all valid
    returnThis<T>(this: T): T;
    array<T>(this: T): T[];
    generics<T>(this: T): Set<T>;
    object<T>(this: T) : { [ key: string ]: T } ;
    record<T>(this: T): { foo: T };
  }
  class B {
    returnThis(): this; // valid
    generics(): Set<this>;  // valid
    array(): this[]; // valid
    object(): { [ key: string ]: this } ; // invalid!!!
    record(): { foo: this }; // invalid!!!
  }
}
```
[playground](http://www.typescriptlang.org/play/#src=declare%20namespace%20%E0%B2%A0_%E0%B2%A0.clutz.nsThis%20%7B%0D%0A%20%20class%20A%20%7B%0D%0A%20%20%20%20returnThis%3CT%3E(this%3A%20T)%3A%20T%3B%0D%0A%20%20%20%20array%3CT%3E(this%3A%20T)%3A%20T%5B%5D%3B%0D%0A%20%20%20%20generics%3CT%3E(this%3A%20T)%3A%20Set%3CT%3E%3B%0D%0A%20%20%20%20object%3CT%3E(this%3A%20T)%20%3A%20%7B%20%5B%20key%3A%20string%20%5D%3A%20T%20%7D%20%3B%0D%0A%20%20%20%20record%3CT%3E(this%3A%20T)%3A%20%7B%20foo%3A%20T%20%7D%3B%0D%0A%20%20%7D%0D%0A%20%20class%20B%20%7B%0D%0A%20%20%20%20returnThis()%3A%20this%3B%0D%0A%20%20%20%20generics()%3A%20Set%3Cthis%3E%3B%0D%0A%20%20%20%20array()%3A%20this%5B%5D%3B%0D%0A%20%20%20%20object()%3A%20%7B%20%5B%20key%3A%20string%20%5D%3A%20this%20%7D%20%3B%0D%0A%20%20%20%20record()%3A%20%7B%20foo%3A%20this%20%7D%3B%0D%0A%20%20%7D%0D%0A%7D)

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -3140,16 +3140,18 @@ class DeclarationGenerator {
       final JSType typeOfThis = ftype.getTypeOfThis();
       // Don't emit for a constructor like `function(new: T)`.
       // A `this` parameter in a constructor is not allowed in TypeScript.
-      if (typeOfThis != null && !ftype.isConstructor()) {
-        final JSDocInfo jsDocInfo = ftype.getJSDocInfo();
-        // Emit for templatized this like `function(this: T)` or JSDoc `@this` type.
-        if (typeOfThis.isTemplateType() || (jsDocInfo != null && jsDocInfo.getThisType() != null)) {
-          emitNoSpace("this :");
-          visitType(typeOfThis);
-          if (parameters.hasNext()) {
-            emit(", ");
-          }
-        }
+      if (typeOfThis == null || ftype.isConstructor()) {
+        return;
+      }
+      final JSDocInfo jsDocInfo = ftype.getJSDocInfo();
+      // Emit for templatized this param like `function(this: T)` or JSDoc `@this` type.
+      if (!typeOfThis.isTemplateType() && (jsDocInfo == null || jsDocInfo.getThisType() == null)) {
+        return;
+      }
+      emitNoSpace("this :");
+      visitType(typeOfThis);
+      if (parameters.hasNext()) {
+        emit(", ");
       }
     }
 
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -3134,7 +3134,7 @@ class DeclarationGenerator {
     /**
      * Emit a this parameter like `func(this: Foo)` in a function parameters.
      *
-     * TODO: emit for non-templatized this like `function(this: HTMLElement)`
+     * <p>TODO: emit for non-templatized this like `function(this: HTMLElement)`
      */
     private void emitThisParameter(FunctionType ftype, Iterator<Node> parameters) {
       final JSType typeOfThis = ftype.getTypeOfThis();
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -3006,6 +3006,7 @@ class DeclarationGenerator {
     private void visitFunctionDeclaration(FunctionType ftype, List<String> skipTemplateParams) {
       visitFunctionParameters(ftype, true, skipTemplateParams);
       JSType type = ftype.getReturnType();
+      final JSType typeOfThis = ftype.getTypeOfThis();
 
       if (type == null) return;
       emit(":");
@@ -3015,6 +3016,9 @@ class DeclarationGenerator {
       // "undefined" types are not the same.
       if (type.isVoidType()) {
         emit("void");
+      } else if (typeOfThis != null && typeOfThis.isTemplateType() && typeOfThis.equals(type)) {
+        // Special case: prefer polymorphic `this` type to templatized `this` param
+        emit("this");
       } else {
         visitType(type);
       }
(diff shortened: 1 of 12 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -3147,6 +3131,28 @@ private void visitFunctionParameters(
...
+    /**
+     * Emit a this parameter like `func(this: Foo)` in a function parameters.
+     *
+     * TODO: emit for non-templatized this like `function(this: HTMLElement)`
+     */
+    private void emitThisParameter(FunctionType ftype, Iterator<Node> parameters) {
+      final JSType typeOfThis = ftype.getTypeOfThis();
+      // Don't emit for a constructor like `function(new: T)`.
+      // A `this` parameter in a constructor is not allowed in TypeScript.
+      if (typeOfThis != null && !ftype.isConstructor()) {
```
mprobst: prefer early exits: invert the condition and return

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -3147,6 +3131,28 @@ private void visitFunctionParameters(
...
+        final JSDocInfo jsDocInfo = ftype.getJSDocInfo();
+        // Emit for templatized this like `function(this: T)` or JSDoc `@this` type.
+        if (typeOfThis.isTemplateType() || (jsDocInfo != null && jsDocInfo.getThisType() != null)) {
```
mprobst: same here, return early

src/test/java/com/google/javascript/clutz/this_type_param.d.ts
```diff
@@ -2,7 +2,8 @@ declare namespace ಠ_ಠ.clutz.nsThis {
   class C {
     private noStructuralTyping_: any;
     bar ( ) : ಠ_ಠ.clutz.nsThis.C ;
-    foo ( ) : this ;
+    foo < THIS > (this : THIS ) : THIS ;
```
mprobst: I think it might be slightly preferable to emit an actual `this` type here.

In TypeScript, if you extend this class, it'll be rather cumbersome to create the overridden method - you'll need to deal with all the template args etc instead of just `foo(): this { ... }`. That syntax is also a bit more readable to TS users.

That said, if that's hard to implement or has other corner cases, I think it's OK to go with this - I expect few people will need to extend `<THIS>` templated base classes written in Closure from TypeScript.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
teppeis: fixed

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
teppeis: fixed

src/test/java/com/google/javascript/clutz/this_type_param.d.ts
teppeis: @mprobst I think so. I've tried to handle special simple case. https://github.com/angular/clutz/pull/802/commits/89f4f2b99cbd1cf021e9cd0949cac51b7eab5e47

teppeis:
@mprobst thanks!

PR #805, opened 2018-12-05
Fix NamedNodeMap

teppeis:
`NamedNodeMap` is templatized in Closure (historical issue of DOM) but not in TypeScript like `NodeList` or `MessageEvent`. 

- https://github.com/google/closure-compiler/blob/50713dceb16b369f8f317e5b49e8a9dec2b47496/externs/browser/w3c_dom1.js#L463-L471
- https://github.com/Microsoft/TypeScript/blob/ee25cdecbca49b2b5a290ecd65224f425b1d6a9c/lib/lib.dom.iterable.d.ts#L127


teppeis:
rebased to latest master

PR #810, opened 2018-12-07
Map AsyncIteratorIterable to AsyncIterableIterator for latest Closure Compiler

teppeis:
Currently [clutz CI is failed](https://travis-ci.org/angular/clutz/jobs/465108148) because latest compiler https://github.com/google/closure-compiler/commit/91c0ace5eeeae92912e12bd84617eaf60ea7f536 introduced `AsyncIteratorIterable`.
clutz should map it to [AsyncIterableIterator](https://github.com/Microsoft/TypeScript/blob/71286e3d49c10e0e99faac360a6bbd40f12db7b6/src/lib/esnext.asynciterable.d.ts) like `IteratorIterable`.


rkirov:
Thanks for the fix! Running tests against Closure HEAD is odd, but better matches our Google internal expectations.

PR #811, opened 2018-12-12
Alias colliding names in Closure Library like goog.ui.Component

teppeis:
Closure Library causes has 18 errors of TS2417 like the following. 

```
TS2417: Class static side 'typeof Container' incorrectly extends base class static side 'typeof Component'.
  Types of property 'EventType' are incompatible.
    Type 'typeof EventType' is missing the following properties from type 'typeof EventType': ACTION, ACTIVATE, BEFORE_SHOW, BLUR, and 17 more.
```
https://github.com/teppeis/closure-library.d.ts/blob/master/errors.txt

In ES (TypeScript) classes, static props of classes are inherited to subclasses.
But in Closure, props of constructor function are not inherited.
For example, `goog.ui.Component.EventType` is a prop of `goog.ui.Component`, but it's goog.provide'd by itself and not inherited to the subclasses.

This PR aliases classes that has such props and causes the errors.
By this change, the props are only imported directory, but not available as static props of classes.

```ts
// OK
import EventType from 'goog:goog.ui.Component.EventType';
console.log(EventType);

// NG
import Component from 'goog:goog.ui.Component';
console.log(Component.EventType);
```


teppeis:
Rebased on latest master

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -241,10 +241,37 @@ public static void main(String[] args) {
 
   /**
    * If one file defines a name and another uses it as a namespace, we have the
-   * Constants.COLLDING_PROVIDE_ALIAS_POSTFIX workaround. In partial mode, Clutz can't see all
+   * Constants.COLLIDING_PROVIDE_ALIAS_POSTFIX workaround. In partial mode, Clutz can't see all
    * definitions of a name, so the list of names that require aliases must be passed as an input.
+   * Also Closure Library uses some names as both classes and namespases. For example,
+   * `goog.ui.Component` is not only a class but also a namespace in `goog.ui.Component.EventType`.
```
rkirov: Can you add a link to the github repo containing goog.ui, so when one reads the comment they can quickly click and see the example at the source.

teppeis: added the link

PR #814, opened 2018-12-18
Adds a new typedef TexImageSource to platform goldens.

rkirov:


rkirov:
Good point added the new type to the list of names to ignore.

PR #817, opened 2018-12-19
Add FullscreenOptions to CLOSURE_EXTERNS_NOT_USED_IN_TYPESCRIPT list

teppeis:
For https://github.com/google/closure-compiler/commit/aed8df734fac7fed10e5f7c9f6582a659b7b1a63

But this test is too flaky...


rkirov:
True. These tests change too much. Maybe we should move the tests that need platform to partial/  and not pass externs.

PR #825, opened 2019-01-09
Add support for emitting iterated inner classes.

rkirov:
Clutz does a separate pass to emit inner properties of a Closure class
(like other classes, interfaces, enums or typedefs) that in TS are not
expressable as statics. The newly found properties are emitted in a
namespace.

This works because TS allows to have a namespace
and a class that share a name (despite both being values).

Previously, this process only worked at a single depth, but with this
change we iterate at aribrary depths for classes and interface, i.e.
inner classes that have their own inner classes.

For now we only interate on classes and interfaces, but technically we
can extend this to enums and typedefs if need arises.


rkirov:
Testing making a PR from angular/clutz branch instead of rkirov/clutz branch, because Travis seems stalled on the other ones.

Issue #824, opened 2019-01-09
gents whitespace regression


rkirov:
Currently, gents is broken at head due to new removal of whitespace in almost all golden files. Likely this is due to the an underlying Closure change.

As a first order of business, we should fix the build and then investigate how to get back the proper whitespace.

teppeis:
@rkirov Test results of snapshots:

- Pass: `com.google.javascript:closure-compiler:1.0-20190107.193122-442`
- Fail: `com.google.javascript:closure-compiler:1.0-20190108.212204-443`

This commit? https://github.com/google/closure-compiler/commit/fdd9b92d01865eaaac85bf6d01b0659a0a624c58

rkirov:
Yep, it's definitely that. Not clear if we can fix forward on our side. @bowenni and I will look through the whitespace reinsertion hack we do in gents and see if we can recover the old behavior. Let us know if you have any insights into this too.

PR #827, opened 2019-01-10
Work around https://github.com/google/closure-compiler/commit/fdd9b92…

bowenni:
…d01865eaaac85bf6d01b0659a0a624c58

Fixes https://github.com/angular/clutz/issues/824
Also run google-java-format on the changed files.


src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -32,8 +32,10 @@ protected void add(Node n, Context ctx) {
 
     String comment = nodeComments.getComment(n);
     if (comment != null) {
-      add(comment);
-      add("\n");
+      // CodeGernator.add("\n") doesn't append anything. Fixing the actual bug in Closure Compiler
+      // is difficult. Works around the bug by passing " \n".
```
rkirov: Add that the spacebar doesn't make it into the final emit, because it is stripped by Closure.

src/main/java/com/google/javascript/gents/GentsCodeGenerator.java
```diff
@@ -80,7 +82,9 @@ private void maybeAddNewline(Node n) {
             || (n.getParent() != null && isPreviousEmptyAndHasComment(n.getParent()));
 
     if (!hasComment && TOKENS_TO_ADD_NEWLINES_BEFORE.contains(n.getToken())) {
-      add("\n");
+      // CodeGernator.add("\n") doesn't append anything. Fixing the actual bug in Closure Compiler
+      // is difficult. Works around the bug by passing " \n".
```
rkirov: same here.

PR #831, opened 2019-01-14
Emit each inner class once

teppeis:
After https://github.com/angular/clutz/pull/822, `goog.module.ModuleLoader.EvaluateCodeEvent` and some classes are emitted twice throwing `TS2300: Duplicate identifier`.
Because `goog.module.ModuleLoader.EvaluateCodeEvent` is emitted as an inner class of not only `goog.module.ModuleLoader` but also `goog.module`.


rkirov:
Looks good. Only caveat is that we like to have tests represent the situation that this is fixing, rather than the closure lib code that was triggering the error. That way the tests are more mainainable.

In order to expedite landing the fix, I fixed your tests and created a new PR - https://github.com/angular/clutz/pull/834. I tried to preserve your ownership of the commit.

teppeis:
@rkirov thanks! 👍

PR #841, opened 2019-01-19
Improve Promise and remove `--noStrictGenericChecks` from Declaration…

rkirov:
…SyntaxTest


rkirov:
We still have `:  any` at the end of the .then :) But one step at a time

PR #844, opened 2019-01-22
Update README.md

devversion:


mprobst:
Can you make the commit message say not *what* you are doing, but *why* you are doing it? I can tell that you changed this file from the commit itself, but the interesting question 1 year later is usually why we did something :-)

devversion:
Hahah good point 😁😁 I was actually just testing if CircleCI works after Rado turning on builds for forked PRs.

Forgot closing the PR, but we can keep it if you guys like the changes :)

PR #847, opened 2019-01-30
Adds TrustedHTML and friends to the extern goldens.

rkirov:


teppeis:
@evmar this update https://github.com/google/closure-compiler/commit/0aee97d3f0d0b5da5531a0f300c75498cb79a3bb

vrana:
I've already added this in the Google repo, we just need to export it to GitHub.

evmar:
Right, what I'm saying is that this change should have some change somewhere around here
https://github.com/angular/clutz/blob/master/build.gradle#L70

PR #852, opened 2019-02-28
Add Trusted Types symbols

vrana:


vrana:
As suggested by @rkirov.

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -1458,6 +1458,13 @@
           "TransitionEventInit",
           "Transport",
           "TreeWalker",
+          "TrustedHTML",
```
rkirov: Add some comments around these as they don't trully match the original comment at the top.

vrana: Done.

koto:
Is it possible to rerun the CircleCI tests and merge the PR? The failures don't seem related.

src/main/java/com/google/javascript/clutz/PlatformSymbols.java
```diff
@@ -1458,13 +1458,16 @@
           "TransitionEventInit",
           "Transport",
           "TreeWalker",
+
+          // https://github.com/WICG/trusted-types
           "TrustedHTML",
```
rkirov: This comment is insufficient for maintainers of clutz to know what's going on. Through conversation with 	
@koto I got a bit more of a context. Please add something of this sort:

```
The TrustedTypes (TT) browser api is experimental, but there is an implementation in Chrome. Closure's platform externs will be made to unconditionally depend on TT types, and similarly all TS compilations will depend on https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/trusted-types/index.d.ts

In that sense, TT can be seen by clutz as a platform API (albeit experimental). By adding it to this list we guarantee that the symbols are emitted without the clutz prefix and transparently mapped between the hand-written externs and .d.ts.

For more info see https://github.com/WICG/trusted-types
```

rkirov:
Rebased and merged with https://github.com/angular/clutz/pull/855

PR #858, opened 2019-03-05
Added instruction to readme

mariakhomenko:
This adds a comment to README on where to look for ./gradelew

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ use it with:
 ## Example usage
 
 We don't offer a binary distribution, so first you need to build from source.
-From top level run:
+From the root directory run:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,7 +26,8 @@ use it with:
 
 ## Example usage
 
-We don't offer a binary distribution, so first you need to build from source:
+We don't offer a binary distribution, so first you need to build from source.
+From top level run:
```
rkirov: "top level" is a bit ambiguous, how about "from the root directory"

mariakhomenko: Made the change you suggested.

PR #860, opened 2019-03-07
 Removes @enum {string} and @enum {number} JsDoc.

mariakhomenko:
Removes @enum {string} and @enum {number} JsDoc. ￼…
@enum {string|number} without any additional doc doesn't add much value
in TypeScript.
If additional documentation is present, then keep @enum <explanation>
without the type.

For enums that are not strings or numbers, keep the type in JsDoc.

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -193,8 +193,7 @@ public final class CommentLinkingPass implements CompilerPass {
 
         for (Pattern p : JSDOC_REPLACEMENTS_NO_KEEP) {
           Matcher m = p.matcher(comment);
-          boolean hasAMatch = m.find();
-          if (hasAMatch) {
+          if (m.find()) {
             if (m.group("eol") != null && m.group("eol").trim().length() == 0) {
               // if the end of the line was matched, then there's nothing to keep, remove the line
               comment = m.replaceAll("");
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -28,8 +28,8 @@ public final class CommentLinkingPass implements CompilerPass {
   private static final String EOL = "(?<eol>[ \t]*\n)?";
 
   /**
-   * These RegExes delete everything except for the `block` (see BEGIN_JSDOC_LINE)
-   * and `keep` capture groups. These RegExes must contain a `keep` group.
+   * These RegExes delete everything except for the `block` (see BEGIN_JSDOC_LINE) and `keep`
+   * capture groups. These RegExes must contain a `keep` group.
    */
   private static final Pattern[] JSDOC_REPLACEMENTS_WITH_KEEP = {
     // Removes @param and @return if there is no description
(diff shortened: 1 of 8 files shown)


bowenni:
Closure enums support arbitrary types while TS only supports string and number. Can you please add a test case for that. This is to make sure that we don't remove the `@enum` JSDoc when the type isn't converted to TS.

mariakhomenko:
There is already test in enums.js singleTest that covers this case.

bowenni:
Ah I see, https://github.com/angular/clutz/blob/5cb77a5f48cca826c62382eaef6569d0e811c805/src/test/java/com/google/javascript/gents/singleTests/enum.js#L43.
LGTM.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -28,26 +28,32 @@
   private static final String EOL = "([ \t]*\n)?";
 
   /**
-   * These jsdocs delete everything except for the `keep` capture group Some regexes contain an
+   * These jsdocs delete everything except for the `keep` capture group. Some regexes contain an
    * empty capture group for uniform handling.
    */
   private static final Pattern[] JSDOC_REPLACEMENTS = {
+    // The <?keep> capture group is intentionally empty because it's not needed.
```
rkirov: I still don't understand why it exists at all, i.e. what would go wrong if I remove it completely? "not needed" sounds like it should be removed

It is just so that the implementation can look for the same named capturing group over any of the regexp? There is no way to handle that with with a null check or something?

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -28,26 +28,32 @@
...
     Pattern.compile(BEGIN_JSDOC_LINE + "@(constructor|interface|record)[ \t]*(?<keep>)" + EOL),
+    // The <?keep> capture group is intentionally empty because it's not needed.
     Pattern.compile(
         BEGIN_JSDOC_LINE
             + "@(private|protected|public|package|const)[ \t]*(\\{.*\\})?[ \t]*(?<keep>)"
             + EOL),
+    // Removes @enum {number} and @enum {string} without a description.
+    Pattern.compile(BEGIN_JSDOC_LINE + "@enum[ \t]*\\{(string|number)\\}[ \t]*(?<keep>\\*\\/|\n)"),
+    // Removes {number} or {string} type from @enum
```
rkirov: I think for this one we should not keep @enum, because tsickle rejects them.

https://github.com/angular/tsickle/blob/693dd61d59cae0ec48e4aa33734afaaa2ed63767/src/jsdoc.ts#L148

Our rules are a bit ad-hoc, so maybe you can make a case for changing it in tsickle, in any case we should be consistent.

mariakhomenko: Changed to always remove @enum.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
mariakhomenko: There isn't a good way to check if a capturing group exists (the two ways I've seen is catching the IllegalArgumentException or using reflection on a non-public Matcher method). I refactored a bit to get rid of <keep>.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -176,17 +179,38 @@ private void linkCommentBufferToNode(Node n) {
...
-      Pattern[] replacements = (type == Type.JSDOC) ? JSDOC_REPLACEMENTS : COMMENT_REPLACEMENTS;
-      for (Pattern p : replacements) {
-        Matcher m = p.matcher(comment);
-        if (m.find() && m.group("keep") != null && m.group("keep").trim().length() > 0) {
-          // keep documentation, if any
-          comment = m.replaceAll("${block}${keep}");
-        } else {
-          // nothing to keep, remove the line
-          comment = m.replaceAll("");
+      if (type == Type.JSDOC) {
```
rkirov: +1, I rather have more verbose but straight forward Java code, than unnecessarily complex regexps.

src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -176,17 +179,38 @@ private void linkCommentBufferToNode(Node n) {
...
+          } else {
+            // nothing to keep, remove the line
+            comment = m.replaceAll("");
+          }
+        }
+
+        for (Pattern p : JSDOC_REPLACEMENTS_NO_KEEP) {
+          Matcher m = p.matcher(comment);
+          boolean hasAMatch = m.find();
+          if (hasAMatch) {
```
rkirov: probably can inline hasAMatch since it is used only once.

mariakhomenko: Done

PR #866, opened 2019-03-12
Fix clutz emit for inner classes inside named exports.

### LINKED ISSUE #865: inner classes not emitted when outer class is emitted with named export
For example:
```js
goog.module('innerclass.namedexport');

class A {}

A.B = class {};

exports = {A};
```

clutz doesn't emit the `B` class. Note that if `exports = A` this works as expected.

rkirov:
Somehow despite all existing machinery we missed this case. Fix simply
extends the existing walkInnerSymbols for non-default exports.

Fixes #865


rkirov:
Not happy with the general state of clutz code around this, but I am secretly hoping that we can stop/minimize supporting goog.provides on day (in an year or two).

PR #867, opened 2019-03-14
Handle goog.requireType identically to goog.require.

rkirov:
goog.requireType is a new addition to the Closure module system.


src/test/java/com/google/javascript/clutz/partial/requireType.js
```diff
@@ -0,0 +1,8 @@
+goog.module('some.module');
+
+const AType = goog.requireType('some.other.module');
```
mprobst: nit: test case might be more readable if you'd have `AType` on the right hand side in the module name as well.

rkirov: done

PR #868, opened 2019-03-15
Remove @suppress {extraRequire} JsDoc comments.

mariakhomenko:
These are no longer useful with import statements.

### DIFF
--- a/src/test/java/com/google/javascript/gents/multiTests/import_rewrite/import.js
+++ b/src/test/java/com/google/javascript/gents/multiTests/import_rewrite/import.js
@@ -5,6 +5,10 @@ var D = goog.require("both.D");
 goog.require("nested.E");
 goog.require("nested.E.F");
 var stuff = goog.require("nested.E.F.Z");
+/** @suppress {extraRequire} My comment */
+goog.require("unused.W");
+/** @suppress  {extraRequire}  */
+goog.require("unused.V");
 
 value.B();
 var num = obj.C.x + obj.C.y;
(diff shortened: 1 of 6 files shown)


src/test/java/com/google/javascript/gents/multiTests/import_rewrite/import.js
```diff
@@ -5,6 +5,10 @@ var D = goog.require("both.D");
 goog.require("nested.E");
 goog.require("nested.E.F");
 var stuff = goog.require("nested.E.F.Z");
+/** @suppress {extraRequire} My comment */
+goog.require("unused.W");
+/** @suppress  {extraRequire}  */
```
rkirov: Odd spaces here, is that intentional?

mariakhomenko: Yes, making sure that space correctly gets matched by regex.

rkirov: Add a comment for the comment :) Otherwise, someone might helpfully fix this later.

mariakhomenko: Ok, I took a look at google3 and there are 0 instances of multiple whitespaces there, so I think I'll just make this normal whitespace.

rkirov: sounds good. I am merging this PR.

PR #875, opened 2019-05-06
Add example of union of partially visible interface collapsing.

rkirov:
In case of interfaces that extend another one beyond the scope of
visibility of partial clutz, it appears that unions with |null or
|undefined are collapsed by Closure.

This adds a test exemplifying the failure.

### DIFF
--- a/src/test/java/com/google/javascript/clutz/partial/missing_base_union.d.ts
+++ b/src/test/java/com/google/javascript/clutz/partial/missing_base_union.d.ts
@@ -1,0 +1,19 @@
+declare namespace ಠ_ಠ.clutz.module$exports$missing$base$union {
+  //!! The double extends here is a bug too.
+  interface Derived extends ಠ_ಠ.clutz.module$exports$some$base.Base extends ಠ_ಠ.clutz.module$exports$some$base.Base {
+    someField ? : string ;
+  }
+  /**
+   * This is currently broken and we are waiting on a fix in Closure, since it
+   * cannot be simply fixed in Clutz.
+   *
+   * Right now, only a is emitted correctly, because it is not an union with
+   * null or undefined, the rest are collapsed into null or undefined, without a
+   * mention of Derived.
+   */
+  function fn (a : ಠ_ಠ.clutz.module$exports$missing$base$union.Derived , b : null , c ? : undefined ) : void ;
+}
+declare module 'goog:missing.base.union' {
+  import union = ಠ_ಠ.clutz.module$exports$missing$base$union;
+  export = union;
+}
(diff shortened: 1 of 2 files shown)


bowenni:
I'm not following. Will you sending another PR to handle this case and fix the example (not collapsing the type)? Is this an example of the current status and an example of the expected behavior?

rkirov:
This happens at the Closure level, so I can't fix it in clutz. The type we receive has no mention of the original Derived.

This is an example of broken behavior, that I want to use for filing a b/ report.

src/test/java/com/google/javascript/clutz/partial/destruct_ctor.d.ts
```diff
@@ -0,0 +1,11 @@
+declare namespace ಠ_ಠ.clutz.module$exports$derived$interface$union$undefined {
+  interface Derived extends ಠ_ಠ.clutz.module$exports$some$base.Base extends ಠ_ಠ.clutz.module$exports$some$base.Base {
```
rkirov: also, why do we emit two extends?

rkirov:
I will fix master first and then clean up the commit message to make this more clear.

src/test/java/com/google/javascript/clutz/partial/destruct_ctor.d.ts
```diff
@@ -0,0 +1,11 @@
...
+    someField ? : string ;
+  }
+  //!! Only a is generated correctly, b and c should be unioned with Derived.
+  function fn (a : ಠ_ಠ.clutz.module$exports$derived$interface$union$undefined.Derived , b : null , c ? : undefined ) : void ;
+}
+declare module 'goog:derived.interface.union.undefined' {
+  import undefined = ಠ_ಠ.clutz.module$exports$derived$interface$union$undefined;
```
bowenni: Can you choose a different name that doesn't conflict with built-in global identifier 'undefined'?

bowenni:
I see. Please add comments to the tests and state that this is not the desired state, but due to limitations of Closure this is the best we can get.

src/test/java/com/google/javascript/clutz/partial/destruct_ctor.d.ts
rkirov: done, changed also file name.

src/test/java/com/google/javascript/clutz/partial/destruct_ctor.d.ts
rkirov: added a comment for that.

PR #877, opened 2019-05-07
Remove no longer relevant comment.

rkirov:
In the past closure had getAlternates and
getAlternatesWithoutStructuralTyping. Since
https://github.com/angular/clutz/commit/733783dc299805a963272d4d77bf11b12668fad4
there is only a single method, so this comment is not relevant.


rkirov:
That was because it used to say why we chose `getAlternatesWithout...` instead of `getAlternates`, but the search and replace made the comment useless.

PR #878, opened 2019-05-07
Support `@abstract` in Clutz.

bowenni:
TGP looks good (after some cleanups).

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -1707,6 +1707,9 @@ class DeclarationGenerator {
         emitBreak();
       }
       if (ftype.isConstructor()) {
+        if (ftype.isAbstract()) {
+          emit("abstract");
+        }
         // "proper" class constructor
         emit("class");
       } else if (ftype.isInterface()) {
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/clutz/abstract_class.js
```diff
@@ -0,0 +1,15 @@
...
+ * @abstract
+ */
+class AbstractClass {
+  /**
+   * @return {string}
+   * @protected
+   * @abstract
+   */
+  myAbstractMethod() {}
+}
```
rkirov: add a test for the static abstract that you have code for.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2853,14 +2859,20 @@ private void emitProperty(
...
+        boolean isAbstract,
         boolean forcePropDeclaration,
         boolean isNamespace,
         List<String> classTemplateTypeNames) {
       if (handleSpecialTTEFunctions(propertyType, propName, isStatic, classTemplateTypeNames))
         return;
 
       if (isProtected) emit("protected");
-      if (isStatic) emit("static");
+      // "static abstract" is illegal TypeScript
```
rkirov: add that we just drop abstract in that case

bowenni:
Done. PTAL.

PR #879, opened 2019-05-07
Sync an internal change.

bowenni:
Original changelist description:

```
Instead of calling Subject.actual(), store the actual value in a field, and read that.
actual() is being removed.
```


bowenni:
Added the original CL description.

PR #880, opened 2019-05-08
Handle optional parameters with a default value of undefined.

### LINKED ISSUE #629: Gents translates optional arg as |undefined
```
class C {
  /**
   * @param {boolean=} x
   */
  f(x = undefined) {}
}
```

Becomes

```
class C {
  /**
   * @param x
   */
  f(x = undefined) {}
}
```

The param type should instead be

```
x?: boolean
```

### LINKED ISSUE #631: gents drops the type when the function has default args
```javascript
/**
 * @param {boolean=} a
 */
function f(a = undefined)
```
Produces `function(a = undefined)` but should be `function(a: boolean|undefined = undefined)`

LucasSloan:
/**
 * @param {boolean=} a
 */
function f(a = undefined)

should translate to

function f(a: boolean|undefined = undefined)

fixes #631, fixes #629

### DIFF
--- a/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
@@ -333,10 +333,7 @@ public final class TypeAnnotationPass implements CompilerPass {
   /** Sets the annotated type expression corresponding to Node {@code n}. */
   private void setTypeExpression(Node n, @Nullable JSTypeExpression type, boolean isReturnType) {
     TypeDeclarationNode node = convert(type, isReturnType);
-    if (node != null) {
-      n.setDeclaredTypeExpression(node);
-      compiler.reportChangeToEnclosingScope(n);
-    }
+    setTypeExpression(n, node);
   }
 
   /** Sets the annotated type expression corresponding to Node {@code n}. */
--- a/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
+++ b/src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
@@ -259,6 +259,7 @@ public final class TypeAnnotationPass implements CompilerPass {
       if (parameterType == null) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/javascript/gents/singleTests/functions.js
```diff
@@ -15,6 +15,11 @@ function twoParams(b, s) {}
 /** @param {!Array<?>=} list */
 function withDefaultValue(list = []) {}
 
+/**
+ * @param {boolean=} a
+ */
+function undefinedDefault(a = undefined) {}
```
rkirov: add a test with something a bit more complicated (maybe !Array<number> or even Array<number>) to make sure the union is done right.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
```diff
@@ -334,6 +339,14 @@ private void setTypeExpression(Node n, @Nullable JSTypeExpression type, boolean
     }
   }
 
+  /** Sets the annotated type expression corresponding to Node {@code n}. */
+  private void setTypeExpression(Node n, @Nullable TypeDeclarationNode type) {
```
rkirov: Is this really useful, despite being 5 lines it does copy/paste the one above, so it can be a maintenance burden.

src/test/java/com/google/javascript/gents/singleTests/functions.js
LucasSloan: Done.

src/main/java/com/google/javascript/gents/TypeAnnotationPass.java
LucasSloan: My first attempt here was to convert the TypeDeclarationNode I created back into a JSTypeExpression before feeding it into the existing function, and that didn't work, so I do think this is necessary.  I DRY'd it up by having the old one call into this one instead of repeating.

rkirov: that's better.

PR #881, opened 2019-05-15
Make sure to associate comments with the correct function args.

### LINKED ISSUE #646: Inline comments are attached to wrong nodes after migration.
```js
foo(p1, p2 /* Comment 2 */, p3, p4 /* Comment 4 */, p5);
```
is converted to
```js
foo(p1, p2, /* Comment 2 */ p3, p4, /* Comment 4 */ p5);
```
The difficulty is that when the comment is on the same line with the code (inline comments) we simply put the comment on the next node. In this case `Comment 2`'s next node is p3. I didn't find a good way to detect whether the comma is before the comment or after the comment.

LucasSloan:
Previously inline comments were attached to the next node after the
comment, but that doesn't work in cases where a comment is after an
argument, but before the comma separating arguments.

fixes #646

### DIFF
--- a/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
+++ b/src/main/java/com/google/javascript/gents/CommentLinkingPass.java
@@ -307,31 +307,7 @@ public final class CommentLinkingPass implements CompilerPass {
         if (parent.isCall()) {
           // We're inside a function call, we have to be careful about which node to attach to, since comments
           // can go before or after an argument.
-          if (n.getNext() == null) {
-            // the last argument in the call, attach the comment here
-            linkCommentBufferToNode(n);
-          } else {
-            int endOfComment = getCurrentComment().location.end.column;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/javascript/gents/CommentLinkingPass.java
```diff
@@ -302,8 +303,37 @@ public final boolean shouldTraverse(NodeTraversal nodeTraversal, Node n, Node pa
 
       if (getLastLineOfCurrentComment() == line) {
         // Comment on same line as code -- we have to make sure this is the node we should attach
-        // it to. It will be the first node after the comment.
-        if (getCurrentComment().location.end.column < n.getCharno()) {
+        // it to.
+        if (parent.isCall()) {
```
rkirov: Maybe extract to a new method - "linkFunctionArgs"

PR #884, opened 2019-05-16
Fix imports from files with multiple namespaces.

LucasSloan:
Previously, gents assumed that each file only had one namespace that it
exported, so when checking if an import was of a default export or not,
it resolved to whether whatever the first exported namespace in the file
was.  This change makes it so each namespace that's exported is tracked
separately.

### DIFF
--- a/src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
+++ b/src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
@@ -3,6 +3,7 @@ package com.google.javascript.gents;
 import static com.google.common.base.MoreObjects.firstNonNull;
 
 import com.google.common.collect.ImmutableSet;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/javascript/gents/CollectModuleMetadata.java
```diff
@@ -192,9 +195,9 @@ void registerProvidesModule(Node n, String file, String namespace) {
     private boolean hasImports = false;
 
     /** namespace of the module in the original closure javascript. */
```
rkirov: I always find the comments in this code confusing because of "namespace" and "module" words used somewhat ad-hoc. Can you take a stab at improving the comments here. For example, adding that we consider goog.module and goog.provides - "namespaces" for the purposes of this calculation.

PR #886, opened 2019-05-29
Add custom type assertions for goog.is* functions.

### LINKED ISSUE #710: clutz: output type guards for goog.base functions
It would be handy if clutz generated type guards for the goog.isX functions:
```
    function isArray(val: any): val is Array<any>;
    function isBoolean(val: any): val is boolean;
    function isDateLike(val: any): val is goog.date.DateLike;
    function isDef<T>(val: T | undefined): val is T;
    function isDefAndNotNull<T>(val: T | undefined | null): val is T;
    function isFunction(val: any): val is Function;
    function isNull(val: any): val is null;
    function isNumber(val: any): val is Number;
    function isObject(val: any): val is Object;
    function isString(val: any): val is string;
```

ribrdb:
Fixes #710

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -2950,6 +2950,47 @@ class DeclarationGenerator {
           && type.toMaybeTemplatizedType().getReferenceName().equals(typeName);
     }
 
+    private boolean handleSpecialTypeAssertionFunctions(String propName) {
+      if (!isGoogNamespace) return false;
+      if (propName.equals("isArray")) {
+        emit("(val : any ) : val is any[] ");
+        return true;
+      }
+      if (propName.equals("isBoolean")) {
+        emit("(val : any ) : val is boolean ");
+        return true;
+      }
+      if (propName.equals("isDef")) {
+        emit("<T> (val : T ) : val is Exclude<T, undefined> ");
+        return true;
+      }
+      if (propName.equals("isDefAndNotNull")) {
+        emit("<T> (val : T ) : val is NonNullable<T> ");
+        return true;
+      }
+      if (propName.equals("isFunction")) {
+        emit("(val : any ) : val is Function ");
+        return true;
+      }
+      if (propName.equals("isNull")) {
+        emit("(val : any ) : val is null ");
+        return true;
+      }
+      if (propName.equals("isNumber")) {
+        emit("(val : any ) : val is number ");
+        return true;
+      }
+      if (propName.equals("isObject")) {
+        emit("(val : any ) : val is Object ");
+        return true;
+      }
+      if (propName.equals("isString")) {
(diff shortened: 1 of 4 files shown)


rkirov:
I am torn here, on one hand this is a good change for people that are still using `goog.is*`, on another hand I rather change gents to replace these with their actual implementations:

isBoolean -> `typeof val == 'boolean'`

which TypeScript has no problem of type narrowing from. I don't recommend anyone keep using `goog.is*` in their TS code.

ribrdb:
Why not? We find the named functions easier to read than the actual implementation.
Especially isDefAndNotNull and isNull are much easier to tell apart than counting the number of = signs.

Also on a related note, I just discovered that goog.reflect.object also needs a custom type:
```
function object <T extends{[key:string]:any}>(type : {new(...args:any[]):{[P in keyof T]:any}}, object : T ) : T ;
```

 The autogenerated one is completely useless. It requires casts to `any` use, which then removes all the type information that Closure Compiler needs for reflection.

rkirov:
Let's get second opinion from @shicks. I think not having this special case will nudge people towards inlining `goog.is*` and friends which we think is beneficial and more idiomatic going forward.

ribrdb:
Are you aware of anything like refasterjs for typescript to help migrate our existing code?
What do you think about adding a special case for goog.reflect.object? The currently generated type is just broken.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -2950,6 +2950,47 @@ private boolean isTemplateOf(JSType type, String typeName) {
           && type.toMaybeTemplatizedType().getReferenceName().equals(typeName);
     }
 
+    private boolean handleSpecialTypeAssertionFunctions(String propName) {
+      if (!isGoogNamespace) return false;
+      if (propName.equals("isArray")) {
+        emit("(val : any ) : val is any[] ");
+        return true;
+      }
+      if (propName.equals("isBoolean")) {
```
shicks: I'm actively working on deprecating and (eventually) deleting all of these.  I'd rather not be encouraging their use.

shicks: In particular, I've got an LSC in progress to replace `isString`, `isBoolean`, `isNumber`, `isDef`, `isNull`, and `isDefAndNotNull` for all internal JS and TS code in one fell swoop.

I may look into doing the other three as well, and possible `goog.typeof` if I can figure out how to do it safely.

rkirov:
@ribrdb about global refactorings, we have been talking about the different options, but for now we just use TSLint with it's autofixer API. Also, for these APIs something as simple as 'sed' might suffice.

rkirov:
Closing this as we are moving away from 'goog.is*' functions. I think something similar will be interesting to do once Type assertions land in 3.7 for Closure's assert methods.

PR #887, opened 2019-06-18
Add support for goog.requireType and goog.forwardDeclare to gents.

rkirov:
They are threated equivalently with goog.require.


evmar:
Typo in description: "threated"

src/main/java/com/google/javascript/gents/ModuleConversionPass.java
```diff
@@ -300,6 +300,12 @@ public void visit(NodeTraversal t, Node n, Node parent) {
         }
       }
     }
+
+    private boolean isARequireLikeCall(Node node) {
```
evmar: Should this check that node itself is a call node or something?

rkirov: done, also cleaned up the destructuring call.

PR #896, opened 2019-07-19
Switch to 'export import' for reexporting aliases.

rkirov:
Clutz needs to emit aliases for any sort of symbol, but in TypeScript
aliases can be awkward. If the symbol is a value and a type we have
been using:

type B = ns.A;
val B: typeof ns.A;

This can be often incorrect because in partial mode we don't know if
ns.A is a value and type or just type or just value. However, it appears
specifically for namespaces there is another syntax:

namespace ns2 {
  export import B = ns.A;
}

which is correct for all possible symbols A (even if A is an interface
with generics).

It appears this only works when one accesses through a namespace 'ns',
but in clutz all internal symbols are in the o_O.clutz namespace.

Thanks to @shick for finding out about 'export import'.

### DIFF
--- a/src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
+++ b/src/main/java/com/google/javascript/clutz/AliasMapBuilder.java
@@ -91,6 +91,12 @@ public class AliasMapBuilder extends ImportBasedMapBuilder {
       } else if (isNamedExportPropAssignment(statement)) {
         // `exports.foo = foo.bar;`
         String localVariableName = getExportsAssignmentPropRootName(statement);
+        if (localVariableName.equals("exports")) {
+          // This is a "local" alias between two exports from the same module.
+          // There is no need to for clutz to special handle this, as the JS
+          // Compiler will resolve this properly.
+          continue;
+        }
         String localPropName = getExportsAssignmentPropName(statement);
         String exportName = getNamedExportName(statement);
         String localNamespaceName =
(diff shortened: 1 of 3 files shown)


evmar:
Should @shicks review?

rkirov:
@shicks should review, but I can't figure out how to assign him in the github UI. It doesn't let me assign anyone. For now I added him to the collaborators list, which should do it.

rkirov:
The null alias was fixed, and the @package visibility is a valid new failure (I am fixing the user's code to remove violation). I think this is almost unblocked, there are just a few more failures to work through and do one more check.

rkirov:
I found one legitimate regression, which happens when in a goog.module one default exports a class and named exports a helper.

For input like:
```js
class C {}
class CInner {}
exports = C;
exports.CInner = CInner
```
we generate 
```ts
declare namespace ಠ_ಠ.clutz { 
  class module$contents$a$b_C {
    a: string;
  }
  class module$contents$a$b_CInner {

  }
}

declare namespace ಠ_ಠ.clutz {                                                                                                    
  export import module$exports$a$b$C = ಠ_ಠ.clutz.module$contents$a$b_C ;  
}                                                                                                                                
declare namespace ಠ_ಠ.clutz.module$exports$a$b$C {                                                           
  export import CInner = ಠ_ಠ.clutz.module$contents$a$b_CInner;                  
}
```

TypeScript now errors with module$exports$a$b$C is defined twice as a namespace. Thankfully there is a single regression, so I will just put an 'any' for now.

rkirov:
The final failure comes from this pattern:

```js
/** @type {string} */
exports.A = '';

/** @type {string} */
exports.B = exports.A;
```

which generates syntactically invalid:

```ts
export import module$exports$a$b$c.B = ಠ_ಠ.clutz.exports.A; 
```
This needs more investigation and probably a fix.

PR #897, opened 2019-07-20
Add a test showing the union typedef collapse feature.

rkirov:
It appears at some point closure has started collapsing typedefs that
only differ by optional fields like:

`A = {t: string, a?: string}` and `B = {t; string, b?: string}`

This results clutz emitting just `A` for the type `A|B`. This is almost
ok, except for the fact that the "unknown object literal" check now
cannot allow assignment of {t: '', b: ''} to what looks like should be
able to accept 'A|B'.

This is adding a test to show this behavior, so that we can notice when
it changes (it apprears to have changed in the last few months).


shicks:
I ran into something like this 6 months ago when I was trying to fix a stack overflow caused by a cyclic typedef (b/120209544).  But I never submitted the change that caused the error, since it broke a bunch of TS code.

Here's some more context on that issue: ultimately it was closed by https://github.com/google/closure-compiler/commit/b43a281849623f18cb645aa4efad264c71bc007c.  The issues came up because I was trying to simplify a subtype check.  I found copy-pasted code in ObjectType#isStructuralSubtypeHelper and PrototypeObjectType#isSubtype.  The difference seemed to be how it dealt with record literal types like the ones you're referring to.  It came down to how to decide whether to do a structural check or not, and when I used the following implementation

```
  boolean shouldMatchStructurally(JSType subType, JSType superType) {
    return subType.isObject()
        && (isStructuralTyping()
           ? superType.isStructuralType()
           : superType.isRecordType());
  }
```

the global presubmits failed because a union type `(function(new:Message)|!Array<function(new:Message)>|{foo:number|undefined})` was collapsing down to just `{foo:number|undefined}` since the latter has only optional fields, so `function` and `Array` are both subtypes.  But then clutz emits the record type only and when TS tries to assign a function or array to it, it fails because TS doesn't consider functions or arrays as assignable to record literal types with only optional fields the same way Closure does.

It's particularly hairy because (technically) Closure is correct, but it's not _useful_ in this case.  I ended up solving this bug in a different way, but I know @concavelenz was working on some changes to how we bookkeep union types w.r.t. structural types, and it's possible that he changed something along those lines, though I'm not sure how it landed without triggering the same sort of failures I saw.

shicks:
I wonder if it's still possible to cleanly revert that commit and see if it's at fault or something else is?  What would it take to bisect to figure out exactly when it started?

rkirov:
The commit you pointed to was from Feb, but I think this behavior changed in the last month or so, because we have internal code that relies on the correct behavior.

PR #898, opened 2019-07-20
Add broken example of package-visible class

shicks:
Note this test is failing because the test actually does verify that the generated .d.ts file is valid.  But this is what is currently generated in production for exported package-private classes.

The upshot is that when the .d.ts error is suppressed, the package-private class is fully usable.

If, however, we change the alias to an `export import`, the error gets promoted out of the .d.ts file and into the actual (erroneous) .ts source - which is great, but the errors need to be fixed upstream before we can land the `export import` change.


rkirov:
rebase and reopen if you think this is still valuable.

PR #899, opened 2019-07-22
Lost enum

rkirov:


rkirov:
Looks like a Closure bug to me. See debugging screenshot attached 
VoidType

PR #903, opened 2019-08-02
Make DeclarationGeneratorTest junit4 parametirized test.

nbeloglazov:
It allows to easily rerun single test in IDE.

PR #904, opened 2019-08-02
Emit "Generated from path/to/file.js" comments for top-level declarations.

nbeloglazov:
The goal is to help developers better understand where symbols coming from and also for tooling to map TS symbols back to original JS files (e.g. for http://kythe.io TS <=> JS integration).

Tested:
  goldens, lots of goldens

Note that at the moment this PR also includes https://github.com/angular/clutz/pull/903. Once #903 is merged I'll update this PR.


nbeloglazov:
@rkirov not really. In fact I only need to annotate `declare namespace` as it contains symbols generated from `.js` files while `declare module` simply references symbols from `declare namespace`. So I removed annotations from `declare module`.

PR #908, opened 2019-08-14
Improve test results output

DominicKramer:
Use `com.adarshr.test-logger` with mocha configuration to list
the tests that have passed and failed and use colorful output.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Wed Aug 14 15:59:18 PDT 2019
```
rkirov: Looks like some tool auto added this. I don't think it's helpful, remove it unless needed.

DominicKramer: I've removed it.

PR #910, opened 2019-08-16
Handle edge case when node might be null when emitting 'generated from' comment.

nbeloglazov:
Tested:
  there is no test for that edge case as I couldn't force it on clean unit test
  tested building some internal Google projects

### DIFF
--- a/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
+++ b/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
@@ -3349,23 +3349,26 @@ class DeclarationGenerator {
         String qualifiedName = innerNamespace + '.' + propName;
         if (provides.contains(qualifiedName)) continue;
         Node node = innerProps.get(namedType);
+        // Node might be null in some edge cases. For example when "type not found in Closure
+        // type registry" comment emmited below.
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -3350,26 +3350,21 @@ void walkInnerSymbols(ObjectType type, String innerNamespace) {
         String qualifiedName = innerNamespace + '.' + propName;
         if (provides.contains(qualifiedName)) continue;
         Node node = innerProps.get(namedType);
+        // Start namespace if this is the first pair being processed.
+        if (!foundNamespaceMembers &&
+            (pType.isEnumType() || isClassLike(pType) || isTypedef(pType))) {
```
rkirov: This is fragile because the clauses here have to match the if stmts below. At least add a comment to that effect.

nbeloglazov: Changed it back. Couldn't think of nice way to reuse the code and prevent possible issue that you mentioned. So decided not worth it :)

nbeloglazov:
Done. I've no idea how to do it properly so I just created new branch from master and applied my changes there. One I need to properly learn git rebase.

PR #912, opened 2019-08-23
Fix exports of local aliases of the form `exports.B = exports.A`.

rkirov:
It appears that the emit was previously very broken for this case
(syntactically incorrect). This change fixes the emit by detecting
rewrites that mention `exports.Foo` and rewrite them to the
corresponding clutz emitted namespace.

Note, that the test .d.ts is still partially incorrect, because we
create a type/let pair for just for a let binding. With #896 this will
be fully correct.


src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1193,8 +1193,17 @@ private void maybeQueueForInnerWalk(
...
    */
-  private void visitKnownTypeValueAlias(String unqualifiedName, String alternativeAliasName) {
-    String emitName = Constants.INTERNAL_NAMESPACE + "." + alternativeAliasName;
+  private void visitKnownTypeValueAlias(String qualifiedName, String alternativeAliasName) {
+    String unqualifiedName = getUnqualifiedName(qualifiedName);
+
+    String emitName = Constants.INTERNAL_NAMESPACE + ".";
+    // If the alias is to a local name, but prefixed with exports.Foo,
+    // we need to rewrite "exports" because it is no longer correct in TS.
+    if (alternativeAliasName.startsWith("exports")) {
```
evmar: `exports.`
with the trailing dot, so you don't match `exportsTheWorld` etc.

rkirov:
On a second thought, a better fix would be to simply not add `exports.Foo` to the alias map, because closure should resolve those locally. Will try that tomorrow.

PR #913, opened 2019-08-27
Fix rawtype warning in Options

shicks:


mprobst:
@rkirov circleci fails due to a Closure Compiler externs change, I believe. Do you have a fix for that already?

rkirov:
Closing this as it will be fixed with the next g3 -> github sync.

PR #917, opened 2019-09-04
The :assemble task builds an executable jar

DominicKramer:
Now when `./gradlew assemble` is run, the jar built contains
all of its dependencies and so it can be run directly.

Run `java -jar build/libs/clutz-1.0.jar clutz ...` to run clutz
and `java -jar build/libs/clutz-1.0.jar gents ...` to run gents.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,3 +107,14 @@ googleJavaFormat {
 testlogger {
   theme 'mocha'
 }
+
+jar {
+  manifest {
+    attributes 'Main-Class': 'com.google.javascript.Main'
+  }
+  from {
+    configurations.compile.collect {
+      it.isDirectory() ? it : zipTree(it)
+    }
+  }
+}
--- a/src/main/java/com/google/javascript/Main.java
+++ b/src/main/java/com/google/javascript/Main.java
@@ -4,20 +4,21 @@ import com.google.javascript.clutz.DeclarationGenerator;
 import com.google.javascript.gents.TypeScriptGenerator;
 
 public class Main {
-    public static void main(String[] args) {
-        if (args.length == 0) {
-            System.out.println(String.format("Usage: %s <clutz|gents> [...args]", Main.class.getSimpleName()));
-            System.exit(1);
-        }
-
-        String tool = args[0];
-        String[] subArgs = new String[args.length-1];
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -107,3 +107,14 @@ googleJavaFormat {
 testlogger {
   theme 'mocha'
 }
+
+jar {
```
rkirov: Can you look into putting it into the "createScript" function. Ideally we should have two separate .jars (with their own deps included) as clutz and gents don't really share that much code or functionality.

If you can't get that to work, it's ok too, but at least document the current approach is not ideal, so that maybe someone else can improve it later.

src/main/java/com/google/javascript/Main.java
```diff
@@ -0,0 +1,24 @@
+package com.google.javascript;
```
mprobst: com.google.javascript.Main is a bit of an odd package/class name. Maybe move it into a separate subpackage?

build.gradle
DominicKramer: Will do.  I'll try to see if I can have separate Jar files, one for gents and one for clutz.  Then the Main class in this PR won't be needed.

src/main/java/com/google/javascript/Main.java
DominicKramer: I'm going to try to see if I can make separate Jars, one for gents and one for clutz, so the Main class isn't needed.

DominicKramer: It is trickier that I thought it would be to have gradle build two jars for the same project.  So I think this approach is the right one to take now.  I don't want to spend more time on this.  I'm fine with changing the name from `com.google.javascript.Main`.  What do you think would be a better name?

rkirov: How about `LinkedJarEntry` or something of sorts. Even more import than the name, add a comment describing that you are adding this for testing/debugging convenience and it is not the intended entry point to either clutz or gents.

DominicKramer: Sounds good.  I'll also add docs.

rkirov:
Is this still relevant give that we are moving to g3 first development?

PR #918, opened 2019-09-04
the CI verifies code style conformance

DominicKramer:


DominicKramer:
It looks like this isn't needed since `google-java-format-gradle-plugin` already runs during `./gradlew check` which the CI already runs.

PR #924, opened 2019-09-19
Replace call to goog.define with assignment to the result of goog.define

lauraharker:
This code no longer works uncompiled and will soon be banned by the Closure Compiler.


shicks:
LGTM

Issue #872, opened 2019-03-23
declare legacy namespace + export obj + enum create wrong output


rkirov:
The following example:

```js
goog.module('legacy.ns.obj');
goog.module.declareLegacyNamespace();

const O = {};

/**
 * @enum {number}
 */
O.E = {
  A: 1,
  B: 2,
  C: 3
};

exports = O;
```

Creates clutz output which makes the expected usage:
```ts
import * as Obj from 'goog:legacy.ns.obj';

const x = Obj.E.A;
```

to fail. Maybe this can be reduced further, so far it looks like the alias

```ts
declare namespace ಠ_ಠ.clutz.legacy.ns.obj {
  let E : ಠ_ಠ.clutz.module$contents$legacy$ns$obj_O.E ;
}
```
for the enum is wrong.

rkirov:
This should be fixed now.

PR #928, opened 2019-09-27
Add test for https://github.com/angular/clutz/issues/872

nbeloglazov:
It seems to be producing correct result now.


nbeloglazov:
Actually hold on. I think the bug is not fixed. Clutz produces different output depending on goog.module name I just got lucky in the test.

nbeloglazov:
Ready for review.

src/main/java/com/google/javascript/clutz/DeclarationGenerator.java
```diff
@@ -1893,6 +1893,7 @@ private void visitEnumType(String symbolName, String qualifiedName, EnumType typ
         emit(elementsTypeName);
         emit(";");
         emitBreak();
+        typesUsed.add(elementsTypeName);
```
rkirov: does this fix the 'enu' or 'enum' test (or both)?

rkirov:
Can you rebase, this looks good.

nbeloglazov:
This change doesn't work in one case like:

```
goog.module('foo');

const bar = goog.require('bar');

/** @enum {number} */
const Color = bar.Color;
exports.Color = Color;
```
where `bar` is coming from another module (so its source is not processed by Clutz). Need to figure out how it should be handled.

PR #931, opened 2019-10-07
Rename goog.provide from 'global' to 'globalNs'

lauraharker:
A Closure Compiler change will begin special casing 'global' as it is native to Node environments. This causes issues when trying to goog.provide it. See also discussion on Google internal
http://cl/270391864.


mprobst:
@rkirov sonatype apparently has a hiccup again. Can we maybe disable the snapshot test to unblock changes while we move to internal first?

rkirov:
Yes, but note that test_released doesn't really pass (green check-mark is showing because it is optional). Making released always pass is at odds with g3 passing tests after sync which runs at head only. Mostly because of golden tests against externs, we can't make a single golden pass against two versions of the JS compiler.

The only way out I see is to prioritize moving to internal first clutz/gents development.

rkirov:
For now, I removed the requirement of "ci/circleci: test_snapshot" passing, so that we can merge trivial changes like this one.

PR #934, opened 2019-10-10
Remove destructuring to fix JSC_GOOG_SCOPE_NON_ALIAS_LOCAL

lauraharker:
The destructuring pattern this test is using is banned in regular compiler invocations, but until an upcoming Closure Compiler change, is allowed in Clutz invocations. [example](https://closure-compiler-debugger.appspot.com/#input0%3Dgoog.provide('a')%253B%26input1%3Dgoog.scope(function()%2520%257B%250A%2520%2520const%2520%257Bb%257D%2520%253D%2520goog.module.get('a')%253B%250A%257D)%253B%26conformanceConfig%26externs%26refasterjs-template%26CHECK_TYPES%3Dtrue%26CLOSURE_PASS%3Dtrue%26PRESERVE_TYPE_ANNOTATIONS%3Dtrue%26PRETTY_PRINT%3Dtrue)

Also adds a regression test for when the right-hand side of a goog.provide assignment uses a goog.module.get. I ran into this case while testing a Closure Compiler change.


lauraharker:
@mprobst sounds good, I updated the commit message.

mprobst:
Thanks. Merged!

